PR #20, opened 2014-02-19
Human-friendly names in JSON serialization and parsing.

isoos:
For testing/debugging, it is easier to write a proto JSON like the following (instead of the integer-based indexing):
{
   "property": "value"
}


BlackHC:
This would solve #37 actually. Any ETA when this is going to be merged in?
Thanks!

BlackHC:
What is the resolution of this? @mit-mit

PR #36, opened 2015-02-10
Fixes dart2js warning about dead code

felixc:
Hi,

dart2js currently (correctly) complains about the "break;" on line 668 being dead code. This minor change should remove the warning while still keeping all cases ending with a break or return (which is my guess for the motivation behind adding the "break" there in the first place).

This is helpful for those who like to keep their builds warning-free, or who treat warnings as erros. Cheers!


felixc:
No CLA needed for Googler contributions; I'm crux@google.com.

skybrian:
Committed.

PR #38, opened 2015-05-05
Add doc comments to GeneratedMessage

srawlins:
This reduces the number of undocumented, public methods of GeneratedMessage from 21 to 13.

I also changed the style of all of the existing doc comments from the `/** */` style to the `///` style.


skybrian:
Merged.

PR #51, opened 2016-01-29
Added implementation and test for freeze and unfreeze.

DavidSouther:
Provides some enforcement for immutability.

### DIFF
--- a/lib/src/protobuf/generated_message.dart
+++ b/lib/src/protobuf/generated_message.dart
@@ -50,7 +50,16 @@ abstract class GeneratedMessage {
 
   UnknownFieldSet get unknownFields => _fieldSet._ensureUnknownFields();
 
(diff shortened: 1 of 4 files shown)


lib/src/protobuf/generated_message.dart
```diff
@@ -50,7 +50,18 @@ abstract class GeneratedMessage {
 
   UnknownFieldSet get unknownFields => _fieldSet._ensureUnknownFields();
 
-  bool get _isReadOnly => false;
+  bool _isFrozen = false;
+  bool get _isReadOnly => _isFrozen;
```
skybrian: Yes that's the simple way, but it only freezes the current object (rather than the tree rooted in this object).

DavidSouther:
Closing in favor of discussion at #53

Issue #69, opened 2017-04-24
CodedBufferWriter should have a method to write to an allocated list


postmasters:
Right now, `CodedBufferWriter` only offers `toBuffer` that returns a newly allocated `Uint8List`. It would be great if we could export `_runningSizeInBytes` and provide another method to write into a pre-allocated buffer. In that case, the caller would be responsible for allocating the buffer. And then `toBuffer` will simply call into this new method.

PR #70, opened 2017-04-24
CodedBufferWriter: Support writing to preallocated buffer. (#69)

postmasters:
This will help reduce the number of allocation and copying when the
serialized message is just part of a bigger buffer.

### DIFF
--- a/test/codec_test.dart
+++ b/test/codec_test.dart
@@ -5,6 +5,7 @@
 
 library pb_codec_tests;
 
+import 'dart:math';
 import 'dart:typed_data';
 
 import 'package:protobuf/protobuf.dart';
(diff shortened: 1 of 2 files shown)


skybrian:
@jakobr-google

test/codec_test.dart
```diff
@@ -751,4 +752,21 @@ void main() {
     expect(readUint64([180, 222, 252, 255, 255, 255, 255, 255, 255, 1]),
         expect64(0xffff2f34, 0xffffffff));
   });
+
+  test('testWriteTo', () {
+    var writer = new CodedBufferWriter()..writeField(0, PbFieldType.O3, 1337);
+    expect(writer.lengthInBytes, 3);
+    var rng = new Random();
```
jakobr-google: I don't like using Random values in a unit test, since it's a gateway to flaky tests (unless you seed the rng with a fixed value).

In this case, you just need sentinel values to make sure the buffer writes don't over/underflow, so a fixed value that's different from the values in `expected` should do the trick.

postmasters:
All fixed. Thanks for the reviews!

jakobr-google:
LGTM. Thanks!

PR #74, opened 2017-07-05
Fix analyzer hints.

jakobr-google:


test/reserved_names_test.dart
```diff
@@ -14,6 +14,7 @@ import 'package:protobuf/mixins_meta.dart' show findMixin;
 import 'mirror_util.dart' show findMemberNames;
 
 // Import the libraries we will access via the mirrors.
+// ignore_for_file: unused_import
```
mit-mit: Not sure I get this comment?

jakobr-google: Analyzer complains about the imports below, since they aren't used directly. They are used through mirrors, though, so the tests fail if I remove the imports. So I opted for telling the analyzer to ignore them.

mit-mit: Got it, thanks

mit-mit:
lgtm

PR #75, opened 2017-08-04
Use real generic syntax instead of comment-based.

jakobr-google:
Also, update pubspec to support v2 dev SDKs.

### DIFF
--- a/lib/src/protobuf/extension_registry.dart
+++ b/lib/src/protobuf/extension_registry.dart
@@ -19,7 +19,7 @@ class ExtensionRegistry {
     map[extension.tagNumber] = extension;
   }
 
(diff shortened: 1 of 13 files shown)


lib/src/protobuf/extension_registry.dart
```diff
@@ -4,29 +4,23 @@
...
     var map = _extensions.putIfAbsent(
         extension.extendee, () => new Map<int, Extension>());
     map[extension.tagNumber] = extension;
   }
 
-  /**
-   * Retrieve an extension from the registry that adds the given tag
-   * number to the given message type.
-   */
+  /// Retrieve an extension from the registry that adds tag number [tagNumber]
```
mravn-google: Retrieves

PR #78, opened 2017-10-23
Add type inference for _inRange

manuelF:
For dart2js builds this doesn't (yet) affect them, but for DDC builds
this is important as _inRange is usually one of the most called methods
and the lack of type annotations make it so all these are dynamic
unnecessarily.

Since the only two usages of this are private and are either int or
double then <T extends num> is a correct restriction.

DDC with this infers either 'int' or 'double' depending on the case.


manuelF:
Done. Feel free to merge whenever you have some time.

PR #81, opened 2018-01-06
Use named JSON fields instead of tag numbers

taisph:
When JSON marshalling protobuf messages using protobuf for Go, the
resulting JSON message uses the actual field names and not tag numbers
as the Dart implementation does. This changes that behaviour to match
the implementation for Go.


taisph:
Sorry. Didn't notice there was already an issue for it. It was actually a change I have had in a fork for quite some time. I just decided to clean it up and submit it yesterday after upgrading protobuf in a separate Dart project. I haven't come across any docs detailing how the JSON encoding should look but I figured the Go implementation was probably more mature and as such implemented by design.

The Go implementation actually has a flag switching between the camel case and snake case field names but I don't remember seeing any flag for using tag numbers instead of field names.

Anyway, I'll take a look at it again later today-ish and see if I can come up with a better solution.

jakobr-google:
Just to be clear, I'm very appreciative of your help! This is a long outstanding request.

The only documented JSON format I've found so far is https://developers.google.com/protocol-buffers/docs/proto3#json. There, they specify using camelCased field names, so looks like we're in luck!

As I said, there are a number of other variations to the JSON encoding that I'd like to (eventually) support, so I'd suggest passing in an options object rather than a bool to switch between these modes, even of the options object only contains a single option for now. Or if you go for another solution, try to make it extensible.

taisph:
Ugh. Requires plugin change apparently. Not sure I even like this proposal.

PR #83, opened 2018-01-21
Support JSON field key names as well as tags

taisph:
Current implementation only allows tag numbers in JSON serialized
protobuf message objects which are not interoperable with other
Proto3 implementations, such as Go. This adds a simple setting to
switch behaviour between tags and names and is easily extensible to
other types if needed.

Example:

import 'package:protobuf/protobuf.dart' as protobuf;

void main() {
  // Changes JSON serialization to expect and write named keys.
  protobuf.jsonFieldKeyType = protobuf.JSONFieldKeyType.proto3;
}


jakobr-google:
I like this approach better, but could we pass the `jsonFieldKeyType` as an argument to the `writeTo`/`mergeFrom` methods instead of having a static field? Otherwise, different modules might step on each other's toes.

taisph:
@jakobr-google Not sure how to best achieve this without polluting the `GeneratedMessage` class with an arguably unrelated JSON specific property because without it I would need to pass it through the `fromJson` constructor (which is how we are currently using this library) which seems to require a change to the dart-protoc-plugin as well and I wanted to avoid that.

But if the addition of the property to `GeneratedMessage` is an acceptable approach, I'll try it out and update the PR.

glenfordwilliams:
are there any updates on this ? currently working on a project where the backend using grpc is written in golang and this would be so useful on the dart web frontend

glenfordwilliams:
@taisph are you still working on this?

taisph:
@glenfordwilliams I asked in https://github.com/dart-lang/protobuf/pull/83#issuecomment-363892287 if it was acceptable to add the property to `GeneratedMessage` but didn't get any comments. It's rather difficult to do anything with the current level of feedback and honestly it works for me™ in its current form. :sweat_smile:

kevmoo:
@mit-mit who should take over here? @sigurdm ?

mit-mit:
Tests are now passing again after I merged in the fix from master

sigurdm:
Hi @taisph 

I agree with other reviewers that the global field is the wrong way to go.

I don't think you need to pollute GeneratedMessage with more fields, just to pass a parameter to the methods.

I made a pull request to yours demonstrating a possible implementation.
My pull request is breaking compatibility though. Another way to go would be to add a new set of methods that did the other kind of serialization.

lib/src/protobuf/json.dart
```diff
@@ -4,6 +4,10 @@
 
 part of protobuf;
 
+enum JSONFieldKeyType { jsonp, proto3 }
```
sigurdm: Needs documentation

sigurdm:
Thinking just a bit more I think we have to add new methods - we should not break the existing generated protos.

rakudrama:
I am still against this change, since a test is at the individual data element level (i.e. inner loop) will make existing code slower.

The difference between `jsonp` and `proto3` should be at the same level as the difference between binary and json - i.e. treated as three independent encodings.

Once we have established a long-term owner with allocated time, I would like to be involved in the design of a much more streamlined implementation. I believe it is possible to get at least a 2x improvement in decoding, but only if the difference in encodings is at the 'outside' rather than the inner loop.

taisph:
I've made another attempt taking all comments and suggestions into consideration. Existing protos should keep working and benchmarks show no impact from the change. It requires the protoc plugin to be adapted which I have a feature branch for but haven't made a PR for yet.

lib/src/protobuf/json.dart
```diff
@@ -5,6 +5,14 @@
...
 
 Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
+  return _writeToJsonMapWithMapper(fs, (fi) => '${fi.tagNumber}');
+}
+
+Map<String, dynamic> _writeToProto3JsonMap(_FieldSet fs) {
+  return _writeToJsonMapWithMapper(fs, (fi) => fi.name);
+}
+
+Map<String, dynamic> _writeToJsonMapWithMapper(_FieldSet fs, Function mapper) {
```
sigurdm: If this is used you should make a typedef for the mapper argument. Something like:
```
typedef _FieldNamer = String Function(FieldInfo fieldInfo);

...

	Map<String, dynamic> _writeToJsonMapWithMapper(_FieldSet fs, _FieldNamer mapper) 
```

lib/src/protobuf/json.dart
```diff
@@ -5,6 +5,14 @@
 part of protobuf;
 
 Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
+  return _writeToJsonMapWithMapper(fs, (fi) => '${fi.tagNumber}');
```
sigurdm: I think we might have to inline the _writeToJsonMapWithMapper in each case to make sure we do not degrade performance. @rakudrama do you agree?

glenfordwilliams:
any updates on this?

mit-mit:
paging @rakudrama

lib/src/protobuf/json.dart
```diff
@@ -68,10 +76,26 @@ Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
...
+
+// Merge fields from a previously decoded JSON object.
+// (Called recursively on nested messages.)
+void _mergeFromJsonMapWithMapper(_FieldSet fs, Map<String, dynamic> json,
+    ExtensionRegistry registry, Function mapper) {
   Iterable<String> keys = json.keys;
   var meta = fs._meta;
   for (String key in keys) {
-    var fi = meta.byTagAsString[key];
+    var fi = mapper(meta, key);
```
rakudrama: This becomes
```dart
FieldInfo fi = keyToFieldInfo[key]
```

lib/src/protobuf/json.dart
```diff
@@ -68,10 +76,26 @@ Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
...
+void _mergeFromProto3JsonMap(
+    _FieldSet fs, Map<String, dynamic> json, ExtensionRegistry registry) {
+  return _mergeFromJsonMapWithMapper(
+      fs, json, registry, (meta, key) => meta.byName[key]);
+}
+
+// Merge fields from a previously decoded JSON object.
+// (Called recursively on nested messages.)
+void _mergeFromJsonMapWithMapper(_FieldSet fs, Map<String, dynamic> json,
+    ExtensionRegistry registry, Function mapper) {
```
rakudrama: Pass in meta.byName or meta.byTagAsString.
This avoids the allocation for the closure that is passed in.
Call the parameter keyToFieldInfo, not the generic name 'mapper'.
```suggestion
    ExtensionRegistry registry, Map<String, FieldInfo> keyToFieldInfo) {
```

I'd also rename `_mergeFromJsonMapWithMapper` to something like `_mergeFromJsonMapWithFieldMap`.

lib/src/protobuf/json.dart
rakudrama: The main issue is how to get the recursive call at line 232 to encode/decode nested messages with the same policy for property names. I believe this CL is broken for nested structures.

Passing through a flag might be the best option.

Secondarily, I don't really like parameterizing with functions. Usually this leads to an allocation at the function creation site, and the indirect call has similar performance to an if-check.

glenfordwilliams:
i was testing this against master and i realized that in the case of messages with embedded messages `writeToProto3Json` doesn't work as it outputs the old way of just `int:value`

### test case
```
import 'package:test/test.dart';
import 'package:site/src/generated/store.pb.dart';
import 'package:site/src/generated/shared.pb.dart';

void main(){

  Store store = Store.create()
  ..email = "someemail.com"
  ..name = "nicks and knacks";

  print("Basic -> ${store.writeToProto3Json()}");


  CreateStoreResponse scr = CreateStoreResponse.create()
  ..message = RequestResponse.create();

  scr.message.message = "message";
  scr.message.error = true;
  scr.message.errorMessage = "errorMessage";

  print("With embedded message -> ${scr.writeToProto3Json()}");


  test("basic serialization", (){
    Map<String, dynamic> res = store.writeToProto3JsonMap();
    expect(res["name"], "nicks and knacks");
    expect(res["email"], "someemail.com");

    Map<String, dynamic> r = scr.writeToProto3JsonMap();

    expect(r["message"], "message");


  });

}
```

### output
```
Basic -> {"name":"nicks and knacks","email":"someemail.com"}
With embedded message -> {"message":{"1":"message","2":true,"3":"errorMessage"}}
Expected: 'message'
  Actual: {'1': 'message', '2': true, '3': 'errorMessage'}
   Which: not an <Instance of 'String'>
package:test_api              expect
test/serialization.dart 31:5  main.<fn>
✖ basic serialization
```

glenfordwilliams:
@taisph

taisph:
@glenfordwilliams You're right. There's a call directly to the non-proto3-mapper on nested messages while converting JSON values. :man_facepalming:

glenfordwilliams:
thanks, gonna do a test and let you know what happens

taisph:
It's not fixed yet. Haven't found a clean solution for nested messages.

glenfordwilliams:
ok no problem

glenfordwilliams:
@taisph  any  updates on this pull ?

mit-mit:
Closing as stale

PR #97, opened 2018-05-16
Dart 2 fixes.

jakobr-google:


lib/src/protobuf/field_set.dart
```diff
@@ -129,7 +129,7 @@ class _FieldSet {
 
   List<T> _getDefaultList<T>(FieldInfo<T> fi) {
     assert(fi.isRepeated);
-    if (_isReadOnly) return const [];
+    if (_isReadOnly) return new List.unmodifiable(<T>[]);
```
lrhn: Alternative:
  return new List<T>.unmodifiable(const <Null>[]);

lib/src/protobuf/rpc_client.dart
```diff
@@ -24,10 +24,10 @@ abstract class RpcClient {
   /// The implementation should serialize the request as binary or JSON, as
   /// appropriate. It should merge the reply into [emptyResponse] and
   /// return it.
-  Future<GeneratedMessage> invoke(
+  Future<T> invoke<T extends GeneratedMessage>(
```
lrhn: This is a breaking change to the signature (if anything implements it without the type argument). 

I assume this is to give a better static type to a function that basically updates `emptyResponse` and returns it?

lib/src/protobuf/field_set.dart
jakobr-google: Yep, that works. Thanks!

lib/src/protobuf/rpc_client.dart
jakobr-google: It's to get rid of a runtime exception (by having a better static type). The generated client stubs look like:

```dart
Future<WhateverResponse> whatever(ClientContext ctx, WhateverRequest request) {
  var emptyResponse = new WhateverResponse();
  return _client.invoke(ctx, 'WhateverService', 'Whatever', request, emptyResponse);
}
```

but since `RpcClient.invoke` returned a `Future<GeneratedMessage>`, we'd get an exception at runtime since `Future<GeneratedMessage>` is not a `Future<WhateverResponse>`.

Externally, dart-lang/dart-protoc-plugin#104 proposed to have the generated code do a cast, but in our internal repo they preferred to add generics to `RpcClient.invoke()`, which should avoid a runtime type check, right?

This part of the PR is just a sync of the internal change. The corresponding change to the protoc plugin is in dart-lang/dart-protoc-plugin#106.

PR #100, opened 2018-05-17
Introduce serialization/deserialization benchmarks.

mraleph:
These are based on benchmarks from original google/protobuf library.


mraleph:
FYI @nichite

PR #101, opened 2018-05-18
Rewrite CodedBufferWriter to speed it up.

mraleph:
The previous implementation was allocating two typed arrays per each written
varint and was using maps to lookup writer functions and wire types for
different field types.

After the rewrite we instead maintain two arrays:

* an array of Uint8List chunks, which serve as temporary output buffers for
small data (ints and floats);
* an array of splices which describes additional bytes to splice in-between
bytes written into chunks.

Note that it is impossible to serialize protobufs in a purely streaming fashion
(using a single continuous output buffer) because some values are serialized
in length-deliminted format - meaning that they are represented as a
varint encoded length followed by the specified number of bytes of data. This
in turn means that we sometimes have to write data before we know how
much space needs to be reseved for the length of the data.

Additionally in some cases value is a raw sequence of bytes that needs to be
written as-is into the output buffer - in this case it does not make sense to
first copy this data into temporary buffer and then copy it again.

These two considerations lead to the splice based design described above.

### DIFF
--- a/lib/src/protobuf/coded_buffer_writer.dart
+++ b/lib/src/protobuf/coded_buffer_writer.dart
@@ -4,204 +4,93 @@
 
 part of protobuf;
 
+/// Writer used for converting [GeneratedMessage]s into binary
+/// representation.
+///
+/// Note that it is impossible to serialize protobuf messages using a one pass
+/// streaming serialization as some values are serialized using
+/// length-delimited representation, which means that they are represented as
+/// a varint encoded length followed by specified number of bytes of data.
+///
+/// Due to this [CodedBufferWritter] maintains two output buffers:
+/// [_outputChunks] which contains all continuously written bytes and
+/// [_splices] which describes additional bytes to splice in-between
+/// [_outputChunks] bytes.
+///
 class CodedBufferWriter {
-  final List<TypedData> _output = <TypedData>[];
-  int _runningSizeInBytes = 0;
-  int get lengthInBytes => _runningSizeInBytes;
+  /// Array of splices representing the data written into the writer.
+  /// Each element might be one of:
+  ///   * a TypedData object - represents a sequence of bytes that need to be
+  ///                          emitted into the result as-is;
+  ///   * a positive integer - a number of bytes to copy from [_outputChunks]
+  ///                          into resulting buffer;
+  ///   * a non-positive integer - a positive number that needs to be emitted
+  ///                              into result buffer as a varint;
+  final List<dynamic> _splices = <dynamic>[];
 
-  static final _WRITE_FUNCTION_MAP = _makeWriteFunctionMap();
+  /// Number of bytes written into [_outputChunk] and [_outputChunks] since
+  /// the last splice was recorded.
+  int _lastSplicePos = 0;
 
-  static ByteData _toVarint32(int value) {
-    // Varint encoding always fits into 5 bytes.
-    Uint8List result = new Uint8List(5);
-    int i = 0;
-    while (value >= 0x80) {
-      result[i++] = 0x80 | (value & 0x7f);
-      value >>= 7;
-    }
-    result[i++] = value;
+  /// Size of the [_outputChunk].
+  static const _chunkLength = 512;
 
-    return new ByteData.view(result.buffer, 0, i);
-  }
+  /// Current chunk used to write data into. Once it is full it is
+  /// pushed into [_outputChunks] and a new one is allocated.
+  Uint8List _outputChunk;
 
-  static ByteData _toVarint64(Int64 value) {
-    // Varint encoding always fits into 10 bytes.
-    Uint8List result = new Uint8List(10);
-    int i = 0;
-    ByteData bytes =
-        new ByteData.view(new Uint8List.fromList(value.toBytes()).buffer, 0, 8);
-    int lo = bytes.getUint32(0, Endian.little);
-    int hi = bytes.getUint32(4, Endian.little);
-    while (hi > 0 || lo >= 0x80) {
-      result[i++] = 0x80 | (lo & 0x7f);
-      lo = (lo >> 7) | ((hi & 0x7f) << 25);
-      hi >>= 7;
-    }
-    result[i++] = lo;
+  /// Number of bytes written into the [_outputChunk].
+  int _bytesInChunk = 0;
 
-    return new ByteData.view(result.buffer, 0, i);
-  }
+  /// ByteData pointing to [_outputChunk]. Used to write primitive values
+  /// more efficiently.
+  ByteData _outputChunkAsByteData;
 
-  static ByteData _int32ToBytes(int value) => _toVarint32(value & 0xffffffff);
+  /// Array of pairs <Uint8List chunk, int bytesInChunk> - chunks are
+  /// pushed into this array once they are full.
+  final List<dynamic> _outputChunks = <dynamic>[];
 
-  static _makeWriteFunctionMap() {
-    writeBytesNoTag(output, List<int> value) {
-      output.writeInt32NoTag(value.length);
-      output.writeRawBytes(
-          new Uint8List(value.length)..setRange(0, value.length, value));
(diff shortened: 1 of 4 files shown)


lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
   }
+
+  /// This function maps a power-of-2 value (2^0 .. 2^31) to a unique value
+  /// in the 0..31 range.
+  ///
+  /// For more details see "Using de Bruijn Sequences to Index a 1 in
+  /// a Computer Word"[1]
+  ///
+  /// Note: this is guaranteed to work after compilation to JavaScript
+  /// where multiplication becomes a floating point multiplication.
```
rakudrama: Fun!

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+            final chunk = _outputChunks[chunkIndex];
+            final bytesInChunk = _outputChunks[chunkIndex + 1];
+
+            // Copy at most bytesToCopy bytes from the current chunk.
+            final leftInChunk = bytesInChunk - chunkPos;
+            final bytesToCopyFromChunk =
+                leftInChunk > bytesToCopy ? bytesToCopy : leftInChunk;
+            final endPos = chunkPos + bytesToCopyFromChunk;
+            while (chunkPos < endPos) {
+              buffer[outPos++] = chunk[chunkPos++];
```
rakudrama: I am wondering if, when compiled with dart2js, indexing chunk is done via an interceptor because the inferred element type of the array is a union.
It might be faster to declare chunk (and bytesInChunk) with types, or use two arrays with types.

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+    ..[_UINT32_BIT_INDEX] = WIRETYPE_VARINT
+    ..[_UINT64_BIT_INDEX] = WIRETYPE_VARINT
+    ..[_FIXED32_BIT_INDEX] = WIRETYPE_FIXED32
+    ..[_FIXED64_BIT_INDEX] = WIRETYPE_FIXED64
+    ..[_SFIXED32_BIT_INDEX] = WIRETYPE_FIXED32
+    ..[_SFIXED64_BIT_INDEX] = WIRETYPE_FIXED64
+    ..[_MESSAGE_BIT_INDEX] = WIRETYPE_LENGTH_DELIMITED;
 }
+
+int _encodeZigZag32(int value) => (value << 1) ^ (value >> 31);
```
rakudrama: This might be broken on dart2js since bitops and shifts are basically unsigned-to-unsigned operations.
An alternative:
(value<<1) ^ (0 - ((value>>32)&1))

Is there a zigzag end-to-end test?

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+    writeTo(result);
+    return result;
+  }
 
-    ByteData makeByteData64(Int64 value) {
-      var data = new Uint8List.fromList(value.toBytes());
-      return new ByteData.view(data.buffer, 0, 8);
+  /// Serializes everything written to this writer so far to [buffer], starting
+  /// from [offset] in [buffer]. Returns `true` on success.
+  bool writeTo(List<int> buffer, [int offset = 0]) {
```
rakudrama: Does this need to be public?
That makes the writes to [buffer] potentially polymorphic.

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
-        return new ByteData(4)..setFloat32(0, value, Endian.little);
-      })
-      ..[PbFieldType._ENUM_BIT] =
-          makeWriter((value) => _int32ToBytes(value.value))
-      ..[PbFieldType._GROUP_BIT] = (output, value) {
-        value.writeToCodedBufferWriter(output);
+      } else {
+        // action is a TypedData containing bytes to emit into the output
+        // buffer.
+        outPos = _copyInto(buffer, outPos, action);
```
rakudrama: The public API is List<int> but _copyInto takes only Uint8List.
Probably best to push the constraint up to the public API, since it is broken is called with a different type of list

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+    _outputChunkAsByteData.setInt32(
+        _bytesInChunk, value & 0xFFFFFFFF, Endian.little);
+    _bytesInChunk += sizeInBytes;
+    _bytesTotal += sizeInBytes;
   }
 
-  void writeRawBytes(TypedData value) {
-    _output.add(value);
-    _runningSizeInBytes += value.lengthInBytes;
+  void _writeInt64(Int64 value) {
```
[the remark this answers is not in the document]
mraleph: * Added test that catches the bug.
* Cleaned up tests in `codec_test.dart` to reduce duplication
* Fixed the code by rewriting it into a pair of `_writeInt32` - this also matches how I implemented writes of varints.

We should definitely consider adding a faster way to access lo/hi from Int64.

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+    _commitSplice();
+
+    int outPos = offset; // Output position in the buffer.
+    int chunkIndex = 0, chunkPos = 0; // Position within _outputChunks.
+    for (int i = 0; i < _splices.length; i++) {
+      final action = _splices[i];
+      if (action is int) {
+        if (action <= 0) {
+          // action is a positive varint to be emitted into the output buffer.
+          int v = -action;
```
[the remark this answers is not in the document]
mraleph: Done.

lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -4,221 +4,494 @@
...
+    final int writtenSizeInBytes = _bytesTotal - _splices[index];
+    _splices[index] = -writtenSizeInBytes;
+    _bytesTotal += _varint32LengthInBytes(writtenSizeInBytes);
+  }
+
+  int _varint32LengthInBytes(int value) {
+    value &= 0xFFFFFFFF;
+    if (value < 0x80) return 1;
+    value >>= 7;
+    if (value < 0x80) return 2;
```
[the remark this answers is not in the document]
mraleph: Done. (Though did not have any statistically significant impact on the benchmark)

lib/src/protobuf/coded_buffer_writer.dart
mraleph: Declared types - which removed the interceptor (there indeed was one).

lib/src/protobuf/coded_buffer_writer.dart
mraleph: Done.

lib/src/protobuf/coded_buffer_writer.dart
mraleph: This method was added in a PR #70 by @postmasters, I don't know if this is really needed.

lib/src/protobuf/coded_buffer_writer.dart
mraleph: I tightened the signature and updated CHANGELOG.md with information about the breaking change.

lib/src/protobuf/coded_buffer_writer.dart
mraleph: Seems to work correctly for _in range integers_: because `>>` still does sign extension for negative numbers (even though it then converts result to unsigned value). 

There are tests for this by they leave behavior of things like: "what should happen if you write `0x8000000` as signed int32" - undefined.

mraleph:
Comments addressed. Merging.

PR #114, opened 2018-07-12
Declare Dart 2.0 compatibility

mit-mit:
@kevmoo can you review if this is how we generally want to update these?

We tested that this works fine with a 2.0 stable build (with the grpc package; except that it still fails as the deps of this package are not updated: https://gist.github.com/sortie/3f1a42b921ee304ea12514076670d223)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.0
+
+* Dart SDK upper constraint raised to declare compatability with Dart 2.0 stable.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.0
```
kevmoo: Can call this 0.9.0+1

No need for a breaking change!

mit-mit:
Good point, incorporated

PR #119, opened 2018-08-23
Sync changes from internal repo.

szakarias:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.0
+
+* Add freezable messages.
```
sigurdm: I think this should be described in more details.

PR #123, opened 2018-08-31
Avoid wrapping builder function from GeneratedMessage.copyWith

sigurdm:


sigurdm:
This will hopefully be made less usefull by https://github.com/dart-lang/protobuf/pull/180

PR #128, opened 2018-09-19
Add type param to ProtobufEnum.initByValue

kevmoo:
Will allow generated code to omit a runtime cast on `valueOf`

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: protobuf
-version: 0.10.3-dev
+version: 0.10.3
 author: Dart Team <misc@dartlang.org>
 description: >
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: protobuf
-version: 0.10.2
+version: 0.10.3-dev
```
szakarias: Why not just 0.10.3?

kevmoo: Are we ready to publish now? We could...

szakarias: Yes, I think we should.

kevmoo: ack

PR #169, opened 2019-01-03
Fix recursive merging

### LINKED ISSUE #139: Recursive merging of sub-messages is not working
For instance consider the proto
```
message Inner  {
 int32 a = 1;
 int32 b = 2;
}

message Outer {
 Inner i = 1;
}
```

and the following dart code 

```dart
Inner i1  = Inner()..a = 6;
Inner i2 = Inner()..b = 7;

Outer o1 = Outer()..i = i1;
Outer o2 = Outer()..i = i2;

o1.mergeFromMessage(o2);
print('a: ${o1.i.a}');
```

The output should be  ```a: 6``` but it is ```a: 0```, meaning that `i` gets overwritten and not merged.

szakarias:
Fixes #139.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -676,8 +676,18 @@ class _FieldSet {
       fieldValue = _cloneMessage(fieldValue);
     }
     if (fi.index == null) {
+      final current = _ensureExtensions()._getFieldOrNull(fi);
+      if (current != null) {
+        fieldValue = current..mergeFromMessage(fieldValue);
+      }
       _ensureExtensions()._setFieldAndInfo(fi, fieldValue);
     } else {
+      if (fi.isGroupOrMessage) {
+        final currentFieldValue = _values[fi.index];
+        if (currentFieldValue != null) {
+          fieldValue = currentFieldValue..mergeFromMessage(fieldValue);
+        }
+      }
       _validateField(fi, fieldValue);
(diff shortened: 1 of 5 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -676,8 +676,18 @@ class _FieldSet {
       fieldValue = _cloneMessage(fieldValue);
     }
     if (fi.index == null) {
+      final current = _ensureExtensions()._getFieldOrNull(fi);
+      if (current != null) {
+        fieldValue = current..mergeFromMessage(fieldValue);
```
sigurdm: In this case I don't think the clone is needed. (and actually harmful, leads to O(n^2) behavior

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -672,22 +672,19 @@ class _FieldSet {
...
     }
 
-    if (mustClone) {
-      fieldValue = _cloneMessage(fieldValue);
+    if (otherFi.isGroupOrMessage) {
+      final currentFi = isExtension
+          ? _ensureExtensions()._getFieldOrNull(fi)
+          : _values[fi.index];
+
+      fieldValue = currentFi == null
```
sigurdm: Perhaps:
```
  fieldValue = currentFi?.mergeFromMessage(fieldValue) ?? _cloneMessage(fieldValue);
```

PR #174, opened 2019-01-08
 Fix freezing of map fields.

szakarias:


protoc_plugin/test/map_field_test.dart
```diff
@@ -267,4 +267,39 @@ void main() {
     testMap.mergeFromBuffer(testMap2.writeToBuffer());
     expect(testMap.int32ToMessageField[2].secondValue, 42);
   });
+
+  test('Freeze message with map field', () {
+    TestMap testMap = TestMap();
+    _setValues(testMap);
+    testMap.freeze();
+
+    try {
```
sigurdm: You can write these tests of throwing a message better with:
```
expect(() => doStuffTheShouldThrow(), throwsA(const TypeMatcher<UnsupportedError>()))
```

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -31,26 +32,37 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+  }
 
   @override
   Iterable<K> get keys => _wrappedMap.keys;
 
   @override
-  V remove(Object key) => _wrappedMap.remove(key);
+  V remove(Object key) {
+    if (_isReadonly)
+      throw new UnsupportedError('Attempted to change a read-only map field');
```
sigurdm: Ideally this would hook into the `frozenMessageModificationHandler`

protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.8
+
+* Add freeze support for map fields.
```
sigurdm: Perhaps this should be phrased as:
```
* Fix freezing of map fields.
```

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -31,26 +32,37 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+    _wrappedMap.remove(key);
```
sigurdm: You need to return the value here

protobuf/lib/src/protobuf/pb_map.dart
szakarias: Leaving this out, per discussion.

protoc_plugin/test/map_field_test.dart
szakarias: Done

protobuf/CHANGELOG.md
szakarias: Done

protobuf/lib/src/protobuf/pb_map.dart
szakarias: Yes, thanks. Done.

PR #179, opened 2019-01-10
Don't create BuilderInfo object for map entries in PBMap constructor.

szakarias:


protoc_plugin/lib/protobuf_field.dart
```diff
@@ -127,18 +127,18 @@ class ProtobufField {
       String valueType = value.baseType.getDartType(fileGen);
       String keyTypeConstant = key.typeConstant;
       String valTypeConstant = value.typeConstant;
+      String valueCreator = (value.baseType.isMessage || value.baseType.isGroup)
+          ? '$valueType.create'
+          : 'null';
+      String valueOf = value.baseType.isEnum ? '$valueType.valueOf' : 'null';
+      String enumValues = value.baseType.isEnum ? '$valueType.values' : 'null';
+      String mapEntryBuildInfo =
+          "new $_protobufImportPrefix.BuilderInfo('entry')\n"
```
sigurdm: Could you move this creating call into `m` in protobuf?

sigurdm: Could we create the correct name for `entry` (including the package-name)? I'm not sure how important it is, but it seems in the spirit to keep the name that protoc creates for the map entry-message.

protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.12.0
+
+* Breaking change: changed PbMap to not create a map entry BuilderInfo object for each PbMap instance, instead it
```
sigurdm: In the breaking change section you should describe the updates to the public api.
Outside that section you can describe the bug fixed.

That way it will be easier for the user to see if/what changes are needed to update.

protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -128,8 +128,19 @@ class BuilderInfo {
...
-  void m<K, V>(int tagNumber, String name, int keyFieldType, int valueFieldType,
-      BuilderInfo mapEntryBuilderInfo) {
+  void m<K, V>(int tagNumber, String name, String entryClassName,
+      int keyFieldType, int valueFieldType,
+      [PackageName packageName,
+      CreateBuilderFunc valueCreator,
+      ValueOfFunc valueOf,
+      List<ProtobufEnum> enumValues]) {
+    BuilderInfo mapEntryBuilderInfo = packageName == null
+        ? BuilderInfo(entryClassName)
```
sigurdm: Do you need the `?:` if `packageName` is `null` you should just be able to pass it there?

protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -128,8 +128,19 @@ class BuilderInfo {
...
+        : BuilderInfo(entryClassName, package: packageName)
+      ..add(1, 'key', keyFieldType, null, null, null, null)
```
sigurdm: Could you extract 1 and 2 as constants somewhere, so they get a name?

protobuf/CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
 ## 0.12.0
 
-* Breaking change: changed PbMap to not create a map entry BuilderInfo object for each PbMap instance, instead it
-  is passed through the static BuilderInfo object in the generated subclasses of GeneratedMessage.
+* Breaking change: Changed `BuilderInfo.m` to take class and package name of the protobuf message representing the map
```
sigurdm: Perhaps write `BuilderInfo.m()` to make it clear that it is a method.

protobuf/lib/src/protobuf/builder_info.dart
szakarias: Not exactly, since the default is not null. I changed it now, so that `packageName` is the last optional parameter in `m()` and `m()` defines a default value.

PR #182, opened 2019-01-17
Ensure getExtension works after freeze.

ekarayel:
# Current (incorrect behaviour)
`getExtension` always returns the default value when it is called after `freeze`, even if the extension was set before the proto was frozen.

# New behaviour
`getExtension` returns extensions when they were set before `freeze()`.
It does not call `_ensureExtensions()` which would mutate the proto.

Note: The additional test fails without the code changes.


ekarayel:
CLA signed.

ekarayel:
Realized that this is a more complex issue.

PR #186, opened 2019-01-18
Fix issue with not being able to read extensions after freeze

szakarias:
Fixes https://github.com/dart-lang/protobuf/issues/171.


szakarias:
cc @ekarayel.

PR #202, opened 2019-02-15
Add more words to the Dart reserved names list.

nichite:
This was causing some issues internally--words that match some data types here (int, double, bool), generate invalid Dart code but work fine in other proto languages.

Since there shouldn't be language-specific unusable words, these should just go through normal name disambiguation.

I tested this out with several other keywords that weren't in the list already: https://www.dartlang.org/guides/language/language-tour#keywords

However, these three were really the only ones that generated broken code when used as field identifiers.

### DIFF
--- a/protobuf/lib/src/protobuf/readonly_message.dart
+++ b/protobuf/lib/src/protobuf/readonly_message.dart
@@ -49,7 +49,7 @@ abstract class ReadonlyMessageMixin {
   void setExtension(Extension extension, var value) =>
       _readonly("setExtension");
 
-  void setField(int tagNumber, var value, [int fieldType = null]) =>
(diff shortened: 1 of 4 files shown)


nichite:
Since this is a small change, added some small tweaks to bring this up to stricter static analysis standards.

sigurdm:
LGTM

Please add a reviewer to each github PR however small. That way there is always someone else knowing what is going on and ff you land it before review write TBR in the conversation.
I think this ought to have an entry in the changelog.
I guess it is not a breaking change because the code relying on these names would have been broken before.

Another (more involved) solution we could have considered is to import dart:core prefixed in the pb.dart file.

PR #204, opened 2019-02-15
Add option for Dart protobuf to emit Kythe metdata.

nichite:
Metadata generation mirrors the logic for the Java proto generator, including how individual fields of a message don't seem to get metadata.


nichite:
I'm not sure why Travis is breaking here--there appear to be no failures? https://travis-ci.org/dart-lang/protobuf/jobs/493915211

sigurdm:
Travis seems to be complaining about formatting of these files:
```
TASK: dartfmt
dartfmt -n --set-exit-if-changed .
test/enum_generator_test.dart
test/file_generator_test.dart
lib/message_generator.dart
```

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 15.0.4
```
sigurdm: This should increment the second number, as it is a feature, not a bug-fix.
I also think you have to merge and make this 16.1.0.

nichite:
Ah, gotcha. This commit chain is embarrassing, don't look. Finally passing though, appears to have been some trivial formatting errors as you said. Yes it's just the same change as internal with no changes, everything looks good! Merging now.

PR #205, opened 2019-02-15
Support for Timestamp

iinozemtsev:
Added generation of `toDateTime` / `fromDateTime` methods to `google.protobuf.Timestamp`. This required adding new imports to generated code, so to avoid potential name collisions I'm adding an extra import like this:

```
import 'dart:core' as $dct show DateTime, Duration;
```

I have added a new file `well_known_types` to make it easier to add similar methods to other types (e.g. to `google.protobuf.Duration`) and also moved `Any` methods generation to it.

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 16.0.1
+
+* Add `DateTime` conversion methods to `google.protobuf.Timestamp`.
+
 ## 16.0.0
 
 * Add ability to generate Kythe metadata files via the
--- a/protoc_plugin/lib/file_generator.dart
+++ b/protoc_plugin/lib/file_generator.dart
@@ -330,12 +330,25 @@ class FileGenerator extends ProtobufContainer {
     }
     if (enumImports.isNotEmpty) out.println();
 
+    _writeWellKnownImports(out);
+
     // Export enums in main file for backward compatibility.
     if (enumCount > 0) {
       Uri resolvedImport =
           config.resolveImport(protoFileUri, protoFileUri, ".pbenum.dart");
       out.println("export '$resolvedImport';");
       out.println();
+    }
+  }
+
+  void _writeWellKnownImports(IndentingWriter out) {
+    for (final message in messageGenerators) {
+      final importStatements =
+          wellKnownTypeForFullName(message.fullName)?.extraImports ??
+              <String>[];
+      for (String importStatement in importStatements) {
+        out.println(importStatement);
+      }
     }
   }
 
--- a/protoc_plugin/lib/well_known_types.dart
+++ b/protoc_plugin/lib/well_known_types.dart
@@ -67,10 +67,11 @@ class _Timestamp extends WellKnownType {
     out.println(r'''
   /// Converts an instance to [DateTime].
   ///
-  /// The result has microsecond precision, as [DateTime] does not support
-  /// nanosecond precision.
+  /// The result is in UTC time zone and has microsecond precision, as
+  /// [DateTime] does not support nanosecond precision.
   $core.DateTime toDateTime() => new $core.DateTime.fromMicrosecondsSinceEpoch(
-      seconds.toInt() * $core.Duration.microsecondsPerSecond + nanos ~/ 1000);
+      seconds.toInt() * $core.Duration.microsecondsPerSecond + nanos ~/ 1000,
+      isUtc: true);
 
   /// Creates a new instance from [dateTime].
   ///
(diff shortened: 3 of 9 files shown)


protoc_plugin/lib/well_known_types.dart
```diff
@@ -0,0 +1,87 @@
...
+      ..typeUrl = '\${typeUrlPrefix}/\${message.info_.qualifiedMessageName}';
+  }''');
+  }
+}
+
+class _Timestamp extends WellKnownType {
+  const _Timestamp();
+
+  List<String> get extraImports =>
+      [r"import 'dart:core' as $dct show DateTime, Duration;"];
```
sigurdm: minor nit: could we call it `$core` instead of `$dct`

protoc_plugin/lib/well_known_types.dart
```diff
@@ -0,0 +1,87 @@
...
+  static Timestamp fromDateTime($dct.DateTime dateTime) {
+    int micros = dateTime.microsecondsSinceEpoch;
+    return new Timestamp()
+      ..seconds = new Int64(micros ~/ $dct.Duration.microsecondsPerSecond)
+      ..nanos = (micros % $dct.Duration.microsecondsPerSecond).toInt() * 1000;
+  }''');
+  }
+}
+
+const _types = {
```
sigurdm: let's call this `_wellKnownTypes` or something similar.

as this is a part file all the names with `_` are available in all the other parts.

protoc_plugin/lib/well_known_types.dart
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of protoc;
+
+abstract class WellKnownType {
+  const WellKnownType();
+
+  List<String> get extraImports => [];
```
sigurdm: add a type to the empty list literal:
```
 List<String> get extraImports => <String>[];
```

protoc_plugin/lib/file_generator.dart
```diff
@@ -312,6 +314,17 @@ class FileGenerator extends ProtobufContainer {
     }
   }
 
+  void _writeWellKnownImports(IndentingWriter out) {
+    for (var m in messageGenerators) {
+      var wellKnownType = wellKnownTypeForFullName(m.fullName);
```
sigurdm: Nit: this could be written:
```dart
for (final message in messageGenerators) {
  final importStatements = wellKnownTypeForFullName(message.fullName)?.extraImports ?? <String>[];
  for (String importStatement in  importStatements) {
    out.println(importStatement);
  }
}
```

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 15.0.4
```
sigurdm: I think this needs a merge already.

Also there needs to be a corresponding update in `protoc_plugin/pubspec.yaml`

protoc_plugin/lib/well_known_types.dart
```diff
@@ -0,0 +1,87 @@
...
+class _Timestamp extends WellKnownType {
+  const _Timestamp();
+
+  List<String> get extraImports =>
+      [r"import 'dart:core' as $dct show DateTime, Duration;"];
+
+  @override
+  void generateMethods(IndentingWriter out) {
+    out.println(r'''
+  /// Converts an instance to [DateTime].
```
sigurdm: Should we somehow explain that this loses nanosecond precision?

protoc_plugin/lib/well_known_types.dart
iinozemtsev: Done

protoc_plugin/lib/well_known_types.dart
iinozemtsev: Done

iinozemtsev:
Thanks for the review!

protoc_plugin/lib/well_known_types.dart
```diff
@@ -0,0 +1,93 @@
...
+      [r"import 'dart:core' as $core show DateTime, Duration;"];
+
+  @override
+  void generateMethods(IndentingWriter out) {
+    out.println(r'''
+  /// Converts an instance to [DateTime].
+  ///
+  /// The result has microsecond precision, as [DateTime] does not support
+  /// nanosecond precision.
+  $core.DateTime toDateTime() => new $core.DateTime.fromMicrosecondsSinceEpoch(
```
sigurdm: Thinking about this again: I think we need to make the timestamp utc instead of local time-zone (That is how the TimeStamp is documented).
```
$core.DateTime.fromMicrosecondsSinceEpoch(..., isUtc: true);
```
Also that needs to be tested

iinozemtsev: Thanks! I'll update the pull request next Tuesday when I'm back from vacation

iinozemtsev: Please take a look at the last commit

PR #206, opened 2019-02-19
Remove usage of closures in _mergeFromCodedBufferReader, make optional parameter mandatory in _readRawVarint32

mkustermann:


protobuf/lib/src/protobuf/coded_buffer_reader.dart
```diff
@@ -164,19 +164,22 @@ class CodedBufferReader {
...
 
-  int _readRawVarint32([bool signed = true]) {
+  int _readRawVarint32(bool signed) {
     // Read up to 10 bytes.
-    int bytes = _currentLimit - _bufferPos;
+    int bufferPos = _bufferPos;
+    int bytes = _currentLimit - bufferPos;
     if (bytes > 10) bytes = 10;
     int result = 0;
     for (int i = 0; i < bytes; i++) {
```
sigurdm: If this loop shows up on the profile we could even consider unrolling it for small values of `bytes`

protobuf/lib/src/protobuf/coded_buffer_reader.dart
```diff
@@ -164,19 +164,22 @@ class CodedBufferReader {
     return _buffer[_bufferPos - 1];
   }
 
-  int _readRawVarint32([bool signed = true]) {
+  int _readRawVarint32(bool signed) {
     // Read up to 10 bytes.
-    int bytes = _currentLimit - _bufferPos;
+    int bufferPos = _bufferPos;
```
szakarias: Maybe you could leave a comment here explaining why you are saving `_bufferPos` in a local variable.

mkustermann: Done.

protobuf/lib/src/protobuf/coded_buffer_reader.dart
mkustermann: Maybe it would help. The issue is that the AOT compiler currently cannot inline the `buffer_[...]` access because it doesn't know which representation it is (_ExternalUint8List, _Uint8ArrayView, _Uint8List, ...)

PR #212, opened 2019-02-22
Avoid storing the BuilderInfo in _FieldSet

sigurdm:
This should allow us to send messages between isolates (https://github.com/dart-lang/protobuf/issues/167). And also make this more efficient (as we do not need to copy the metadata.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -33,7 +33,6 @@ FrozenMessageErrorHandler frozenMessageModificationHandler =
 /// be faster when compiled to JavaScript.
 class _FieldSet {
   final GeneratedMessage _message;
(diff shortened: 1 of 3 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -53,8 +52,9 @@ class _FieldSet {
   final Map<int, int> oneofCases = <int, int>{};
 
   _FieldSet(this._message, BuilderInfo meta, this._eventPlugin)
```
szakarias: Remove the meta parameter?

sigurdm:
It seems that this is slowing down our benchmarks significantly :( .
I need to investigate a bit more. Maybe there's somethings we can do.

sigurdm:
I'm not gonna find time to mitigate this.
Closing the PR for now.

PR #213, opened 2019-02-22
Chunked decoder

sigurdm:
This is still work in progress

This would allow dividing the decoding of messages between e.g. frames of a Flutter app.

### DIFF
--- a/protobuf/test/event_test.dart
+++ b/protobuf/test/event_test.dart
@@ -186,6 +186,7 @@ main() {
     r.mergeFromBuffer(bytes);
     // The changes should not include the repeated message.
     r.deliverChanges();
+    print(r);
     checkLog(log, [
       [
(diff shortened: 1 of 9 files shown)


protobuf/test/event_test.dart
```diff
@@ -186,6 +186,7 @@ main() {
     r.mergeFromBuffer(bytes);
     // The changes should not include the repeated message.
     r.deliverChanges();
+    print(r);
```
athomas: Is that just a leftover from debugging?

sigurdm: Yes - thanks for pointing it out!

sigurdm:
Will close this for now.
It set back decoding speed - and the real solution to ui-jank in  is better support for isolates.

PR #217, opened 2019-03-07
import dart:core as core in the generated code

### LINKED ISSUE #211: String Type Conflict
I originally posted this on SO and was advised to make an issue here too
(https://stackoverflow.com/questions/54783804/dart-type-conflicts)

I'm working with a Proto file which defines a number of types to be used within a system, it's not something I have control over and it's causing a conflict after compiling to Dart as one of the defined types is named String which clashes with Dart's core String type

```
syntax = "proto3";

message String {
    string value = 1;
}
```

Which when compiled to Dart gives me 

```
import 'dart:core' show int, bool, double, String, List, Map, override;
import 'package:fixnum/fixnum.dart';
import 'package:protobuf/protobuf.dart' as $pb;
import 'Framework.pbenum.dart';
export 'Framework.pbenum.dart';

class String extends $pb.GeneratedMessage {
  static final $pb.BuilderInfo _i = new $pb.BuilderInfo('String', package: const $pb.PackageName('sila2.org.silastandard'))
    ..aOS(1, 'value')
    ..hasRequiredFields = false
  ;

  String() : super();
  String.fromBuffer(List<int> i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) : super.fromBuffer(i, r);
  // Warning here
  String.fromJson(String i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) : super.fromJson(i, r);
  String clone() => new String()..mergeFromMessage(this);
  String copyWith(void Function(String) updates) => super.copyWith((message) => updates(message as String));
  $pb.BuilderInfo get info_ => _i;
  static String create() => new String();
  String createEmptyInstance() => create();
  static $pb.PbList<String> createRepeated() => new $pb.PbList<String>();
  static String getDefault() => _defaultInstance ??= create()..freeze();
  static String _defaultInstance;
  static void $checkItem(String v) {
    if (v is! String) $pb.checkItemFailed(v, _i.qualifiedMessageName);
  }

  // Warning here
  String get value => $_getS(0, '');
  set value(String v) { $_setString(0, v); }
  bool hasValue() => $_has(0);
  void clearValue() => clearField(1);
}
```
Error:
`
Error: The argument type 'String (.../Framework.pb.dart)' can't be assigned to the parameter type 'String (C:\src\flutter\bin\cache\pkg\sky_engine\lib\core\string.dart)'. (argument_type_not_assignable at (.../Framework.pb.dart:50)`

One way around seems to be using a prefix, but I don't really want to be modifying generated code

```
import 'dart:core' show int, bool, double, List, Map, override;
import 'dart:core' as core;

core.String get value => $_getS(0, '');
set value(core.String v) { $_setString(0, v); 
```
I understand this is more of a problem with a poor proto definition but would it be possible or worthwhile to add this prefixing automatically to protoc? Or is there another solution?

szakarias:
and add camel_case_types to analysis_options.

fixes #211  

cc @nichite

### DIFF
--- a/protoc_plugin/lib/service_generator.dart
+++ b/protoc_plugin/lib/service_generator.dart
@@ -150,14 +150,16 @@ class ServiceGenerator {
   }
 
   void _generateRequestMethod(IndentingWriter out) {
-    out.addBlock('$_generatedMessage createRequest(String method) {', '}', () {
+    out.addBlock(
+        '$_generatedMessage createRequest($_coreImportPrefix.String method) {',
+        '}', () {
       out.addBlock("switch (method) {", "}", () {
         for (MethodDescriptorProto m in _methodDescriptors) {
           var inputClass = _getDartClassName(m.inputType);
           out.println("case '${m.name}': return new $inputClass();");
         }
         out.println("default: "
-            "throw new ArgumentError('Unknown method: \$method');");
+            "throw new $_coreImportPrefix.ArgumentError('Unknown method: \$method');");
       });
     });
     out.println();
@@ -166,7 +168,7 @@ class ServiceGenerator {
   void _generateDispatchMethod(out) {
     out.addBlock(
         '$_future<$_generatedMessage> handleCall($_serverContext ctx, '
-        'String method, $_generatedMessage request) {',
+        '$_coreImportPrefix.String method, $_generatedMessage request) {',
         '}', () {
       out.addBlock("switch (method) {", "}", () {
         for (MethodDescriptorProto m in _methodDescriptors) {
@@ -175,7 +177,7 @@ class ServiceGenerator {
               "case '${m.name}': return this.$methodName(ctx, request);");
         }
         out.println("default: "
-            "throw new ArgumentError('Unknown method: \$method');");
+            "throw new $_coreImportPrefix.ArgumentError('Unknown method: \$method');");
       });
     });
     out.println();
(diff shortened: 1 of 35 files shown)


protoc_plugin/lib/service_generator.dart
```diff
@@ -193,8 +195,10 @@ class ServiceGenerator {
       _generateRequestMethod(out);
       _generateDispatchMethod(out);
       _generateMoreClassMembers(out);
-      out.println("Map<String, dynamic> get \$json => $jsonConstant;");
-      out.println("Map<String, Map<String, dynamic>> get \$messageJson =>"
+      out.println(
+          "$_coreImportPrefix.Map<$_coreImportPrefix.String, $_coreImportPrefix.dynamic> get \$json => $jsonConstant;");
+      out.println(
```
sigurdm: Can you split or refactor this line?

protoc_plugin/lib/well_known_types.dart
```diff
@@ -39,7 +39,7 @@ class _Any extends WellKnownType {
   ///
   /// Can be used with a default instance:
   /// `any.canUnpackInto(Message.getDefault())`
-  bool canUnpackInto($_protobufImportPrefix.GeneratedMessage instance) {
+  $_coreImportPrefix.bool canUnpackInto($_protobufImportPrefix.GeneratedMessage instance) {
```
sigurdm: Ouch, this is another reason to move these into mixins in the protobuf library.

sigurdm:
It would be lovely to split out camel_case_types into another PR.
You can also land them together if it's too much

PR #224, opened 2019-03-20
Override == and hashCode in PbMap

szakarias:


protobuf/pubspec.yaml
```diff
@@ -9,6 +9,7 @@ environment:
   sdk: '>=2.0.0-dev.17.0 <3.0.0'
 dependencies:
   fixnum: '>=0.9.0 <0.11.0'
+  quiver: ^2.0.2
```
sigurdm: I'd prefer we do not get further dependencies if we can avoid it. I think it is better to duplicate it.
Hopefully we can have a resolution of https://github.com/dart-lang/sdk/issues/11617 someday, and use core library functionality.

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,35 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+    if (identical(other, this)) {
+      return true;
+    }
+    if (other is! PbMap) {
+      return false;
+    }
+    if (other.length != length) {
+      return false;
+    }
+    if (other.hashCode != hashCode) {
```
sigurdm: I think calculating the hash code during equality testing is slower than just doing structural equality testing.

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,35 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+    }
+    for (final key in keys) {
+      if (other[key] != this[key]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  int get hashCode {
```
sigurdm: This should also have an `@override` annotation

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,35 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+    return hashObjects(_wrappedMap.keys
```
sigurdm: Ideally we should be able to share code with _FieldSet.hashCode and PbList.hashCode

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,35 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+        .map((key) => hash2(key.hashCode, _wrappedMap[key].hashCode))
+        .toList(growable: false)
+          ..sort());
```
sigurdm: I think we can just rely on the default dart map being ordered (it is a LinkedHashMap)?

sigurdm: No of course not! - different order of insertion should still result in the same hashcode. I think we can xor the hashcodes of all the entries.

rakudrama:
Question: why do we need structural `==` and `hashCode` for PbMaps when Dart's corelib Map implementations have identity `==` and `hashCode`?

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,35 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+      return false;
+    }
+    if (other.length != length) {
+      return false;
+    }
+    if (other.hashCode != hashCode) {
+      return false;
+    }
+    for (final key in keys) {
+      if (other[key] != this[key]) {
```
rakudrama: `other[key]` could be `null`.
I'd add a comment that `this[key]` can never be `null`, so it is not necessary to check `other.containsKey(key)`, since the test will fail for missing `key`.

As a TODO - It might be more efficient to check all the keys first, since they are usually smaller than the values.

protobuf/lib/src/protobuf/pb_map.dart
rakudrama: hash2 takes two Objects, and calls their hashCodes.
So the calls the hashCode here are unnecessary.

hashObjects takes an Iterable of Objects and calls the hashCodes.

In effect we are doing a lot of x.hashCode.hashCode.
At best that seems inefficient, and may result in a poor hashCode if the underlying combiner doesn't have good dispersal.
At least we can fix the two calls here.

protobuf/lib/src/protobuf/pb_map.dart
rakudrama: Much slower if the first key is not in the other map!

It is also exponential in the nesting depth of the PbMaps when the map values contain PbMaps -
you walk the maps for this hashCode call, and again in the `!=` below.

Perhaps you can make a benchmark for PbMaps nested 20x deep to prove we avoid this problem.

sigurdm:
> Question: why do we need structural == and hashCode for PbMaps when Dart's corelib Map implementations have identity == and hashCode?

It is a really good question.

It is to fit in with the existing protobuf implementation.
Currently `PbList` has structural `==` and `hashCode`, changing that would be a breaking change. It seems weird that `PbMap` should not have. Especially given that `GeneratedMessage` also implements structural `==` and `hashCode` that recurses into its fields.

I agree this is a wrong design: protobuf messages are by nature open-ended and structural equality is more or less meaningless (unknown fields cannot be compared). But it seems very late to change that now, and the equality checks have some use in writing tests.

protobuf/lib/src/protobuf/utils.dart
```diff
@@ -35,4 +35,29 @@ bool _areByteDataEqual(ByteData lhs, ByteData rhs) {
   return _areListsEqual(asBytes(lhs), asBytes(rhs));
 }
 
+List<T> sorted<T>(Iterable<T> list) => new List.from(list)..sort();
+
+// Jenkins hash functions
```
sigurdm: provide a link to where we loaned this from

protobuf/lib/src/protobuf/utils.dart
```diff
@@ -35,4 +35,29 @@ bool _areByteDataEqual(ByteData lhs, ByteData rhs) {
   return _areListsEqual(asBytes(lhs), asBytes(rhs));
 }
 
+List<T> sorted<T>(Iterable<T> list) => new List.from(list)..sort();
```
sigurdm: This is duplicate

protobuf/lib/src/protobuf/utils.dart
```diff
@@ -35,4 +35,29 @@ bool _areByteDataEqual(ByteData lhs, ByteData rhs) {
...
+
+// Jenkins hash functions
+
+int _combine(int hash, int value) {
```
sigurdm: Either give these a more distinct name or collect them as static methods in a (private) class

protoc_plugin/test/map_field_test.dart
```diff
@@ -212,6 +213,27 @@ void main() {
...
+
+    PbMap<int, String> m = t.int32ToStringField;
+    PbMap<int, String> m2 = t2.int32ToStringField;
+
+    expect(t, t2);
+    expect(t.hashCode, t2.hashCode);
+
+    expect(m, m2);
+    expect(m == m2, isTrue);
+    expect(m.hashCode, m2.hashCode);
```
sigurdm: Add a negative test?

protoc_plugin/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
+## 16.0.4
```
sigurdm: Perhaps name this a dev-version, we don't need to publish a new protoc version.

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -39,6 +39,36 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @override
```
sigurdm: We should add dartdoc to pbmap explaining how the hash and equals semantics work.

protobuf/lib/src/protobuf/utils.dart
```diff
@@ -35,4 +35,27 @@ bool _areByteDataEqual(ByteData lhs, ByteData rhs) {
   return _areListsEqual(asBytes(lhs), asBytes(rhs));
 }
 
-List<T> sorted<T>(Iterable<T> list) => List.from(list)..sort();
+List<T> sorted<T>(Iterable<T> list) => new List.from(list)..sort();
```
sigurdm: Let's deprecate this, and make a private version. It has nothing to do in the exports of protobuf.

PR #225, opened 2019-03-25
Generate '@Deprecated' Annotation for Deprecated Proto Fields

nichite:
This change brings Dart closer in line to other languages like Java, which already generate this annotation. It should be dropped in dart2js and have no real code impact size for production users.

### DIFF
--- a/protoc_plugin/lib/file_generator.dart
+++ b/protoc_plugin/lib/file_generator.dart
@@ -285,9 +285,8 @@ class FileGenerator extends ProtobufContainer {
       out.println(_asyncImport);
     }
 
-    out.println('// ignore: UNUSED_SHOWN_NAME');
     out.println(
-        '$_coreImport show int, bool, double, String, List, Map, override, Deprecated;\n');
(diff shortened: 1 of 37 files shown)


protoc_plugin/lib/file_generator.dart
```diff
@@ -292,7 +292,7 @@ class FileGenerator extends ProtobufContainer {
 
     out.println('// ignore: UNUSED_SHOWN_NAME');
     out.println(
-        '$_coreImport show int, bool, double, String, List, Map, override;\n');
+        '$_coreImport show int, bool, double, String, List, Map, override, Deprecated;\n');
```
sigurdm: Now you're here, could you sort this show-list?

sigurdm:
Do you plan to also support deprecated messages, enums, files etc.?

protoc_plugin/lib/file_generator.dart
nichite: Done.

nichite:
I think adding support for the messages, etc is probably a good idea too. I'll look into what the other languages are doing.

PR #229, opened 2019-04-02
Fix JSON serialization of unsigned 64-bit fields.

### LINKED ISSUE #44: Int64 is signed, but we use it to store unsigned 64-bit values
We are storing both signed and unsigned 64-bit integers in as instances of Int64, even though the Int64 class is intended for signed values. It's up to the caller to treat the the value appropriately.

This is confusing because methods on Int64 such as isNegative() return the wrong thing for numbers from 1<<31 to the top of the range.

To fix this we would need to add another type to the fixnum library:
https://github.com/dart-lang/fixnum/issues/20

nichite:
The current behavior treats signed and unsigned 64-bit fields the same, which relies on users needing to know to flip the negative appropriately. With the introduction of `toStringUnsigned` in fixnum v 0.10.9, we can handle this correctly.

This should fix #44.

### DIFF
--- a/protobuf/pubspec.yaml
+++ b/protobuf/pubspec.yaml
@@ -8,7 +8,7 @@ homepage: https://github.com/dart-lang/protobuf
 environment:
   sdk: '>=2.0.0-dev.17.0 <3.0.0'
 dependencies:
-  fixnum: '>=0.10.9 <0.11.0'
+  fixnum: ^0.10.9
 dev_dependencies:
   test: '>=1.2.0'
   benchmark_harness: any
--- a/protobuf/pubspec.yaml
+++ b/protobuf/pubspec.yaml
@@ -1,5 +1,5 @@
 name: protobuf
-version: 0.13.7
+version: 0.13.8
 author: Dart Team <misc@dartlang.org>
 description: >
   Runtime library for protocol buffers support.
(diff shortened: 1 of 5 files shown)


protobuf/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ homepage: https://github.com/dart-lang/protobuf
 environment:
   sdk: '>=2.0.0-dev.17.0 <3.0.0'
 dependencies:
-  fixnum: '>=0.9.0 <0.11.0'
+  fixnum: '>=0.10.9 <0.11.0'
```
sigurdm: Could this just be `^10.9` ?

protoc_plugin/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
 dependencies:
   fixnum: ^0.10.5
```
sigurdm: Any reason not to update this to `^10.9` as well?

protobuf/pubspec.yaml
nichite: Good point, done.

protoc_plugin/pubspec.yaml
nichite: Also done.

PR #233, opened 2019-04-05
Move eventPlugin callback when setting a field.

szakarias:


nichite:
Since we want to get this synced internally quickly, I'm going to merge this PR if that's okay.

PR #237, opened 2019-04-09
Configure travis to run on all OS'es

mit-mit:
Run tests on Mac, Linux, and Windows

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,11 @@ jobs:
 stages:
   - format_analyzer_tests
 
+os:
+  - linux
+  - osx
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -70,6 +70,11 @@ jobs:
 stages:
   - format_analyzer_tests
 
+os:
```
sigurdm: This file is generated, the update should happen in mono_repo.yml

mit-mit: New attempt in https://github.com/dart-lang/protobuf/pull/238

PR #239, opened 2019-04-11
Only initialize _FieldSet._onofCases if the message contains oneofs

### LINKED ISSUE #236: _FieldSet.oneofCases mostly unused
`_FieldSet.oneofCases` field is useful in less than 10% of protobuf messages. It wastes memory to allocate a map for all these cases, approximately doubling the size of a small protobuf message. I think all accesses are guarded by tests on `_meta.oneofs`, so I think this simple change or something like it would work:

```
   : oneofCases = meta.oneofs.isEmpty ? null : {}
```

There should be a memory size benchmark that captures this effect.

Nit: The field has a public name whereas all the other fields have private names - it would be nice if they were consistent.

szakarias:
Fixes #236

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.13.11
+
+* Save memory by only initialzing `_FieldSet.oneofCases` if the message contains oneofs.
+
 ## 0.13.10
 
 * Fix recursive merging of repeated elements.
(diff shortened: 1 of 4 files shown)


protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.13.11
+
+* Save memory by only initialzing `_FieldSet.oneofCases` if the message contains oneofs.
```
sigurdm: Typo in `initializing`

PR #241, opened 2019-04-11
Preserve the original order of proto fields when generating metadata.

### LINKED ISSUE #240: Kythe metadata sometimes points to incorrect proto field
Consider this sample proto:

message MyProto {
  optional int64 field_1 = 1;
  optional int64 field_3 = 3;
  optional int64 field_5 = 9;
  optional int64 field_4 = 8;
  optional int64 field_2 = 2;
}

When using the generated metadata for Kythe:
field_1 will show references to field_1
field_3 will show references to field_2
field_5 will show references to field_3
field_4 will show references to field_4
field_2 will show references to field_5

Somewhere the fields are getting sorted based on the tag number, and then when you follow reference from the Nth field (where N is the position, when counted in the source file) the references for the Nth field in the sorted list is shown. (missing tag numbers are also fine; it just picks the Nth item in the sorted list)

As long as all fields in the proto as ordered based on the tag numbers, the references continue to work fine.

nichite:
This corrects metadata edges pointing to the wrong field if the fields were not originally sorted by tag number. This should fix #240.

### DIFF
--- a/protoc_plugin/test/message_generator_test.dart
+++ b/protoc_plugin/test/message_generator_test.dart
@@ -110,7 +110,7 @@ void main() {
     mg.generate(writer);
 
     var eq = ListEquality();
-    var fieldStringsMap = new HashMap(
+    var fieldStringsMap = HashMap(
         equals: eq.equals, hashCode: eq.hash, isValidKey: eq.isValidKey);
     fieldStringsMap[[4, 0]] = ['PhoneNumber'];
     fieldStringsMap[[4, 0, 2, 0]] = ['type', 'hasType', 'clearType'];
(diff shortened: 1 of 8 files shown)


protoc_plugin/test/message_generator_test.dart
```diff
@@ -87,4 +94,45 @@ void main() {
...
+    GeneratedCodeInfo metadata = writer.sourceLocationInfo;
+    for (var annotation in metadata.annotation) {
+      String annotatedName =
+          generatedContents.substring(annotation.begin, annotation.end);
+      var expectedStrings = fieldStringsMap[annotation.path];
+      if (expectedStrings == null) {
+        fail('The field path ${annotation.path} '
+            'did not match any expected field path.');
+      }
+      expect(annotatedName, isIn(expectedStrings));
```
sigurdm: I like this explicit testing of the source code.

nichite: Me too, I think this is a solid pattern.

PR #246, opened 2019-05-06
Ignore fields with tag number 0

sigurdm:
Except for index calculations.

This will allow us to change fields to dummy-fields without disturbing
the index calculations.


sigurdm:
cc nichite@ zarah@ sra@

PR #250, opened 2019-05-28
Always prefix .pb.dart imports in .pbserver, .pb.json, .pbgrpc files.

nichite:
A lack of prefixing has led to internal issues where invalid code is generated (names generated in the main .pb.dart file or its imports collide with names in the .pbserver, .pb.json, or .pbgrpc generated files.

### DIFF
--- a/protoc_plugin/lib/file_generator.dart
+++ b/protoc_plugin/lib/file_generator.dart
@@ -312,12 +312,12 @@ class FileGenerator extends ProtobufContainer {
     _findProtosToImport(imports, enumImports);
 
     for (var target in imports) {
-      _writeImport(out, config, target, ".pb.dart");
+      _writeImport(out, config, target, ".pb.dart", true);
     }
     if (imports.isNotEmpty) out.println();
 
     for (var target in enumImports) {
-      _writeImport(out, config, target, ".pbenum.dart");
+      _writeImport(out, config, target, ".pbenum.dart", true);
     }
     if (enumImports.isNotEmpty) out.println();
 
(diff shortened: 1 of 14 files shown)


protoc_plugin/lib/file_generator.dart
```diff
@@ -570,12 +570,20 @@ class FileGenerator extends ProtobufContainer {
...
   /// (Possibly the same .proto file.)
+  ///
+  /// When generating the main file (if [forMainFile] is true), all imports
+  /// should be prefixed unless the target file is the main file. For other
+  /// files (e.g. .pbserver.dart, .pbgrpc.dart), all .pb.dart imports should
+  /// always be prefixed, even if they're for the main proto file.
   void _writeImport(IndentingWriter out, OutputConfiguration config,
-      FileGenerator target, String extension) {
+      FileGenerator target, String extension,
+      [bool forMainFile = false]) {
```
rakudrama: I'd use a named optional parameter - it is easier to understand what 'true' means at the call site.

protoc_plugin/lib/file_generator.dart
```diff
@@ -570,12 +570,20 @@ class FileGenerator extends ProtobufContainer {
 
   /// Writes an import of a .dart file corresponding to a .proto file.
   /// (Possibly the same .proto file.)
+  ///
+  /// When generating the main file (if [forMainFile] is true), all imports
```
rakudrama: What do you want to omit the import prefix?
Can the logic be simplified by never not using a prefix?

sigurdm: I am also a bit confused. I think we might as well always prefix...

protoc_plugin/lib/file_generator.dart
nichite: I ended up just taking the parameter out and always prefixing as per the other suggestion.

protoc_plugin/lib/file_generator.dart
nichite: Done, might as well always prefix as you both suggested.

nichite:
So everything's green again--to clarify on the parameter: this doesn't decide whether or not to prefix `pb.dart` file imports--it always prefixed them. However, I decided it was doing the same thing as a basic check against the file extension, so I removed it in `file_generator.dart` and just check if it's a `pb.dart` file.

The other check in `_getDartClassName` remains (in `service_generator.dart`). The whole reason for this PR is essentially: the `protoFileUri != target.protoFileUri` check doesn't always do what we want it to. It evaluates to true when generating, say, `hello.pb.dart` from `hello.proto`, where we know those generated Dart classes will be in the same file, but it also evaluates to true for `hello.pbserver.dart`, `hello.pbgrpc.dart`, etc. In those cases collisions can occur from the generated class, so best to always prefix `pb.dart` imports.

The one exception would be `client_generator.dart`--the generated client API doesn't get its own file (unlike the server), but it uses the same `_getDartClassName` method as the server file, so it uses the param to say "hey, I'm in the same main pb.dart file where all these classes are used." I added some more comments to make that more clear.

PR #256, opened 2019-06-11
Add method for retrieving extensions from unknown fields

sigurdm:


sigurdm:
I would love suggestions for a better name for `getExtensionReparsing`!

ekupershlak:
Consider folding this functionality into getExtension (with an optional ExtensionRegistey paramter), since that is a natural place to look for this behavior. When the registry is absent, the behavior would be unchanged; when it's present, the registry would be used to parse the extension. Ideally, this could also update the message object to make subsequent calls faster, but I can imagine this being more contentious.

Also, would it be possible to add an equivalent hasExtension method to determine if the extension is not present?

sigurdm:
I moved the reparsing functionality into getExtension when an explicit registry is provided.
I think it makes sense - hopefully the documentation is enough to make people do the correct error handling.

sigurdm:
Sorry - I committed that too fast. There is still a bunch of issues to work out.

sigurdm:
Re-using getExtension opens too many issues of conflicts with existing behavior.
Especially related to repeated fields (getExtension of an unset repeated field creates an empty list in that field and returns that, it is not really clear what we should do when given an extensionRegistry)

So my new suggestion is to have ExtensionRegistry.reparseMessage(message) that returns a shallow copy of message with all extensions in the registry parsed.

This is much more orthogonal to the existing functionality, and seems much more predictable.

WDYT?

ekupershlak:
That would sovle the problem.

Would this require reparsing the whole message or just the extensions that were previously not parsed?

sigurdm:
> Would this require reparsing the whole message or just the extensions that were previously not parsed?

It only needs to reparse the extensions that were previously not parsed.

See https://github.com/dart-lang/protobuf/pull/256/files#diff-0b809233615748bef2b2ce4405133ce9R54 in the implementation where we loop over the extensions of the registry, checking which ones are present in the unknown fields.

protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -7,7 +7,7 @@ part of protobuf;
 /// Per-message type setup.
 class BuilderInfo {
   /// The fully qualified name of this message.
-  final String qualifiedMessageName;
+    final String qualifiedMessageName;
```
szakarias: indent

protoc_plugin/test/extension_test.dart
```diff
@@ -244,76 +245,115 @@ void main() {
...
-        withExtension
-            .getExtensionReparsing(Extend_unittest.outer)
-            .getExtension(Extend_unittest.extensionInner)
-            .value,
-        'def');
   });
 
   test(
-      'getExtensionReparsing reparses extensions that was not in the original registry',
+      'ExtensionRegistry.reparseMessage reparses extensions that was not in the original registry',
```
szakarias: was -> were

protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -33,13 +33,94 @@ class ExtensionRegistry {
     }
     return null;
   }
+
+  /// Returns a shallow copy of [message], with all extensions in [this] parsed
+  /// from its unknown fields.
+  ///
+  /// Extensions already present in [message] will
```
szakarias: Incomplete sentence

sigurdm: fixed

PR #261, opened 2019-06-27
Use FieldDescriptorProto.json_name for representing field names.

sigurdm:
This seems to be the right thing to do if we want to emit
proto3 JSON.
Currently we were using the dart mangled name, that is also camel-case
but has a number of dart-specific disambiguations applied making it unsuitable for
interop.

If we want to parse JSON we need also know the field_name (with
underscores). That is for a later PR...

### DIFF
--- a/protoc_plugin/pubspec.yaml
+++ b/protoc_plugin/pubspec.yaml
@@ -1,5 +1,5 @@
 name: protoc_plugin
-version: 18.0.0
+version: 18.0.0-dev
 author: Dart Team <misc@dartlang.org>
 description: Protoc compiler plugin to generate Dart code
 homepage: https://github.com/dart-lang/protobuf
--- a/protoc_plugin/pubspec.yaml
+++ b/protoc_plugin/pubspec.yaml
@@ -1,5 +1,5 @@
 name: protoc_plugin
-version: 17.0.2
+version: 16.0.6
 author: Dart Team <misc@dartlang.org>
 description: Protoc compiler plugin to generate Dart code
 homepage: https://github.com/dart-lang/protobuf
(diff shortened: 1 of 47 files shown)


sigurdm:
https://github.com/dart-lang/protobuf/issues/153

sigurdm:
CC @nichite

protoc_plugin/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: protoc_plugin
-version: 17.0.2
+version: 16.0.6
```
szakarias: wrong version?

protoc_plugin/test/goldens/imports.pb
```diff
@@ -13,9 +13,9 @@ import 'package2.pb.dart' as $2;
 
 class M extends $pb.GeneratedMessage {
   static final $pb.BuilderInfo _i = $pb.BuilderInfo('M')
-    ..a<M>(1, 'm', $pb.PbFieldType.OM, M.getDefault, M.create)
-    ..a<$1.M>(2, 'm1', $pb.PbFieldType.OM, $1.M.getDefault, $1.M.create)
-    ..a<$2.M>(3, 'm2', $pb.PbFieldType.OM, $2.M.getDefault, $2.M.create)
+    ..a<M>(1, '', $pb.PbFieldType.OM, M.getDefault, M.create)
```
szakarias: Consider adding a JSON name in the test that generates this file.

protoc_plugin/pubspec.yaml
sigurdm: Fixed

protoc_plugin/test/goldens/imports.pb
sigurdm: Done

PR #266, opened 2019-07-09
Introduce GeneratedMessage isFrozen

sigurdm:


sigurdm:
cc @nichite @rakudrama do you have any concerns about adding this?

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -61,6 +61,11 @@ abstract class GeneratedMessage {
     return this;
   }
 
+  /// Returns `true` if this message is marked read-only. Otherwise `false`.
+  ///
+  /// Even when false, some sub-message could be read-only.
```
rakudrama: when true, can submessages be modifiable?

sigurdm: Should not happen, we always freeze submessages recursively. Clarified.

PR #267, opened 2019-07-11
Reduce getDefault code size by adding specialized getters.

nichite:
- Add specialized getters for String, int and bool with usual default values.

- The closure that generates the default singleton instance is created
  from the `create` method. `GeneratedMessage._defaultMakerFor(create)`
  returns a function that returns the singleton instance.

- BuilderInfo.aOM (optional) and BuilderInfo.aQM (required) add a
  message field, using the `_defaultMakerFor`.

- {Message}.getDefault() now uses `GeneratedMessage.$_defaultFor`
  to ensure the singleton, and still caches the result in a static
  variable to keep the same performance when called repeatedly.


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -390,6 +392,46 @@ class _FieldSet {
...
+      if (defaultValue != null) return defaultValue;
+      value = _getDefault(_nonExtensionInfoByIndex(index));
+    }
+    bool result = value;
+    return result;
+  }
+
+  /// The implementation of a generated getter for `bool` fields that default to
+  /// `false`.
+  bool _$getBF(int index) {
```
sigurdm: Could we reduce duplication with something like:
```dart
  T _getWithDefault<T>(int index, T default) {
    var value = _values[index];
    T result = value ?? default;
    return result;
  }

  _getIntegerDefaultZero(int index) => _getWithDefault(index, 0);
  _getStringDefaultEmpty(int index) => _getWithDefault(index, '');
  ...
```

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 17.0.5
+
+* Generate specialized getters for String, int, and bool with usual
+  default values.
```
sigurdm: Add that the minimum supported protobuf version has gone up.
I guess technically that turns this into a breaking change of protoc_plugin.

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -369,8 +369,10 @@ class _FieldSet {
...
   }
 
-  /// The implementation of a generated getter. Common case for submessages.
-  T _$getN<T>(int index) {
+  /// The implementation of a generated getter for a default value determined by
+  /// the field definition value. Common case for submessages. dynamic type
+  /// pushes the type check to the caller.
+  dynamic _$getND(int index) {
     var value = _values[index];
     if (value != null) return value as T;
```
sigurdm: This is no longer generic in T. Travis also tells this. I guess we just want to remove the cast?

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -369,8 +369,10 @@ class _FieldSet {
     return _getDefault(_nonExtensionInfoByIndex(index)) as T;
   }
 
-  /// The implementation of a generated getter. Common case for submessages.
-  T _$getN<T>(int index) {
+  /// The implementation of a generated getter for a default value determined by
+  /// the field definition value. Common case for submessages. dynamic type
+  /// pushes the type check to the caller.
+  dynamic _$getND(int index) {
```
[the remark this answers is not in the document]
sigurdm: Nevermind - I had the same objections before internally.
Sorry for the noise.

protobuf/lib/src/protobuf/field_set.dart
rakudrama: @sigurdm  I would not do that - casts / checks of generic type are more expensive that casts to a fixed type because casts to a fixed type can be specialized. This is especially true of primitive types like bool, num, String where the specialized JavaScript predicates JIT to a few instructions.

Knowing the default also allows dart2js to potentially generate better code for ??.

protobuf/lib/src/protobuf/field_set.dart
rakudrama: Looks like a merge error.

protobuf/lib/src/protobuf/field_set.dart
sigurdm: What a shame. But good to know!

nichite:
Merged in a separate PR, closing this one.

PR #270, opened 2019-08-12
Fix dummy fields

sigurdm:
The FieldInfo object created be the treeshaker before would trigger assertions.

### DIFF
--- a/protobuf/test/dummy_field_test.dart
+++ b/protobuf/test/dummy_field_test.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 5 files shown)


protobuf/test/dummy_field_test.dart
```diff
@@ -0,0 +1,23 @@
...
+    ..add(0, null, null, null, null, null, null);
+  Message createEmptyInstance() => Message();
+
+  @override
+  GeneratedMessage clone() => throw UnimplementedError();
+}
+
+main() {
+  test('Has no known fields', () {
+    expect(Message(), Message());
```
szakarias: remove?

sigurdm: Done

sigurdm:
> Curious why it was published well after being merged and even after later changes?
I had forgot to publish it back when it landed. And now when I published it I forgot to include: https://github.com/dart-lang/protobuf/pull/271

So sorry.

PR #273, opened 2019-08-20
 Introduce protoName 

sigurdm:
This is in anticipation of proto3 json support that requires both the json name and the proto-definition name of each field.

@nichite and @rakudrama you might be interested.

### DIFF
--- a/protobuf/lib/src/protobuf/builder_info.dart
+++ b/protobuf/lib/src/protobuf/builder_info.dart
@@ -28,20 +28,27 @@ class BuilderInfo {
       dynamic defaultOrMaker,
       CreateBuilderFunc subBuilder,
       ValueOfFunc valueOf,
-      List<ProtobufEnum> enumValues) {
+      List<ProtobufEnum> enumValues,
+      {String protoName}) {
     var index = byIndex.length;
     final fieldInfo = (tagNumber == 0)
         ? FieldInfo.dummy(index)
-        : FieldInfo<T>(name, tagNumber, index, fieldType, defaultOrMaker,
-            subBuilder, valueOf, enumValues);
+        : FieldInfo<T>(name, tagNumber, index, fieldType,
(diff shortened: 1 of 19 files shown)


protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -71,71 +80,65 @@ class BuilderInfo {
   }
 
   void a<T>(int tagNumber, String name, int fieldType,
-      [dynamic defaultOrMaker,
+      {dynamic defaultOrMaker,
```
szakarias: Do we know how much we increase the output size by changing to named parameters in the field-adder methods?

sigurdm: According to @johnniwinther it should make no difference at least for the dart2js case.

rakudrama: Will `protoName:` always be used?
It seems like adding an extra piece of information for every field will make apps quite a bit larger.
Can the field be computed?

Even if you don't use the optional argument, the default value will be passed, adding to the size of each call. In minified code you can expect the extra to look like `,null` for small protos or `,x` for large ones (where there is also `var x=null`), so the size overhead is 5 bytes per field, and slightly less for protos with many fields.

This is high, considering that the code-size-saving specializations like `aOS` have only two arguments.

sigurdm: The field can be computed in most cases, and thus used rarely.
But once in a while there is a name where the mapping is not bijective.
I was assuming that an unused argument would not take up more space when not passed...
That is indeed a high cost.
I could encode the two names in the name argument somehow. I just worry about the complexity.

rakudrama: I realize I am wrong ... I was describing what happens with a static method, but this is an instance method, so there will be a stub to add the extra argument.

Sorry for the false alarm!

PR #275, opened 2019-08-26
Generate exports for proto libraries that are publicly imported.

sigurdm:


protoc_plugin/test/protos/import_public.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
+
+import public "foo.proto";
+
+message A {
```
nichite: Maybe a quick comment here or in the test file linking to the import public syntax or what it means to have a public import? https://developers.google.com/protocol-buffers/docs/proto#importing-definitions

protoc_plugin/test/import_public_test.dart
```diff
@@ -0,0 +1,14 @@
...
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'package:test/test.dart';
+
+import '../out/protos/import_public.pb.dart';
+
+main() {
+  test('can reference a message type imported publicly', () {
+    expect(Foo(), TypeMatcher<Foo>());
```
nichite: nit: a slightly more fluent way to use `TypeMatcher` is to use `isA<Foo>()` instead of the matcher directly: https://api.flutter.dev/flutter/package-matcher_matcher/isA.html

protoc_plugin/test/protos/import_public.proto
sigurdm: Good idea. Done

protoc_plugin/test/import_public_test.dart
sigurdm: I used isA at first, but that is incompatible with the versions of test that support the oldest sdk we support.
Maybe we should soon raise the lower sdk constraint?

PR #277, opened 2019-09-05
Change ExtensionRegistry.reparseMessage to also reparse extensions on…

szakarias:
… any submessage.

### DIFF
--- a/protobuf/lib/src/protobuf/extension_registry.dart
+++ b/protobuf/lib/src/protobuf/extension_registry.dart
@@ -35,11 +35,13 @@ class ExtensionRegistry {
   }
 
   /// Returns a shallow copy of [message], with all extensions in [this] parsed
-  /// from the unknown fields of [message].
+  /// from the unknown fields of [message] and of every nested submessage.
   ///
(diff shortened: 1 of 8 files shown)


protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -104,14 +107,58 @@ T _reparseMessage<T extends GeneratedMessage>(
         unknownField.writeTo(tagNumber, codedBufferWriter);
       }
       resultUnknownFields._fields.remove(tagNumber);
+      reparsedExtensions = true;
```
sigurdm: I guess we should only do this (and I guess the line above...) if `unknownField != null`

protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -92,6 +94,7 @@ T _reparseMessage<T extends GeneratedMessage>(
     T message, ExtensionRegistry extensionRegistry) {
   T result = message.createEmptyInstance();
```
sigurdm: Can we avoid creating the empty instance and the shallow copy until we find something?

PR #280, opened 2019-09-08
Make mapEntryBuilderInfo public

### LINKED ISSUE #276: Allow public access to `enumValues` for map values
note: I'll referring to the [field_info.dart](https://github.com/dart-lang/protobuf/blob/master/protobuf/lib/src/protobuf/field_info.dart) file below:

There exists in `MapFieldInfo` the exposed `CreateBuilderFunc valueCreator`, for creating values of the map. However, a similar construct is not exposed for getting enumValues. This is exposed for normal `FieldInfo` through the variable `final List<ProtobufEnum> enumValues`.
I believe such a variable (perhaps `valueEnumValues`) should be exposed in `MapFieldInfo` as it:
  1. Is valid Proto to have Enums as map values
  2. Follow the pattern of exposing this as is done in `FieldInfo`
  3. Gives developers using this library more flexibility (if they're searching for it)

nwparker:
This change exposes the private `BuilderInfo` variable `mapEntryBuilderInfo` as a public variable.

The benefits are described in the issue, but I will restate/refine here:
  1. Allows accessing `enumValues` for map values that are enums (was not possible before)
  2. Also allows general access to the FieldInfo of map key/values (developer flexibility)
  3. Exposing the `BuilderInfo` allows the direct use of the `PbMap` constructor as opposed to needing to call `createMapField` on an instance of a `GeneratedMessage`

Fixes #276

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.14.2
+
+*  Expose `mapEntryBuilderInfo` in `MapFieldInfo`.
+
 ## 0.14.1
 
 * Support for `import public`.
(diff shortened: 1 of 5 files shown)


nwparker:
@googlebot I signed it!

PR #282, opened 2019-09-09
Annotate fields accessors with tagNumber 

sigurdm:
This is to support the protobuf aware treeshaker.

See https://dart-review.googlesource.com/c/sdk/+/116208

Annotations are not present in the output of dart2js or aot compiled code (Flutter release mode). So this should have no influence on the size of the output.


sigurdm:
@nichite & @rakudrama are you ok with this?

PR #296, opened 2019-09-23
Reduce generated code size via getDefault and specialized getters.

nichite:
* Add specialized getters for `String`, `int`, and `bool` with usual default values.
* Shrink dart2js generated code for `getDefault()`.

Due to a miscommunication on our side, this change was languishing on one of our internal forks for a while. Important to get it in now.


protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 18.0.3
```
sigurdm: This is actually a breaking change since it requires an updated protobuf version

sigurdm: I thought to publish 
https://github.com/dart-lang/protobuf/pull/294
and
https://github.com/dart-lang/protobuf/pull/290
and
https://github.com/dart-lang/protobuf/pull/282

Together as 19.0.0 to make fewer breaking revisions. This could be part of that.

nichite: That sounds reasonable. How do you want go about merging them?

PR #297, opened 2019-09-25
Fixed a few lints

jonasfj:


protobuf/pubspec.yaml
```diff
@@ -16,3 +16,4 @@ dev_dependencies:
   test: '>=1.2.0'
   benchmark_harness: any
   js: any
+  pedantic: ^1.8.0
```
kevmoo: add trailing newline

kevmoo: Also, use ^1.0.0 so we can keep testing on older SDKs

jonasfj: Also configured my editor to do this. `"files.insertFinalNewline": true,` in vscode :)

PR #298, opened 2019-09-27
Add dart_package option

sigurdm:


sigurdm:
I'm not entirely sure that this is good enough.
If you generate the protobuf code into a subdirectory of a package (like lib/generated) the dart_package annotation would have to look like 
```
option (dart_options.dart_package) = "foo/generated/";
```

domesticmouse:
Sigurd, `dartfmt --fix -w .`?

protoc_plugin/lib/file_generator.dart
```diff
@@ -130,13 +132,22 @@ class FileGenerator extends ProtobufContainer {
...
-      throw "FAILURE: Import with absolute path is not supported";
+  static Uri calculateUri(FileDescriptorProto descriptor) {
+    // protoc should never generate an import with an absolute path.
+    assert(!Uri.file(descriptor.name).isAbsolute,
+        "Import with absolute path is not supported");
+
+    if (descriptor.options.hasExtension(Dart_options.dartPackage)) {
+      String dartPackage =
+          descriptor.options.getExtension(Dart_options.dartPackage);
+      return Uri(scheme: 'package', path: '${dartPackage}/${descriptor.name}');
```
dave, line 143: Hey @sigurdm I've been having a look at this, and I think this should be:

```dart
return Uri(scheme: 'package', path: '${dartPackage}/${path.basename(descriptor.name)}');
```

`descriptor.name` returns the full path of the proto file.

In my project the proto root is not the same as the package directory, so I get something like:

dartPackage: "foo/bar"
descriptor.name: "bar/baz.proto"

Combining them together gives "foo/bar/bar/baz.proto" when it should be "foo/bar/baz.proto".

I believe it would be better to ignore the path from the descriptor because the dart package name should always point directly to the directory that contains the dart source file. Right?

sigurdm: I belive the dart package should not have slashes. It should just be a single name (the name of the package where the proto belongs)

dave: ooh interesting... That would be tricky in my project... I have one directory structure with several dart packages, and they all share a common proto root.

dave: Yeah it seems like an unnecessary restriction that your proto file structure should be the same as your dart package structure. Let me explain my project structure:

```
+ project root
  + server (go package)
    + main go package
    + repo1
    + repo2
    + ...
    + tests (go tests)
  + lib (main dart root)
    + main dart package
  + repositories (other dart packages)
    + repo1 (dart package)
    + repo2 (dart package)
    + ...
  + proto (proto root)
    + main (proto files for main package)
    + repo1 (proto files for "repo1" package)
    + repo2 (proto files for "repo2" package)
    + ...
  + test (dart tests)
```

It's convenient to have the proto files all in the same directory structure, but the file structure of the Dart packages doesn't match this. I'm trying to ensure my proto files stay platform independent, so I don't really want to rearrange my proto files to match my Dart directory structure. Nor do I really want to have a separate proto root for each package.

sigurdm: I think it makes sense to have one proto-root per package, if the protos truly belongs in that package. (At least the generated files have to be moved into `lib/` of that package.

dave: OK I'll have a go at rearranging my project / protoc commands. I'll let you know how I get on!

fzyzcjy:
👀

oklimberg:
Any plans for the missing review?

PR #299, opened 2019-09-27
Avoid name collisions with Int64 and values beginning with digits.

nichite:
This fix prevents us from generating invalid code when certain proto fields are "Int64" or start with digits, which apparently is allowed for at least proto enums.

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 19.0.1
+
+* Fix: avoid naming collisions with Int64 and values beginning with digits.
+
 ## 19.0.0
 * Breaking: Generates code that requires at least `protobuf` 0.14.4.
   If protoc_plugin is installed in your path with `pub global activate` you can upgrade with `pub global activate protoc_plugin 19.0.0`
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 19.0.1
 
-* Fix: avoid naming collisions with Int64 and values beginning with digits.
+* Fix: avoid naming collisions with `Int64` and enum names beginning with digits
+  after an initial underscore.
 
 ## 19.0.0
 * Breaking: Generates code that requires at least `protobuf` 0.14.4.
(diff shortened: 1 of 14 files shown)


protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 19.0.1
+
+* Fix: avoid naming collisions with Int64 and values beginning with digits.
```
sigurdm: * Fix: avoid naming collisions with `Int64` and names beginning with digits.

protoc_plugin/lib/names.dart
```diff
@@ -80,6 +84,9 @@ String avoidInitialUnderscore(String input) {
   while (input.startsWith('_')) {
```
sigurdm: Perhaps rename this function. Not sure I have a good name, but the current name becomes misleading.
Also update the comment

sigurdm:
I am not able to compile a proto with an enum-value starting with a digit.
If you are able to, please add a test/protos/x.proto file deomnstrating this, and a test importing and using it. That is a much  stronger test than a golden file test.

nichite:
I believe this can only happen in cases where the name starting with a digit is also prefixed with an underscore. So, the name mangling removes the underscore and makes it an invalid Dart name. I'll fix up the tests as you described.

sigurdm:
Ahhh - that makes sense. Good find!

protoc_plugin/lib/names.dart
```diff
@@ -70,16 +70,24 @@ class OneofNames {
...
+
 /// Move any initial underscores in [input] to the end.
 ///
 /// According to the spec identifiers cannot start with _, but it seems to be
-/// accepted by protoc.
+/// accepted by protoc. These identifiers are private in Dart, so they have to
+/// be transformed.
 ///
-/// These identifiers are private in Dart, so they have to be transformed.
+/// If [input] starts with a digit after transformation, prefix with an 'x'.
```
sigurdm: Just a thought: maybe it would be prettier to not move the underscore to the end when there is also a digit, we anyway prefix it with an x.

protoc_plugin/lib/names.dart
sigurdm: Sorry that comment came too late. I see you already merged. Nevermind, it is so rare anyway.

PR #300, opened 2019-09-30
useProtoNames optional argument for `toProto3Json`

gleyba:


protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -217,8 +217,9 @@ abstract class GeneratedMessage {
   /// message encoding a type not in [typeRegistry] is encountered, an
   /// error is thrown.
   Object toProto3Json(
-          {TypeRegistry typeRegistry = const TypeRegistry.empty()}) =>
-      _writeToProto3Json(_fieldSet, typeRegistry);
+          {TypeRegistry typeRegistry = const TypeRegistry.empty(),
+          bool useProtoNames = false}) =>
```
sigurdm: Perhaps it would be better named `useJsonNames` with a default of `true`.
protoName is something I invented because `name` was already given to the jsonName

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -217,8 +217,9 @@ abstract class GeneratedMessage {
   /// message encoding a type not in [typeRegistry] is encountered, an
   /// error is thrown.
```
sigurdm: The comment needs a description of  the new option

PR #302, opened 2019-10-01
Avoid using `extension` as an import prefix.

sigurdm:
It is a built-in-identifier with the new static extension methods. Leading this test to fail.

https://github.com/dart-lang/language/blob/master/accepted/future-releases/static-extension-methods/feature-specification.md


sigurdm:
We need to make another PR that ensures that the name `extension` gets mangled.

PR #308, opened 2019-10-10
Remove the lines about proto3 compatibility in the README file

MusiKid:
This commit removes the lines about the proto3 compatibility since there isn't open issues about it.


MusiKid:
@googlebot I signed it!

PR #324, opened 2019-11-15
Handle unknown fields part of a oneof

sigurdm:


sigurdm:
Companion CL https://dart-review.googlesource.com/c/sdk/+/125345 has tests.

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -342,14 +343,17 @@ class _FieldSet {
     return newValue;
   }
 
-  /// Sets a non-extended field and fires events.
-  void _setNonExtensionFieldUnchecked(FieldInfo fi, value) {
-    int tag = fi.tagNumber;
-    int oneofIndex = _meta.oneofs[tag];
+  void _updateOneOfCase(int newTagnumber) {
```
szakarias: Call this in ` _$set` (around line 500)

sigurdm: Done

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -342,14 +343,17 @@ class _FieldSet {
...
+    int oneofIndex = _meta.oneofs[newTagnumber];
     if (oneofIndex != null) {
       _clearField(_oneofCases[oneofIndex]);
-      _oneofCases[oneofIndex] = tag;
+      _oneofCases[oneofIndex] = newTagnumber;
```
mkustermann: Maybe only call it conditionally

```dart
final currentTag = _oneofCases[oneofIndex];
if (currentTag != null) _clearField(currentTag);
```

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -300,8 +300,11 @@ abstract class GeneratedMessage {
 
   /// Clears the contents of a given field.
   ///
-  /// If it's an extension field, the Extension will be kept.
-  /// The tagNumber should be a valid tag or extension.
+  /// If it's an extension field, the value will be removed, but the Extension
+  /// will be kept.
+  ///
+  /// If the tagNumber is present as an unknown field, that field will be
+  /// removed.
```
mkustermann: `GeneratedMessage.mergeFromMessage()` to call `_FieldSet._mergeFromMessage()`, which looks like this:
```dart
  void _mergeFromMessage(_FieldSet other) {
    for (FieldInfo fi in other._infosSortedByTag) {
      var value = other._values[fi.index];
      if (value != null) _mergeField(fi, value, isExtension: false);
    }
    if (other._hasExtensions) {
      var others = other._extensions;
      for (int tagNumber in others._tagNumbers) {
        var extension = others._getInfoOrNull(tagNumber);
        var value = others._getFieldOrNull(extension);
        _mergeField(extension, value, isExtension: true);
      }
    }
  
    if (other._hasUnknownFields) {
      _ensureUnknownFields().mergeFromUnknownFieldSet(other._unknownFields);
    }
  }
```

This merging does not seem to account for `oneof`s. Does it need to be updated as well? Are there more places?

protobuf/lib/src/protobuf/field_set.dart
sigurdm: Yes - that's better

protobuf/lib/src/protobuf/generated_message.dart
sigurdm: `_FieldSet._mergeFromMessage()` calls `_mergeField` that ends up calling `_setNonExtensionFieldUnchecked` that handles the oneofs

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -265,8 +265,9 @@ class _FieldSet {
       }
     }
 
-    // neither a regular field nor an extension.
-    // TODO(skybrian) throw?
+    if (_hasUnknownFields) {
```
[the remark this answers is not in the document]
sigurdm: Good question!
I think we could change the _hasField semantics also.
Done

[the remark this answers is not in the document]
szakarias: Since we are now changing semantics to include unknowns, I agree that it is inconsistent if we don't also update  `_hasField`. Good catch!

protobuf/lib/src/protobuf/extension_field_set.dart
```diff
@@ -21,7 +21,10 @@ class _ExtensionFieldSet {
     // I think this was originally here for repeated extensions.
     _addInfoUnchecked(fi);
     var value = _getFieldOrNull(fi);
-    if (value == null) return fi.makeDefault();
+    if (value == null) {
```
szakarias: I think this was supposed to be part of another CL

protoc_plugin/test/oneof_test.dart
```diff
@@ -183,6 +183,14 @@ void main() {
     expect(foo.hasIndex(), true);
     expect(foo.index, Bar());
   });
+
+  test('clone', () {
```
szakarias: Remove?

sigurdm: Done

protobuf/lib/src/protobuf/generated_message.dart
mkustermann: Maybe my comment was not clear, what I meant is the following:

We seem to have an **invariant** now
  * there is at most one of the `oneof` cases is set (either a) it's set in field set b) it's set in unknown field set c) none are set)
  * whenever we add a new field value, we are guaranteed to clear the old one if one exists.


This invariant is violated because of this line:
```dart
    if (other._hasUnknownFields) {
      _ensureUnknownFields().mergeFromUnknownFieldSet(other._unknownFields);
    }
```

We might have already a oneof candidate in the fieldset (or the unknown fieldset). Yet this code blindly copies `other._unknownFields`, which possibly has a new `oneof` value, but we **do not clear** the old one.

sigurdm: Ah - now I get it - thanks! I updated the merging of unknown fields to go field by field, and also update the one-ofs.

sigurdm:
I have also improved the _clearField methods handling of oneofs. It was doing the work twice, and only for known fields @szakarias Could you look over that?

protobuf/lib/src/protobuf/generated_message.dart
szakarias: Good catch, Martin. Thanks.

osa1:
@sigurdm it seems like the changes to `_FieldSet._hasField` are separate from the issues with `oneof` and can be merged separately, with the test you added. Should we do that?

PR #326, opened 2019-11-29
Add type signature of operator ==

bholst:
Otherwise our projects tests do not build. I have tried to provide a minimal example to
demonstrate this issue as we do not understand the problem, but unfortunately I failed doing this :/.

### DIFF
--- a/protobuf/lib/src/protobuf/pb_list.dart
+++ b/protobuf/lib/src/protobuf/pb_list.dart
@@ -168,7 +168,7 @@ abstract class PbListBase<E> extends ListBase<E> {
         check = _checkNotNull;
 
   @override
-  bool operator ==(other) =>
+  bool operator ==(dynamic other) =>
       (other is PbListBase) && _areListsEqual(other, this);
 
   @override


bholst:
@googlebot I signed it!

osa1 (commented):
I'm surprised that this change fixes downstream compile errors. I wonder if this was a bug in the Dart compiler when this issue was reported.

Here are my notes on this change. I'm new at Dart so please correct me if I'm wrong :-)

- `operator ==` argument type is `Object` (https://github.com/dart-lang/sdk/blob/43dedb589fd8fccb9e4a8773de2a3fc455c47dfe/sdk/lib/core/object.dart#L52)

- Normally when an argument type is unspecified the type is inferred as `dynamic`, but since this is a subclass overriding a method, the method type is inherited. So in this case `other` has type `Object`.

- Only two types that are more general than `Object` are `Object?` and `dynamic`.

- (Unrelated: The difference between `Object?` and `dynamic` cannot be observed in call sites: both accept any value. The differences between these two argument types are observed in the function body.)

- However, making argument type of `operator ==` `Object?` or `dynamic` should not make any difference, because this method is never called on `null` values, and the method body does not require the argument type to be `dynamic`. Example to the first point:
  ```dart
  class Test {
    @override
    bool operator ==(Object other) {
      print("`operator ==` called");
      return true;
    }
  }
  
  void test(Test? t2, Test? t2) {
    print(t2 == t2);
  }
  
  void main() {
    test(Test(), Test());
    test(Test(), null);
    test(null, Test());
    test(null, null);
  }
  ```
  Here only the first call calls `operator ==`. Changing `operator ==` argument type to `dynamic` or `Object?` does not make any difference.

**Conclusion:** This change should not make any downstream code that were not building before buildable. If this is really the case, then it should be a compiler bug and should be reported as such.

bholst:
@osa1 Thanks for the explanation. We do not use this fix anymore!

PR #332, opened 2019-12-17
Fix hashcode of bytes fields

sigurdm:


nichite:
Alternatively, the build is failing currently anyway. So from a different perspective it wouldn't matter if this change or lint fixes change comes first, in which case this LGTM.

sigurdm:
I think we should wait with fixing the lints until we have synced internal and internal.

PR #347, opened 2020-02-04
Fix wire format for negative int32 values

iinozemtsev:
According to https://developers.google.com/protocol-buffers/docs/encoding#signed-integers, negative int32 values should be encoded differently:

> If you use int32 or int64 as the type for a negative number, the resulting varint is always ten bytes long – it is, effectively, treated like a very large unsigned integer.

However CodedBufferWriter does not distinguish between negative and positive values, and this may cause problems when protos encoded by Dart are decoded by other proto implementations.

To update expected bytes in codec_test I used encoded bytes produced by C++ protobuf implementation.

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fix hashcode of bytes fields.
 * Fix issue with the `permissiveEnums` option to `mergeFromProto3Json`.
   The comparison did not work properly.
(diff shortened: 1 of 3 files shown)


mraleph:
Good catch!

iinozemtsev:
Also updated a change log. How do I get it merged?

sigurdm:
Thanks!

PR #348, opened 2020-02-04
Fix Travis build by using older lint settings

iinozemtsev:
pedantic 1.9.0 enforced 17 new lints. Constrain pedantic version to `<1.9.0`.

### DIFF
--- a/protobuf/pubspec.yaml
+++ b/protobuf/pubspec.yaml
@@ -15,4 +15,4 @@ dev_dependencies:
   test: '>=1.2.0'
   benchmark_harness: any
   js: any
-  pedantic: ^1.0.0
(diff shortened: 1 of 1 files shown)


jakemac53:
How much work is it to fix the lints instead?

We should file an issue to update to the latest lints again if we do pin to an older version for now.

iinozemtsev:
there's 590 lint violations, I'll check how much I can go through in an hour or so, I'll explicitly disable prefer_spread_collections though

jakemac53:
Ah if its that many just pinning an older version for now is probably fine

iinozemtsev:
https://github.com/dart-lang/protobuf/pull/350 fixes all violations

PR #349, opened 2020-02-04
benchmarks: update build_web_compilers and default to dart2js -O3

jakemac53:


jakemac53:
merging on red as this is a trivial change and travis is broken at head :sob:

PR #350, opened 2020-02-04
Fix lint violations

iinozemtsev:
After pedantic package was upgraded to 1.9.0, several new lints became enforced and broke Travis build. This PR fixes most violations and disables `prefer_collection_spreads` lint, as it seems to be incompatible with minimal SDK requirements.

### DIFF
--- a/protobuf/analysis_options.yaml
+++ b/protobuf/analysis_options.yaml
@@ -1,1 +1,5 @@
 include: package:pedantic/analysis_options.yaml
+
+linter:
+  rules:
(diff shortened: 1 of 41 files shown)


iinozemtsev:
no more lint violations, but some tests are failing because some autofixes removing local types also require explicit generic arguments, I'll try fixing those

davidmorgan:
Generally LG but I guess this should be reviewed by someone who knows the package. I see Sigurd  is on the reviewer list, that should work :)

Thanks.

PR #352, opened 2020-02-21
Add unnecessary_const to generated ignore_for_file

emadfarag:
The protoc plugin produces code with unnecessary consts. It's fine to keep the generated as is to work with older versions of Dart, but we should disable the linter for those error.

For example, generating dart from the gRPC example code yields a hello_world.pbjson.dart file with the following code:
```
const HelloRequest$json = const {
  '1': 'HelloRequest',
  '2': const [
    const {'1': 'name', '3': 1, '4': 1, '5': 9, '10': 'name'},
  ],
};
```

### DIFF
--- a/protoc_plugin/lib/file_generator.dart
+++ b/protoc_plugin/lib/file_generator.dart
@@ -546,7 +546,7 @@ class FileGenerator extends ProtobufContainer {
 //  source: ${descriptor.name}
 //
 // @dart = 2.3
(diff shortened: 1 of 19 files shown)


emadfarag:
@googlebot I signed it!

emadfarag:
How do I get this pushed on the master branch? (I have no write access).

sigurdm:
You need to update the golden test files.
The easiest way to do this is to delete them, and rerun the tests.
```
cd protoc_plugin
rm rm test/goldens/*
pub run test
```

PR #353, opened 2020-02-24
Add UTC flag to Timestamp's toDateTime()

emadfarag:
Calling toDateTime() on a google.protobuf.Timestamp time will always return a DateTime in UTC. This commit adds a flag to allow users to choose between UTC or local time.

### DIFF
--- a/protobuf/lib/src/protobuf/mixins/well_known.dart
+++ b/protobuf/lib/src/protobuf/mixins/well_known.dart
@@ -153,9 +153,9 @@ abstract class TimestampMixin {
   ///
   /// The result is in UTC time zone and has microsecond precision, as
   /// [DateTime] does not support nanosecond precision.
-  DateTime toDateTime() => DateTime.fromMicrosecondsSinceEpoch(
+  DateTime toDateTime({utc = true}) => DateTime.fromMicrosecondsSinceEpoch(
       seconds.toInt() * Duration.microsecondsPerSecond + nanos ~/ 1000,
-      isUtc: true);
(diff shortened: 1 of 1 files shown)


sigurdm:
I am somewhat hesitant to approve this.

Here is the documentation from Timestamp.proto:
```
// A Timestamp represents a point in time independent of any time zone or local
// calendar, encoded as a count of seconds and fractions of seconds at
// nanosecond resolution. The count is relative to an epoch at UTC midnight on
// January 1, 1970, in the proleptic Gregorian calendar which extends the
// Gregorian calendar backwards to year one.
```

To me it seems that a timestamp is always UTC.

Is something similar supported by Java/Python/C++ ....?

sigurdm:
If we were to move forward with this it would also need a test or two.

emadfarag:
A UNIX timestamp is always timezone-independent and that is what google.protobuf.Timestamp represents (number of nanoseconds since epoch). However, Dart's DateTime object also has a notion of timezone which affects how the object is printed, but it still doesn't change its (timezone-independent) timestamp.

So if I use this new flag to create two dart DateTime objects from the same Timestamp object that print:
"2020-02-24 12:00PM UTC" and "2020-02-24 5:00PM EST"

Then these two objects reference the same point in time and have the same timestamp, but just a different timezone component.

I can add test cases.

emadfarag:
Actually, it might be easier to just call .toLocal() on the resulting DateTime object instead of adding a flag, which would have the same effect.

PR #354, opened 2020-02-25
Permit unknown enum values when ignoreUnknownFields=true.

jackdreilly:
To be consistent with: https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L1925

This change allows adding additional values to an enum definition without parsing failures occurring on clients with older enum definitions.

### DIFF
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -227,8 +227,8 @@ abstract class GeneratedMessage {
   /// Well-known types and their special JSON encoding are supported.
   ///
   /// If [ignoreUnknownFields] is `false` (the default) an
-  /// [FormatException] is be thrown if an unknown field name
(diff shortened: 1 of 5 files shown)


jackdreilly:
@sigurdm Is there a process required for requesting a review? Apologies if simple patience is the procedure :)

sigurdm:
Sorry for the delay - no official process.
I think we should do this, especially given the reference to java.

sigurdm:
Sorry for the delay - no official process.
I think we should do this, especially given the reference to java.
I would like to hear the opinion from @nichite before landing.

sigurdm:
Sorry for the delay - no official process.
I think we should do this, especially given the reference to java.
I would like to hear the opinion from @nichite before landing.

nichite:
I think it sounds fine, and it would help compatibility across software versions.

CupNoodleFork:
I need this feature, when to release

PR #355, opened 2020-02-25
Enable hashCode memoization for frozen protos.

nichite:
If protos are frozen, and the `defaultFrozenMessageModificationHandler` is being used and has never been messed with, there should be no way to illegally mutate proto field sets. In this case, the hashCode should be stable, and we can memoize the value.

This should significantly speed up `hashCode` computations for large protos.


nichite:
Looks like the failures are just due to `--fatal_infos`, I can send out another change after to clean up all the violations

PR #356, opened 2020-02-27
Fix analyzer lints

sigurdm:
To be compliant with pedantic 1.9

Also reorganizes the protobuf/benchmarks to be actually runable.


sigurdm:
This was already done by https://github.com/dart-lang/protobuf/pull/350/files

PR #357, opened 2020-02-28
Draft of typed getExtension

sigurdm:
cc: @nichite

### DIFF
--- a/protobuf/lib/src/protobuf/field_info.dart
+++ b/protobuf/lib/src/protobuf/field_info.dart
@@ -5,7 +5,8 @@
 part of protobuf;
 
 /// An object representing a protobuf message field.
-class FieldInfo<T> {
+/// For a repeated field S is List<T> for a non-repeated field S=T.
+class FieldInfo<T, U> {
   FrozenPbList<T> _emptyList;
 
   /// Name of this field as the `json_name` reported by protoc.
@@ -24,9 +25,9 @@ class FieldInfo<T> {
 
   // Constructs the default value of a field.
   // (Only used for repeated fields where check is null.)
-  final MakeDefaultFunc makeDefault;
+  final U Function() makeDefault;
 
-  // Creates an empty message or group when decoding a message.
+  // Creates an empty message when decoding a message.
   // Not used for other types.
   // see GeneratedMessage._getEmptyMessage
   final CreateBuilderFunc subBuilder;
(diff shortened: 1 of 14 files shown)


protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -5,7 +5,8 @@
 part of protobuf;
 
 /// An object representing a protobuf message field.
-class FieldInfo<T> {
+/// For a repeated field S is List<T> for a non-repeated field S=T.
```
nichite: I see some type parameters S and U throughout the PR. Are they meant to be separate types? The comment above mentions S, but the parameter is U here. I am guessing either is meant to represent the List<T> as mentioned

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -439,7 +439,7 @@ class _FieldSet {
   List<T> _$getList<T>(int index) {
     var value = _values[index];
     if (value != null) return value as List<T>;
-    return _getDefaultList<T>(_nonExtensionInfoByIndex(index));
+    return _getDefaultList<T, dynamic>(_nonExtensionInfoByIndex(index));
```
nichite: Why would this be `dynamic` and not `List<T>` here?

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -236,7 +236,7 @@ class _FieldSet {
     return value;
   }
 
-  List<T> _getDefaultList<T>(FieldInfo<T> fi) {
+  List<T> _getDefaultList<T, U>(FieldInfo<T, U> fi) {
```
nichite: Is the new parameter plumbed through anywhere from here?

protobuf/lib/src/protobuf/field_info.dart
sigurdm: Sorry - I did a rename, but forgot to rename the comment.

fzyzcjy:
👀

osa1:
Do you remember what problem this is solving @sigurdm? PR description doesn't refer to any issues.

sigurdm:
> Do you remember what problem this is solving @sigurdm? PR description doesn't refer to any issues.

Currently you get back a `dynamic` from `GeneratedMessage.getExtension()`. But we have an `Extension` object that can carry the type of the extension. This would lead to stronger typing. Every `dynamic` ends up including a cast.

I don't think I'll find the time to pursue this - feel free to close.

PR #358, opened 2020-02-28
Properly type generated proto extensions.

nichite:
Sync from internal repo--by removing the unnecessary and unparameterized `Extension` type on the LHS of the generated extension fields, the proper type should be inferred by Dart type system.


osa1:
This change was merged with #401.

PR #367, opened 2020-05-11
Add map for resolving of type by enum for oneof

yanivshaked:
What was added: A new static const map in class that is built on 'oneof'
Why was added: This map allows to resolve this object to the specific type it holds: using whichSubtype() as the map index, you get the specific class type. 

Example: Very useful for implementation of an Api Bus which can dispatch messages to callbacks according to the message type:

```

class ApiBus {
  Map _callbackByMessageType = new Map<Type, List<Function>>();

  void publish(ApiObj msg) {
    var type = ApiObj.TypeByApiObj_Subtype[msg.whichSubtype()];

    var callbacks = _callbackByMessageType[type];
    if (null == callbacks) {
      return;
    }
    for (int c = 0; c < callbacks.length; c++) {
      callbacks[c](msg);
    }
  }

  void subscribe<T>(void callback(ApiObj message)) {
    if (null == _callbackByMessageType[T]) {
      _callbackByMessageType[T] = List<Function>();
    }
    _callbackByMessageType[T].add(callback);
  }
}

```

### DIFF
--- a/protoc_plugin/lib/message_generator.dart
+++ b/protoc_plugin/lib/message_generator.dart
@@ -314,6 +314,20 @@ class MessageGenerator extends ProtobufContainer {
           out.println('0 : ${oneof.oneofEnumName}.notSet');
         });
       }
+
+      for (OneofNames oneof in _oneofNames) {
+        out.addBlock(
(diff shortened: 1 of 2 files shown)


yanivshaked:
@googlebot I signed it!

yanivshaked:
Note to myself: need to create a new pull request aligned with the latest version of the plugin

yanivshaked:
Created new PR with relates to latest version, closing this one

PR #381, opened 2020-06-17
message_generator.dart: generate setters for repeated fields

franklinhu:
## Summary
When generating Dart code for protobufs that include repeated fields, `message_generator.dart` only generates the getters. This appears to be an oversight in the branching logic; there are a bunch of sanity checks that apply to repeated fields, but it otherwise skips all setter generation.

## Motivation
I'd like to use generated protobuf code with a proto that contains a repeated field

## Testing
I couldn't find any docs on how to test this locally, so going to rely on Travis CI

### DIFF
--- a/protoc_plugin/lib/message_generator.dart
+++ b/protoc_plugin/lib/message_generator.dart
@@ -477,73 +477,73 @@ class MessageGenerator extends ProtobufContainer {
         throw 'Field ${field.fullName} cannot override '
             '${names.clearMethodName}() because it is repeated.';
       }
+    }
+
+    var fastSetter = field.baseType.setter;
+    _emitDeprecatedIf(field.isDeprecated, out);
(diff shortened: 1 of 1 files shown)


franklinhu:
@googlebot I signed it!

franklinhu:
Actually, looks like `pub run test` in the `protoc_plugin` dir seems to do the trick. I'll re-open once I get something working

hypnoglow:
Hi,

What is the status of this issue? 

Is there a possibility to generate setters for repeated fields in current version?

Thanks

sigurdm:
We don't generate setters for repeated fields. Instead use methods of the returned list to modify it.

This has been discussed before: https://github.com/dart-lang/protobuf/issues/8

sigurdm:
Closing as duplicate

PR #385, opened 2020-07-13
Optimize field set for oneof case

yanivshaked:
The current implementation initializes an empty List with n fields for a one-of case.
This means, that while one of of the elements in the list will actually be populated, the entire list is initialized taking n times the amount of memory than required.
The suggested change uses a Map instead of a List for the one-of case, reducing amount of memory drastically if one-of has many types to choose from (which is mostly the case).

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -52,10 +52,12 @@ class _FieldSet {
   final BuilderInfo _meta;
   final EventPlugin _eventPlugin;
 
-  /// The value of each non-extension field in a fixed-length array.
+  /// The value of each non-extension field in a dynamic type:
+  /// - Map<int, dynamic> for the case for one-of, since no need to allocate such long list
+  /// - List (fixed-length array) for non one-of cases
   /// The index of a field can be found in [FieldInfo.index].
   /// A null entry indicates that the field has no value.
-  final List _values;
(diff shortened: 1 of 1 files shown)


osa1:
The problem is we represent `oneof`s as message even though they have nothing in common.

We should have a separate value class for `oneof`s (similar to `ProtobufEnum`, which I wish we called `PbEnum` for consistency with `PbList` and `PbMap`) instead of optimizing message code (`_FieldSet`) to support this use case.

Message values are like product types (tuples, records, structs) and they're represented as `GeneratedMessage`.

Enum values are just integers with a mapping from the integers to names, and represented as `ProtobufEnum`.

Oneofs are sums (e.g. enhanced enums in Dart), and we don't have a way to represent sums efficiently currently. Instead we try to use `GeneratedMessage` which as demonstrated in this PR is not a good idea as it wastes `O(n)` space where `n` is the number of variants of the `oneof`.

sigurdm:
> Oneofs are sums (e.g. enhanced enums in Dart), and we don't have a way to represent sums efficiently currently. Instead we try to use GeneratedMessage which as demonstrated in this PR is not a good idea as it wastes O(n) space where n is the number of variants of the oneof.

I guess generating a class with a single dynamic field would be the easiest solution here without proper language support for sum-types.

We could generate the appropriate getters that will do the cast....

PR #391, opened 2020-08-06
Add the test for fixed32 int that could be negative

mightyvoice:
Fix that fixed32 int could be negative and then convert it to the corresponding unsigned 32bit int.


iinozemtsev:
Could you sync this PR and also update changelog?

nichite (approved):
LGTM

PR #393, opened 2020-08-10
Add changes stream to GeneratedMessage

iinozemtsev:
An actual implementation should be provided via mixins.

This is a port of an internal change.


iinozemtsev:
the use-case is too specific and should be rather refactored internally

PR #394, opened 2020-08-10
Extension methods to replace `copyWith` and `clone`

iinozemtsev:
Extension methods `GeneratedMessage.rebuild` and `GeneratedMessage.deepCopy` replacing `copyWith` and `clone`.

This is a port of an internal change.

PR #397, opened 2020-08-20
Remove unnecessary dart:async import.

MichaelRFairhurst:
I'm trying to enforce this inside google3, and so it would be nice to be
able to land this.

This requires Dart 2.1, so this is technically a breaking change. If
that's not OK, then I can abandon this PR and make an exception inside
google3 instead.

Note that Dart 2.1 has been out for quite a long time now.

### DIFF
--- a/protobuf/lib/protobuf.dart
+++ b/protobuf/lib/protobuf.dart
@@ -4,7 +4,6 @@
 
 library protobuf;
 
-import 'dart:async' show Future;
 import 'dart:collection' show ListBase, MapBase;
 import 'dart:convert'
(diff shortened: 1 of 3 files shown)


iinozemtsev:
Hi Michael, thanks a lot! I don't think Dart 2.1 is an issue, I have a pending PR which requires Dart 2.7 (#394), but we'd need to change package version in pubspecs and add something to a CHANGELOG.  I plan to start active work on syncing internal/external next week. Could you wait a bit until I merge #394 and then update this PR? Feel free to send me equivalent google3 change any time.

iinozemtsev:
Michael, could you sync your PR? Minimum Dart SDK version is now 2.7

PR #402, opened 2020-08-28
Fix escaping of string default values.

iinozemtsev:
Port of an internal change.

This PR is built on top of #401, once that one is submitted, I'll update this one


iinozemtsev:
Please review changes from the last commit only

PR #421, opened 2020-10-22
add fromJsonObject method so that generated code could parse Proto3 json

yangyanzhan:
Generated code could only parse json with integer keys.

This commit adds a fromJsonObject to the generated code so that end user could use fromJsonObject to parse json with string keys (i.e. field names in ProtoBuf Message).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Protobuf support for dart
+# A Better ProtoBuf Library For Dart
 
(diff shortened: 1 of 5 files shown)


yangyanzhan:
@googlebot I signed it!

yangyanzhan:
When ProtoBuf data is transmitted as a Json string, we'd like it to be as readable as possible. There are several benefits for improved readability for Json string. The major benefit is that we can identify the meaning of the field at first glance thus significantly reducing the cost for debug.
With numbers as keys, the readability of the json string is very low since the meaning of every field could only be revealed by looking up the proto file.

PR #439, opened 2020-11-12
Set protoc_plugin version to 20.0.0-nullsafety.0

iinozemtsev:
Even though protoc_plugin code is not migrated to null safety, we
still need to bump major version to avoid surprises for
packages using protoc_plugin with `^19.0.0` version spec.

Also add dependency override to fix api_benchmark.


sigurdm (approved):
LGTM

jakemac53 (approved):

PR #440, opened 2020-11-13
Regenerate bootstrap protos with previous version of protoc_plugin

iinozemtsev:
As protoc_plugin is not null-safe yet, I cannot publish it to pub with
mixed sources, so downgrading bootstrap protos until protoc_plugin is
migrated to null safety


mraleph (approved):

sigurdm:
LGTM

Another option would have been to add a 
```
// @dart=2.11
```
language marker in all protoc_plugin's own files and up the sdk constraint in protoc_plugin/pubspec.yaml.

PR #442, opened 2020-11-16
Added constructors for messages

### LINKED ISSUE #155: named arguments to set fields in a constructor
I am not sure if this feature request is worth it, but I wanted to share the idea, so here it is:

What if we take generated messages and make a constructor that takes as arguments the actual fields from the proto buff? for example:

``` proto
message A {
  required someField = 1;
  optional otherField = 2;
}
```

generates:

``` dart
class A extends GeneratedMessage {
  A.filled(this.someField, {this.otherField});
}
```

I just often see patterns where developers have all the data they need to fill in, and end up writing code like:

``` dart
  new A()
      ..someField = a
      ..otherField  = b
```

instead of:

``` dart
  new A(a, otherField: b)
```

Kavantix:
Fixes #155

### DIFF
--- a/protoc_plugin/lib/base_type.dart
+++ b/protoc_plugin/lib/base_type.dart
@@ -55,6 +55,9 @@ class BaseType {
   String getRepeatedDartType(FileGenerator fileGen) =>
       "$_coreImportPrefix.List<${getDartType(fileGen)}>";
 
+  String getRepeatedDartTypeIterable(FileGenerator fileGen) =>
+      "$_coreImportPrefix.Iterable<${getDartType(fileGen)}>";
+
   factory BaseType(FieldDescriptorProto field, GenerationContext ctx) {
     String constSuffix;
 
--- a/protoc_plugin/lib/message_generator.dart
+++ b/protoc_plugin/lib/message_generator.dart
@@ -360,7 +360,44 @@ class MessageGenerator extends ProtobufContainer {
       out.printlnAnnotated('${classname}._() : super();', [
         NamedLocation(name: classname, fieldPathSegment: fieldPath, start: 0)
       ]);
-      out.println('factory ${classname}() => create();');
+      out.print('factory $classname(');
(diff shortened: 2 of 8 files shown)


sigurdm (changes requested):
This is a nice addition, greatly improving the user-experience.

However many tests are still failing. Try to look into the travis logs...

sigurdm (commented):
@rakudrama do you have a feeling how this would affect size of the generated code?
My guess would be that the constructor would be inlined everywhere - maybe we would have to try it.

Kavantix:
The tests also fail on the current master so not sure what's my part in this

I'm also considering adding the same thing to the copyWith (and then make the current function parameter optional position)
Would that be a good addition for this pr?

sigurdm:
I just ran all tests on master and they don't fail.

You'll have to add a line:
```
dependency_overrides: {protobuf: {path: ../protobuf}}
```

to protoc_plugin/pubspec.yaml

Followed by `pub get`

to test it locally.

Kavantix:
Ah i see, using that there are still tests failing on master though.
But if I only run the tests that start with `message` then they indeed pass on master and the golden file was just not updated yet so I did.

sigurdm:
What tests are failing on master?

Kavantix:
Nvm not all needed files were generated.
Fixed the last error now

Kavantix:
Hmm some more golden files are failing, I'll update these later today, tbc

Kavantix:
Should be ready to merge now

sigurdm (approved):

osa1:
Port to internal in cl/449707130.

rakudrama:
> @rakudrama do you have a feeling how this would affect size of the generated code? My guess would be that the constructor would be inlined everywhere - maybe we would have to try it.

I expect this change to be quite detrimental to code size.
I'm sad I didn't notice your question earlier.
I would have suggested not doing this.

PR #444, opened 2020-11-19
Support optional proto3 fields

iinozemtsev:
Also update plugin.proto and descriptor.proto to the latest version


sigurdm (approved):
LGTM

PR #452, opened 2020-12-14
Allow dart protos to be sent through sendports (and therefore to other isolates)

mkustermann:
Decoded protos in Dart are represented in memory as instances of
`GeneratedMessage` subclasses. The `GeneratedMessage` stores the field
state in a `_FieldSet` and `UnknownFieldSet`.

The `_FieldSet._meta` field points to a `BuilderInfo` object, which
represents the metadata of the proto message.

Sending this reference to `BuilderInfo` makes it hard to send dart
protos to other isolates, because

  a) `BuilderInfo` hangs on to non-static closures, which cannot
      be sent to to other isolates atm

  b) Sending a transitive copy would copy the `BuilderInfo` object (and
     it's transitive closure). That can cause big waste of memory
     since an isolate would have this metadata possibly many times.

To avoid these two issues, this CL does the following:

  * We replace `_FieldSet._meta` with a getter that accesses the
    `BuilderInfo` via the `_FieldSet._message.info_` getter. This getter
    is implemented in `GeneratedMessage` subclasses and reads from a
    static field.

    => We effectively replace a load of the `_meta` field with load +
       interface call + static field load.

  * To avoid performance regressions we avoid accessing the new
    `_FieldSet._meta` getter in the hot decoding loops. Instead we
    obtain this metadata once and pass it down to various methods that
    are part of decoding.

  * For similar reasons, we remove the `PbMap._entryBuilderInfo`
    field - which also refers to a `BuilderInfo`. Instead we obtain the
    neccessary map-specific `BuilderInfo` in the main decoder loop.

This CL has to remove the 2 year long deprecated `PbMap.add()` function.

Closes https://github.com/dart-lang/protobuf/issues/167


mkustermann:
The corresponding g3 change is in [cl/347370918](http://cl/347370918). Let's do the review in g3 and I'll port comments here as well.

mkustermann:
@iinozemtsev Could you take a look at this PR (to avoid divergence I will wait with landing the g3 change until this PR was also reviewed, so they can both land at the same time)

iinozemtsev (approved):

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -192,6 +192,9 @@ String _unCamelCase(String name) {
 }
 
 class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
+  static dynamic throwYouShouldNotCallThis() =>
+      throw 'you should not call this';
```
sigurdm, line 196: This could be a StateError?

sigurdm (approved):

protobuf/lib/src/protobuf/field_info.dart
mkustermann: Done.

mkustermann:
Thank you, @sigurdm @iinozemtsev !

PR #453, opened 2020-12-15
Restore possibility to use reflective API to create default field values using getField()

mkustermann:
We do not actually have to prevent/throw when trying to create a default
value for a PbMap field.

Initially when working on removing BuilderInfo references it was
necessary to remove default creation of map fields due to not
having the [BuilderInfo] object for the map which was required for the
PbMap constructor.

Though a later iteration of that CL removed the 3rd parameter to PbMap
(or rather made that parameter optional and ignored):

```diff
  -  PbMap(this.keyFieldType, this.valueFieldType, this._entryBuilderInfo)
  +  // The provided [info] will be ignored.
  +  PbMap(this.keyFieldType, this.valueFieldType, [BuilderInfo? info])
```

So we can restore `defaultOrMake()`, which fixes the reflective API usage.

### DIFF
--- a/protobuf/lib/src/protobuf/field_info.dart
+++ b/protobuf/lib/src/protobuf/field_info.dart
@@ -192,9 +192,6 @@ String _unCamelCase(String name) {
 }
 
 class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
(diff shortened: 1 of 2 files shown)


iinozemtsev (approved):

mkustermann:
Thanks, @iinozemtsev

PR #458, opened 2021-01-08
Fix copyWith's return type.

domesticmouse:
This is an attempt to fix an issue I am seeing in `dart-services`: https://github.com/dart-lang/dart-pad/issues/1712


sigurdm (commented):
This looks good.
Could you write a test exposing the fix?

domesticmouse:
@sigurdm as the fix is add yet another typecast I'm curious what sort of test you would like over this code? 

I'm also conscious that I am patching the usage of a call point that is slated to go away, as indicated by the `deprecated_member_use` on that line, so I'm figuring this code will get deleted sooner or later.

sigurdm:
Just a test that fails before (probably at compile time) because the return type is wrong - and succeeds after the fix.

In other words: a test that the issue you experienced was fixed, and ensuring we don't regress.

domesticmouse:
Quick question, is there a way of generating code and then rolling the pedantic linters over it?

parlough (approved):

sigurdm:
Hmm - the linter should be run on generated code from the travis script - I don't think you need to do it programatically.

domesticmouse:
How about I add the `pedantic` package to `protobuf/analysis_options.yaml`? That should be enough to prevent this regressing.

sigurdm:
> Why wasn't this caught by the linter previously?

Good question.

Looking into this, the difference is to be that dartpad uses:
```
analyzer:
  strong-mode:
    implicit-casts: false
```

sigurdm:
I made the change to protoc_plugin entirely.

Question is if we should land this in master, or only on null_safety as this is a lint-only fix.

@iinozemtsev WDYT?

iinozemtsev:
Sorry, I somehow did not see this earlier. I think merging it into master is fine, I'm going to merge master into null safety in the next few days (I think since github squashes PRs before merge this won't be a real merge, rather a cherry-picking missing commits from master)

domesticmouse:
@sigurdm do you need me to do anything w.r.t fixing up this PR such that all tests pass?

PR #460, opened 2021-01-27
Migrate from Travis to GitHub Actions

iinozemtsev:
It seems that the last completed Travis build was 6 days ago, and
travis-ci.org is going to be shut down anyway.

As with travis, api_benchmark/query_benchmark have a single stage
which generates protos checks format and runs analyzer.

For protobuf and protoc_plugin I've split a single format_analyze_test
stage into three separate stages for a better control of OSes and Dart
SDK versions.

`protobuf` tests run on windows, linux, and osx, but `protoc_plugin`
tests run on linux only (we'd need a way to make `make protos` work on
osx and windows to enable tests).

Fixes #459

PR #406, opened 2020-08-31
Emit binary coded descriptors from dart protoc_plugin

iinozemtsev:
Port of an internal change

These descriptors contain extensions, and thus can be used to reflect over the options given to the descriptor

Fixes: https://github.com/dart-lang/protobuf/issues/341

PR #463, opened 2021-01-28
Merge master into null safety

iinozemtsev:
This brings the following changes from master branch:

- Support optional proto3 fields (#444)
- Fix handling of oneofs in dart proto3 json parsing. (#447)
- Use @dart = 2.7 in generated files (#455)
- Migrate from Travis to GitHub Actions (#460)
- Emit binary coded descriptors from dart protoc_plugin. (#406)

Github Actions config modified to run only on dev sdk. Also modify analyzer flags to remove `--fatal-infos`, there's one mixin-related violation:

```
hint • The library 'package:protoc_plugin/testing/mixins.dart' is legacy, and should not be imported into a null safe library. • out/protos/mixins.pb.dart:12:8 • import_of_legacy_library_into_null_safe
```

Also slightly speeding up a build by removing `make protos` step from `format` stage: generated protos are not checked in and formatted anyway, but generating and checking format takes 3-4 minutes.


protoc_plugin/lib/src/dart_options.pb.dart
```diff
@@ -2,7 +2,7 @@
 //  Generated code. Do not modify.
 //  source: dart_options.proto
 //
-// @dart = 2.3
+// @dart = 2.7
```
sigurdm, line 5: Hmm - should this not be 2.12 for the null-safety branch?

sigurdm (approved):
LGTM - but perhaps you need to regenerate the descriptors...

iinozemtsev:
No, see https://github.com/dart-lang/protobuf/pull/440, until protoc_plugin
is null safe we can't really migrate those. I'll send a separate PR to
change version on protoc sources to 2.11 and then regenerate bootstrap
protos, WDYT?

sigurdm:
> until protoc_plugin is null safe we can't really migrate those.

Ah of course!

> I'll send a separate PR to
change version on protoc sources to 2.11 and then regenerate bootstrap
protos, WDYT?

Sounds good!

PR #464, opened 2021-01-29
Use null-safe bootstrap protos

iinozemtsev:
- Upgrade dependencies to null safety version where possible
- Remove accidentally committed dependency_override
- Regenerate bootstrap protos with null-safe version of protoc_plugin
- Bump sdk to 2.11 and add `@dart=2.11` to all own sources of protoc_plugin
- Further speed up build by not checking format of query_benchmark
  generated protos and by testing api_benchmark and query_benchmark in
  parallel
- Merge Add badges to readme (#462)

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -279,16 +279,16 @@ jobs:
       - job_003
       - job_004
   job_009:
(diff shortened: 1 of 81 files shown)


sigurdm (approved):
LGTM

PR #466, opened 2021-02-01
Generate null-safe .pbgrpc files

iinozemtsev:
Now as grpc-dart migrated to null safety, we can generate null-safe .pbgrpc files


davidmorgan:
Looks good, thanks!

PR #478, opened 2021-02-22
Fix proto deserialization issue for repeated and map enum value field…

mightyvoice:
Fix proto deserialization issue for repeated and map enum value fields where the enum value is unknown.
This is first PR to update the protobuf lib first and add more to protoc_plugin, otherwise will fail protoc_plugin since cannot use the most recent protobuf lib.


nichite (approved):
LGTM

Issue #485, opened 2021-03-02
Unable to use GRPC service RPC call "New"


daniel-sullivan:
When attempting to compile Dart libraries for our existing platform, we came across a strange issue with one of the repositories where we have a definition similar to:

```proto
service Account {
  rpc Get(api.base.GetRequest) returns (api.account.Model);
  rpc Search(api.base.SearchRequest) returns (stream api.account.SearchResponse);
  rpc New(api.account.Model) returns (api.account.Model);
  rpc Update(api.account.Model) returns (api.account.Model);
  rpc Destroy(api.base.DestroyRequest) returns (google.protobuf.Empty);
}
```

We received an output similar to the below examples with the main error seeming to be "Constructors can't have type parameters.". Unfortunately Google didn't provide any suggestions.

Doing a bit of digging through commenting out lines, we discovered that the issue was being caused by this line:
```
  rpc New(api.account.Model) returns (api.account.Model);
```

Further reading into Dart suggests that with "new" being a reserved term in Dart, us using it as the rpc function name is causing a conflict.

This was verified by creating two files:

test_create.proto
```proto
syntax = "proto3";

message SimpleReq {
  string test = 1;
}

message SimpleResponse {
  string test = 1;
}

service TestService {
  rpc Create(SimpleReq) returns (SimpleResponse);
}
```

```
❯ protoc --dart_out=grpc:. test_create.proto
< successful return > 
```

test_new.proto
```proto
syntax = "proto3";

message SimpleReq {
  string test = 1;
}

message SimpleResponse {
  string test = 1;
}

service TestService {
  rpc New(SimpleReq) returns (SimpleResponse);
}
```

```
❯ protoc --dart_out=grpc:. test_new.proto
Unhandled exception:
Could not format because the source could not be parsed:

line 50, column 36: A function body must be provided.
   ╷
50 │   $async.Future<$0.SimpleResponse> new($grpc.ServiceCall call, $0.SimpleReq request);
   │                                    ^^^
   ╵
line 28, column 24: Expected to find '>'.
   ╷
28 │   $grpc.ResponseFuture<$0.SimpleResponse> new($0.SimpleReq request, {$grpc.CallOptions options}) {
   │                        ^^
   ╵
line 28, column 23: Constructors can't have type parameters.
   ╷
28 │   $grpc.ResponseFuture<$0.SimpleResponse> new($0.SimpleReq request, {$grpc.CallOptions options}) {
[...]

PR #487, opened 2021-03-02
Disambiguate names of service methods in grpc output

sigurdm:
Fixes: https://github.com/dart-lang/protobuf/issues/485

### DIFF
--- a/protoc_plugin/lib/names.dart
+++ b/protoc_plugin/lib/names.dart
@@ -487,31 +487,31 @@ bool _isDartFieldName(name) => name.startsWith(_dartFieldNameExpr);
 final _dartFieldNameExpr = RegExp(r'^[a-z]\w+$');
 
 /// Names that would collide as top-level identifiers.
-final List<String> forbiddenTopLevelNames = <String>[
+const forbiddenTopLevelNames = <String>[
   'List',
   'Function',
   'Map',
   ..._dartReservedWords,
 ];
 
-final List<String> reservedMemberNames = <String>[
+const reservedMemberNames = <String>[
   ..._dartReservedWords,
   ...GeneratedMessage_reservedNames,
   ..._generatedMessageNames
 ];
 
-final List<String> forbiddenExtensionNames = <String>[
+const forbiddenExtensionNames = <String>[
   ..._dartReservedWords,
   ...GeneratedMessage_reservedNames,
   ..._generatedMessageNames
 ];
 
-final List<String> serviceReservedMemberNames = <String>[
+const serviceReservedMemberNames = <String>[
   ..._dartReservedWords,
   ..._serviceNames,
 ];
 
-final List<String> _serviceNames = <String>[
+const _serviceNames = <String>[
   // From `client`.
   r'$createCall',
   r'$createUnaryCall',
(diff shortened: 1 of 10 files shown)


protoc_plugin/lib/names.dart
```diff
@@ -506,6 +506,23 @@ final List<String> forbiddenExtensionNames = <String>[
   ..._generatedMessageNames
 ];
 
+final List<String> serviceReservedMemberNames = <String>[
+  ..._dartReservedWords,
+  ..._serviceNames,
+];
+
+final List<String> _serviceNames = <String>[
```
nichite, line 514: Can this be `const`? Also, do we need the explicit type on the left-hand side?

protoc_plugin/test/goldens/grpc_service.pbgrpc
```diff
@@ -30,6 +30,18 @@ class TestClient extends $grpc.Client {
       '/Test/Bidirectional',
       ($0.Input value) => value.writeToBuffer(),
       ($core.List<$core.int> value) => $0.Output.fromBuffer(value));
+  static final _$new_ = $grpc.ClientMethod<$0.Input, $0.Output>(
```
nichite, line 33: One thing I was curious of, is it possible to have a name collision with the library import aliases? i.e. do we also need to add `r'$grpc'`, `r'$async'` etc to the reserved words?

nichite (commented):
Mostly LGTM

protoc_plugin/test/goldens/grpc_service.pbgrpc
sigurdm: $ cannot occur in a protobuf identifier, (https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#identifiers) - I also tried it out and the protoc compiler gave a syntax error.

I believe that is why we chose to use $ in the first place for these names.

That means we could actually remove all the reserved words with $!

I won't do that in this PR to keep the scope smaller.

protoc_plugin/lib/names.dart
sigurdm: yes!

Done - also in surrounding code.

protoc_plugin/test/goldens/grpc_service.pbgrpc
sigurdm: (I removed it from those added).

sigurdm:
I realized we also needed disambiguation for the .pbserver.dart code, and added that - this will require a bit of integration work when moving into google3.

@nichite @iinozemtsev do you know if we have any real uses of the pbserver.dart files internally? Searching for 'import.*.pbserver.dart' gives no real results. Maybe it is time to retire that code, and have the grpc code be generated by default?

nichite:
I agree with retiring the .pbserver files, I don't see any internally. I was thinking the same thing about the generated RPC client stubs that protobuf puts out that aren't functional. For those, there are one or two subclassed internally, but I was thinking they both could just be replaced by gRPC.

nichite (approved):
LGTM

osa1:
Why was this PR closed @sigurdm @kevmoo ? Might be good to revive this as it fixes an issue (#485).

kevmoo:
Possible I was doing branch cleanup and didn't realize this was NOT the null safety migration branch.

sigurdm:
> Why was this PR closed @sigurdm @kevmoo ? Might be good to revive this as it fixes an issue.

Yeah good question, we should try and revive it. Not trivial without the branch...

sigurdm:
Reconstructed in https://github.com/google/protobuf.dart/pull/625

PR #492, opened 2021-03-20
Fix comparison of empty lists from frozen messages.

LorenVS:
The existing implementation returns an unmodifiable list, which does not obey the value-comparison semantics expected for protobuf lists. Returning a `FrozenPbList` instead resolves this.

### DIFF
--- a/protobuf/test/list_equality_test.dart
+++ b/protobuf/test/list_equality_test.dart
@@ -1,0 +1,52 @@
+// Test for ensuring that protobuf lists compare using value semantics.
+library list_equality_test;
+
+import 'package:test/test.dart';
+
+import 'mock_util.dart' show T;
+
+void main() {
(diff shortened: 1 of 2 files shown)


protobuf/test/list_equality_test.dart
```diff
@@ -0,0 +1,52 @@
...
+import 'mock_util.dart' show T;
+
+void main() {
+  test('empty lists compare as equal', () {
+    final first = T();
+    final second = T();
+    expect(first.int32s == second.int32s, isTrue);
+  });
+
+  test('empty frozen lists compare as equal', () {
```
LorenVS, line 15: The rest of the test cases are for completeness. This is the test that was failing before this change

nichite (approved):
LGTM

PR #494, opened 2021-03-25
Fix the FrozenPbList type returned for absent fields.

LorenVS:
The `const []` in the previous iteration won't have the correct generic type parameter for the field. Using a non-const list allows the inclusion of an explicit type parameter.


osa1 (changes requested):
Thanks for the PR. Since this fixes a bug it needs a regression test. Could you add it please? If you don't have the time please explain the issue in a bit more detail (maybe show the code that doesn't work as expected) so I that I can add the regression test.

osa1:
Ping @LorenVS. I checked this code again, I think this change should not make any difference in comparisons as `FieldInfo.readonlyDefault` returns the same type of value for repeated fields: https://github.com/google/protobuf.dart/blob/c35d787726173380348e744085d02d34d56d3ec1/protobuf/lib/src/protobuf/field_info.dart#L138-L140

osa1:
Ping @LorenVS. I checked this code again, I think this change should not make any difference in comparisons as `FieldInfo.readonlyDefault` returns the same type of value for repeated fields: https://github.com/google/protobuf.dart/blob/c35d787726173380348e744085d02d34d56d3ec1/protobuf/lib/src/protobuf/field_info.dart#L138-L140

osa1:
Maintainer's note: "previous iteration" mentioned in the PR description is #492.

osa1:
While updating some internal code to master branch I found a case where without this PR we get a runtime error. I will add a test case in this PR if I'm allowed to edit, or in a separate PR, and then merge it.

osa1:
Closing in favor of #654.

PR #500, opened 2021-04-14
fixbug of typo - duplicated return statement

fzyzcjy:
Hi thanks for the lib! When looking at source code I see 

```
  void setField(int tagNumber, Object value) {
    _fieldSet._setField(tagNumber, value);
    return; // ignore: dead_code
    return; // ignore: dead_code
  }
```

Thus the second return seems to be useless. So I removed it.


kevmoo:
I swear there's a good reason this is here!

CC @rakudrama @sigmundch – remember?

sigmundch:
maybe and old-style way to avoid getting the code inlined? (dart2js used to not inline methods with multiple return statements, even if one of them was dead code, I believe)

@rakudrama may remember if that was the case.

In that case, the proper fix would be to remove the `return`s but also to add an annotation on the method: `@pragma('dart2js:noInline')`

rakudrama:
@sigmundch is correct.
This weird code must have been done for a reason. Unfortunately the 'blame' shows that this change was from up-streaming a change made in an internal repo, so there is no description.

If it is desired to proceed with the change, it must preserve the effect on how dart2js generate JavaScript code.
- The annotation `@pragma('dart2js:noInline')` should be added
- BOTH return statements should be removed - it is a `void` method so even one return statement is superfluous.

fzyzcjy:
@rakudrama @sigmundch done!

rakudrama (approved):
Thanks!

fzyzcjy:
You are welcome!

PR #514, opened 2021-06-08
Add grpc example to dart protoc docs.

aebrahim:
This is likely a much more used option than kythe_info, but was not included in the documentation, and took reading quite a bit of code to find. Thought I'd save time for the next person!


osa1:
Thanks @aebrahim!

PR #521, opened 2021-07-09
Widen fixnum range

robbecker-wf:
I'm hitting a problem getting pub to solve with Dart 2.13.4 on our codebase due to not being able to use fixnum 1.0.0. I've tracked it back to this repo that limits fixnum to `^0.10.9` but specifies a SDK range of `>=2.7.0 <3.0.0`. 
It seems like if you wanted to use this library with null safety, you'd want to be able to solve to fixnum 1.0.0 which is null-safe.

Indeed when running the migration tool it reports that the version of fixnum isn't nullsafe.
`pub outdated --mode=null-safety` shows
image

So, it seems like a good change in order support both older and newer SDK version. 

If this ends up merging, I would need this released as a patch version on the 1x line.

### DIFF
--- a/protobuf/pubspec.yaml
+++ b/protobuf/pubspec.yaml
@@ -9,7 +9,7 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
 
 dependencies:
-  fixnum: ^0.10.9
+  fixnum: '>=0.10.9 <2.0.0'
 
 dev_dependencies:
   test: '>=1.2.0'


robbecker-wf:
To get eyes, pinging @kevmoo and @sigurdm.

kevmoo:
@robbecker-wf – take a look at the failures here?

robbecker-wf:
Failing on dart analyze in the api_benchmark dir from something not being generated
```
error - lib/benchmark.dart:7:8 - Target of URI hasn't been generated: 'generated/benchmark.pb.dart'. Try running the generator that will generate the file referenced by the URI. - uri_has_not_been_generated
```
The other failures are formatting code under protoc_plugin.

I wouldn't expect this range widening to have changed either of these other subdirs. Is it possible this was failing before this PR or newer SDK or dart_style means code is being formatted differently?

kevmoo:
> Failing on dart analyze in the api_benchmark dir from something not being generated
> 
> ```
> error - lib/benchmark.dart:7:8 - Target of URI hasn't been generated: 'generated/benchmark.pb.dart'. Try running the generator that will generate the file referenced by the URI. - uri_has_not_been_generated
> ```
> 
> The other failures are formatting code under protoc_plugin.
> 
> I wouldn't expect this range widening to have changed either of these other subdirs. Is it possible this was failing before this PR or newer SDK or dart_style means code is being formatted differently?

Indeed! Hrm...

robbecker-wf:
Aaah I just discovered we can avoid needing to update this repo to widen the fixnum range if we remove dependencies on build_vm_compilers. So if you want to close this PR and forget it ever happened, I understand. On the otherhand, I may have uncovered something legit that other people haven't reported yet (null safety & failing CI).

robbecker-wf:
Closing this for now since there's no urgency on either side.

kevmoo:
Hey @robbecker-wf – FYI. There is a `null-safe` branch that needs to get merged in.

This has the v2 of things.

PR #525, opened 2021-07-23
null safe cleanup

kevmoo:


nichite (approved):
Great cleanup, thanks for doing this. LGTM

PR #526, opened 2021-07-23
protoc_plugin: drop version back to 19.3.2-dev

kevmoo:
19.3.1 was the last stable release in the not-null-safe chain


nichite (approved):
LGTM

PR #528, opened 2021-07-23
was it the proto native version?

kevmoo:


kevmoo:
nope! it's the SDK version, for some reason...

PR #473, opened 2021-02-11
Stable null safety release of protobuf

iinozemtsev:

PR #441, opened 2020-11-16
Added constructors for messages (null-safety)

Kavantix:

PR #435, opened 2020-11-11
Null safe package:protobuf and null safe code generation

iinozemtsev:
Makes `protobuf` null safe and makes `protoc_plugin` to generate null safe code. `protoc_plugin` itself is not yet migrated to null safety because it depends on `dart_style`. 

Also cannot make `api_benchmark` and `query_benchmark` working yet, as they transitively depend on `fixnum ^0.10.9`, which is incompatible with protobuf dependency on `fixnum ^1.0.0`.

PR #530, opened 2021-07-26
merge null safety

kevmoo:
- Null safe package:protobuf and null safe code generation (#435)
- Merge master into null_safety (#438)
- Set protoc_plugin version to 20.0.0-nullsafety.0 (#439)
- Regenerate bootstrap protos with previous version of protoc_plugin (#440)
- Allow dart protos to be sent through sendports (and therefore to other isolates) (#452)
- Restore possibility to use reflective API to create default field values using getField() (#453)
- Prepare query_benchmark/ for NNBD (#451)
- Let protoc_plugin depend on protobuf from null-safety and fix changelog. (#456)
- named optional arguments in constructors for messages (#441)
- Merge master into null safety (#463)
- Use null-safe bootstrap protos (#464)
- Generate null-safe .pbgrpc files (#466)
- Stable null safety release of protobuf (#473)
- Stable release of protoc_plugin for null safety (#474)
- Fix protoc_plugin publish errors and warnings (#475)
- Fix lints (#488)
- null safe cleanup (#525)

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -5,10 +5,10 @@
 part of protobuf;
 
 typedef FrozenMessageErrorHandler = void Function(String messageName,
-    [String methodName]);
+    [String? methodName]);
 
 void defaultFrozenMessageModificationHandler(String messageName,
-    [String methodName]) {
+    [String? methodName]) {
   if (methodName != null) {
     throw UnsupportedError(
         'Attempted to call $methodName on a read-only message ($messageName)');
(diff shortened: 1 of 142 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -74,12 +73,18 @@ class _FieldSet {
   /// code as an `int`.
   Object _frozenState = false;
 
+  /// The [BuilderInfo] for the [GeneratedMessage] this [_FieldSet] belongs to.
+  ///
+  /// WARNING: Avoid calling this for any performance critical code, instead
```
nichite, line 78: I see that `meta` is now plumbed through to several methods in accordance with this comment. Just curious: what exactly is the source of the slowdown? It's not immediately obvious to me why accessing the same reference one way would be less efficient than the other.

nichite (approved):
So exciting!

protobuf/lib/src/protobuf/field_set.dart
kevmoo: I have *zero* idea. CC @sigurdm

Issue #508, opened 2021-05-24
In null safety mode, an error is thrown when the value of the map<string,string> is string empty


innwin:
In null safety mode, an error is thrown when the value of the map<string,string> is string empty

![1621851174799](https://user-images.githubusercontent.com/6335833/119332871-c3834b80-bcbb-11eb-910c-654ad65a1a47.jpg)

![87511BED397FE585D3331D3587737962](https://user-images.githubusercontent.com/6335833/119332930-d7c74880-bcbb-11eb-929a-fe4c9c7835ec.jpg)

test proto: 

```proto
syntax = "proto3";

message MapTest {
  map<string,string> area = 5;
}
```

```dart
import 'dart:core' as $core;

import 'package:protobuf/protobuf.dart' as $pb;

class MapTest extends $pb.GeneratedMessage {
  static final $pb.BuilderInfo _i = $pb.BuilderInfo(const $core.bool.fromEnvironment('protobuf.omit_message_names') ? '' : 'MapTest', createEmptyInstance: create)
    ..m<$core.String, $core.String>(5, const $core.bool.fromEnvironment('protobuf.omit_field_names') ? '' : 'area', entryClassName: 'MapTest.AreaEntry', keyFieldType: $pb.PbFieldType.OS, valueFieldType: $pb.PbFieldType.OS)
    ..hasRequiredFields = false
  ;

  MapTest._() : super();
  factory MapTest({
    $core.Map<$core.String, $core.String>? area,
  }) {
    final _result = create();
    if (area != null) {
      _result.area.addAll(area);
    }
    return _result;
  }
  factory MapTest.fromBuffer($core.List<$core.int> i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) => create()..mergeFromBuffer(i, r);
  factory MapTest.fromJson($core.String i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) => create()..mergeFromJson(i, r);
  @$core.Deprecated(
  'Using this can add significant overhead to your binary. '
  'Use [GeneratedMessageGenericExtensions.deepCopy] instead. '
  'Will be removed in next major version')
  MapTest clone() => MapTest()..mergeFromMessage(this);
  @$core.Deprecated(
  'Using this can add significant overhead to your binary. '
  'Use [GeneratedMessageGenericExtensions.rebuild] instead. '
  'Will be removed in next major version')
[...]

PR #539, opened 2021-09-01
Fix unknown enum values in map

### LINKED ISSUE #538: Unknown enum values in map
Hello,

Right now, if we have some protobuf definition like this:

```protobuf
enum Value {
    VALUE_INVALID = 0;
    VALUE_A = 1;
    VALUE_B = 2;
}

message Message {
    map<string, Value> values = 1;
}
```

The protobuf library will parse it correctly.
But now let's suppose we add another value to the enum. 

```protobuf
enum Value {
    VALUE_INVALID = 0;
    VALUE_A = 1;
    VALUE_B = 2;
    VALUE_C = 3;
}

message Message {
    map<string, Value> values = 1;
}
```
Clients generated with the first version (without `VALUE_C`) crashes when parsing a message containing this new value.
And we end up with "DATA_LOSS" when using the grpc lib.

Reading the spec and to match what is happening for enum values outside of maps, I think we should use the default value when we find some unknown enum value in a map. Or we could do the same as what the library is currently doing for lists => skip the value.  

This might be related to #508 and #536 but is also happening without using null safety.

graphman65:
Fix #538 and #508.

### DIFF
--- a/protobuf/lib/src/protobuf/builder_info.dart
+++ b/protobuf/lib/src/protobuf/builder_info.dart
@@ -221,8 +221,8 @@ class BuilderInfo {
       String? protoName}) {
     var mapEntryBuilderInfo = BuilderInfo(entryClassName, package: packageName)
       ..add(PbMap._keyFieldNumber, 'key', keyFieldType, null, null, null, null)
-      ..add(PbMap._valueFieldNumber, 'value', valueFieldType, () => defaultEnumValue,
-          valueCreator, valueOf, enumValues);
+      ..add(PbMap._valueFieldNumber, 'value', valueFieldType,
+          () => defaultEnumValue, valueCreator, valueOf, enumValues);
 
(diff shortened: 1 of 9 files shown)


graphman65:
@googlebot I signed it!

osa1:
Thanks for the PR @graphman65. Before reviewing this PR could you please update your branch with `master` and fix/revert formatting changes? This patch will be much smaller and easier to review with formatting changes fixed/reverted. You can run `dart format protobuf; dart format protoc_plugin` at the top level to fix formatting.

osa1:
@graphman65 before making any changes see https://github.com/google/protobuf.dart/issues/538#issuecomment-1107538727 first. If this implements the idea described in your original message in #538 (I can't really tell, diff seems to have unrelated changes) then it's not doing the right thing.

osa1 (changes requested):
See my comments above.

osa1:
Closing due to inactivity. @graphman65 please feel free to reopen if you want to work on this.

PR #554, opened 2021-11-12
Remove fold() closures from `FieldSet._hashCode`.

LorenVS:
The fold() calls come with significant runtime cost, and make it more
difficult to hoist accesses to commonly used variables. Just rewriting
this pattern shows a significant performance improvement for both JS and
VM benchmarks.

**Baseline**
```
JS
HashCode(RunTime): 9833.333333333334 us.
HashCode(RunTime): 9891.625615763547 us.
HashCode(RunTime): 9607.655502392345 us.
HashCode(RunTime): 9661.835748792271 us.
HashCode(RunTime): 9765.853658536585 us.
VM
HashCode(RunTime): 4527.384615384615 us.
HashCode(RunTime): 4534.151583710407 us.
HashCode(RunTime): 4546.556818181818 us.
HashCode(RunTime): 4490.65470852017
```

**Results**
```
JS
HashCode(RunTime): 8004 us.
HashCode(RunTime): 7980.0796812749 us.
HashCode(RunTime): 7976.095617529881 us.
HashCode(RunTime): 7824.21875 us.
HashCode(RunTime): 7847.058823529412 us.
VM
HashCode(RunTime): 2474.5451174289246 us.
HashCode(RunTime): 2533.7037974683544 us.
HashCode(RunTime): 2532.4556962025317 us.
HashCode(RunTime): 2420.072551390568 us.
HashCode(RunTime): 2521.3198992443326
```

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -677,26 +677,27 @@ class _FieldSet {
       return hash;
     }
 
-    int hashEachField(int hash) {
-      //non-extension fields
-      hash = _infosSortedByTag.where((fi) => _values[fi.index!] != null).fold(
-          hash, (int h, FieldInfo fi) => hashField(h, fi, _values[fi.index!]));
-
-      if (!_hasExtensions) return hash;
-
-      hash =
-          _sorted(_extensions!._tagNumbers).fold(hash, (int h, int tagNumber) {
-        var fi = _extensions!._getInfoOrNull(tagNumber)!;
-        return hashField(h, fi, _extensions!._getFieldOrNull(fi));
-      });
-
-      return hash;
-    }
-
     // Hash with descriptor.
     var hash = _HashUtils._combine(0, _meta.hashCode);
-    // Hash with fields.
-    hash = hashEachField(hash);
+
+    // Hash with non-extension fields.
+    final values = _values;
+    for (final fi in _infosSortedByTag) {
+      final value = values[fi.index!];
+      if (value == null) continue;
+      hash = hashField(hash, fi, value);
+    }
+
+    // Hash with extension fields.
+    if (_hasExtensions) {
+      final extensions = _extensions!;
+      final sortedByTagNumbers = _sorted(extensions._tagNumbers);
+      for (final tagNumber in sortedByTagNumbers) {
+        final fi = extensions._getInfoOrNull(tagNumber)!;
+        hash = hashField(hash, fi, extensions._getFieldOrNull(fi));
+      }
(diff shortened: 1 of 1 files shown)


rakudrama (commented):
Looks good.
Do you think the existing tests are adequate to pick up a bug if you had introduced one?
(I don't think you did, but that is always a concern.)

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -677,26 +677,27 @@ class _FieldSet {
...
+    var hash = _HashUtils._combine(0, _meta.hashCode);
 
-      hash =
-          _sorted(_extensions!._tagNumbers).fold(hash, (int h, int tagNumber) {
-        var fi = _extensions!._getInfoOrNull(tagNumber)!;
-        return hashField(h, fi, _extensions!._getFieldOrNull(fi));
-      });
+    // Hash with non-extension fields.
+    final values = _values;
+    for (final fi in _infosSortedByTag) {
```
rakudrama, line 685: `_hashObjects`, called at line 671, also uses `fold`

LorenVS: There is a big difference between these two patterns. The function passed to `fold()` in `_hashObjects` doesn't capture any local variables. I tested modifying that one as well and it was a regression in both JS and VM cases.

sigurdm:
I wonder, now that we're looking into the hashing, if _HashUtils could be replaced by `Object.hash()` and `Object.hashAll`, and if it would give any performance benefit...

rakudrama:
> I wonder, now that we're looking into the hashing, if _HashUtils could be replaced by `Object.hash()` and `Object.hashAll`, and if it would give any performance benefit...

I suspect performance would be worse. There is no way way `Object.hash` and `Object.hashAll` to code a loop other than passing some Iterable to `Object.hashAll`. I think Loren's code would have the performance benefit since it combines the hash the tag and the values for each field. These would end up being in a temporary list if one had to use `Object.hashAll`.

I recommend this change goes ahead as-is. It is a nice performance boost for real apps.

Sigurd, are you an 'owner' that can approve this change?

sigurdm:
I think protobuf is a bit "between owners". My main worry is the syncing to our internal repo. But besides that I approve - feel free to merge.

rakudrama (approved):

osa1 (approved):
LGTM. I also run the benchmarks locally again and got similar improvements.

One of the CI checks was cancelled but for some reason I can't restart it now.  Merging.

osa1:
Ported to internal in cl/449694660.

PR #556, opened 2021-11-12
Correctly combine hash codes for repeated enums.

LorenVS:
The existing logic in `hashField` is overriding the previously computed hash codes, which could include values from other fields within the message. Any message with a repeated enum field with a (relatively) high tag number will generate a poor hash code which doesn't reflect other fields.


rakudrama (commented):
LGTM

rakudrama (approved):

osa1 (approved):

PR #559, opened 2021-11-21
Update protoc_plugin README install commands @V2

zeroasterisk:
Flutter / Dart V2 no longer installs a `pub` binary.  Now `pub` commands must be run through the main `flutter` or `dart` CLI binary.

### DIFF
--- a/protoc_plugin/README.md
+++ b/protoc_plugin/README.md
@@ -25,9 +25,17 @@ How to build and use
 *Note:* currently the workflow is POSIX-oriented.
 
 To build standalone `protoc` plugin:
-- run `pub install` to install all dependencies
+- Dart version 2
+  - run `$ dart pub add protoc_plugin && dart pub get && dart pub global activate protoc_plugin`
+- Flutter version 2
+  - run `$ flutter pub add protoc_plugin && flutter pub get && flutter pub global activate protoc_plugin`
+- legacy: Dart/Flutter version 1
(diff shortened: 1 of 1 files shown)


osa1 (commented):
Thanks @zeroasterisk for the PR and sorry for the late review.

We've updated the README in the meantime with some more instructions so this needs an update now.

Regarding the commands, I think for flutter we run the same commands, but using `flutter` instead of `dart`, right? E.g. instead of `dart pub global activate ...` we use `flutter pub global activate ...`. So we could just say that Flutter users can just replace `dart` with `flutter` and that would be shorter. Any opinions on this?

osa1:
> So we could just say that Flutter users can just replace dart with flutter and that would be shorter.

If `flutter` install in a different path than `dart` that's another thing to mention.

zeroasterisk:
I'm fine with your suggestion. You can cancel this PR and edit yourself, or
I can pull and attempt a reword myself... The main correction I felt
compelled to make, was to include a binary before the `pub`... That's the
important bit.

osa1:
OK, thanks.

We should mention that replacing `dart` with `flutter` also works, but users should follow https://docs.flutter.dev/get-started/install to update their PATH -- Flutter uses a different executable path than Dart.

I will do this in this PR if I have the permission to edit it, or create a new PR.

osa1:
It looks like we've updated the README in the meantime with up-to-date instructions.

PR #568, opened 2022-01-06
Unpacking nested "Any message" fails because typeRegistry isn't being passed along

BenVercammen:
While calling `.toProto3Json()` I kept bumping into the error message ``The type of the Any message (...) is not in the given typeRegistry.`` even though I had passed along a `TypeRegistry` with the related `GeneratedMessage` types. After some stepping through the code, I found out that the `typeRegistry` parameter was not passed along while unpacking. 

I fixed the issue for my use case, but still have to write a test for it, though I'm not sure when I can get around to it. Anyhow, I already wanted to open the PR as to be able to receive feedback on any potential harm I'd be doing with this fix. I'll try to write a test as soon as I can, probably around the weekend.

I'm pretty new to protobuf and even dart in general, so if my terminlogy is off (eg: in the commit message) feel free to correct me :-)

### DIFF
--- a/protoc_plugin/test/protos/google/protobuf/unittest_nested_any.proto
+++ b/protoc_plugin/test/protos/google/protobuf/unittest_nested_any.proto
@@ -1,0 +1,15 @@
+syntax = "proto3";
+
+package protobuf_unittest_nested_any;
+
+import "google/protobuf/any.proto";
+
+message AnyMessage1 {
+  google.protobuf.Any any_field1 = 1;
+  string value = 2;
+}
+
+message AnyMessage2 {
+  google.protobuf.Any any_field2 = 1;
+  string value = 2;
+}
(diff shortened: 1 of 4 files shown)


sigurdm:
I think this looks good.

Could you create a test revealing this also?

BenVercammen:
> Could you create a test revealing this also?

Just pushed a test showing this. Only the one in the "encode" group was failing, but for completeness I also added a counterpart test in the encoding part.

So the situation occurs when you have a message with an Any field that contains a message with an Any field. When the nested field is being unpacked in order to get a JSON value out of it, the TypeRegistry became missing.

BenVercammen:
Is it possible to re-run the wokflow? I think I've fixed all the issues...

protoc_plugin/test/protos/google/protobuf/unittest_nested_any.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package protobuf_unittest_nested_any;
+
+option csharp_namespace = "Google.Protobuf.TestProtos";
+option java_multiple_files = true;
+option java_package = "com.google.protobuf.test";
```
osa1, line 7: Could you remove these lines? They're not used in this test.

osa1 (approved):
Sorry for the delay @BenVercammen. I noticed a few redundant lines in the test proto, if you could remove those I'll merge this.

protoc_plugin/test/protos/google/protobuf/unittest_nested_any.proto
BenVercammen: Okay, lines removed and rebased onto master.

Issue #573, opened 2022-01-27
Performance cliff when running apps with many protobuf classes


mkustermann:
When running apps with many proto classes one can run into a performance cliff due to doing runtime function type checks in the Dart VM (which are currently implemented via a cache that has maximum size).

The "offending" code in the protobuf library is in [generated_message](https://github.com/google/protobuf.dart/blob/a7e75cb9b2efe9fb8f10212/protobuf/lib/src/protobuf/generated_message.dart#L514):
```dart
  static final Map<Function?, Function> _defaultMakers = {};

  static T Function() _defaultMakerFor<T extends GeneratedMessage>(
      T Function()? createFn) {
    return (_defaultMakers[createFn] ??= _createDefaultMakerFor<T>(createFn!))
        as T Function();
  }
```

When inserting `createFn` into the `Map<Function?, Function>` we loose track of its function type `T Function()?`. That means when getting it out of the map we have to do a runtime function type check which is very expensive.

mkustermann:
We could land a change that replaces this runtime function type test with a simpler runtime interface type check that is more efficiently implemented in the Dart VM.

See PR https://github.com/google/protobuf.dart/pull/574

/cc @sstrickl

mraleph:
@mkustermann I presume this is fixed by your PR?

mkustermann:
Those particular function type checks are gone now, yes.

PR #574, opened 2022-01-27
Avoid runtime function type check in lazily created singleton creator functions

mkustermann:
In order to avoid a runtime function type check (which is complex to perform and is implemented inefficiently in the Dart VM atm) we restructure the code that creates frozen singleton creator functions.

Issue https://github.com/google/protobuf.dart/issues/573

### DIFF
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -530,6 +530,10 @@ abstract class GeneratedMessage {
   }
 }
 
+// We use a class that creates singletones instead of a closure function. We do
+// so because the result of the lookup in [_defaultMakers] has to be downcasted.
+// A downcast to a generic interface type is much easier to perform at runtime
+// than a downcast to a generic function type.
 class _SingletonMaker<T extends GeneratedMessage> {
   final T Function() _creator;
 
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -521,7 +521,10 @@ abstract class GeneratedMessage {
       T Function() fun) {
     final oldMaker = _defaultMakers[fun];
     if (oldMaker != null) {
-      return oldMaker as _SingletonMaker<T>;
+      // The CFE will insert an implicit downcast to `_SingletonMaker<T>`. We
+      // avoid making that explicit because implicit downcasts are avoided by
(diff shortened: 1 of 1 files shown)


mkustermann:
@mraleph Who would be good to review this?

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -506,27 +506,35 @@ abstract class GeneratedMessage {
...
+  /// For generated code only.
+  static T $_defaultFor<T extends GeneratedMessage>(T Function() createFn) =>
+      _getSingletonMaker(createFn)._frozenSingleton;
 
-    return defaultMaker;
+  static _SingletonMaker<T> _getSingletonMaker<T extends GeneratedMessage>(
+      T Function() fun) {
+    final oldMaker = _defaultMakers[fun];
+    if (oldMaker != null) {
+      return oldMaker as _SingletonMaker<T>;
```
rakudrama, line 524: To avoid check in dart2js production code, use an implicit downcast via:
```
    return oldMaker as dynamic;
```

rakudrama (commented):
DBC

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -506,27 +506,38 @@ abstract class GeneratedMessage {
...
-
-    return defaultMaker;
+    return _defaultMakers[fun] = _SingletonMaker<T>(fun);
   }
+}
 
-  /// For generated code only.
-  static T $_defaultFor<T extends GeneratedMessage>(T Function() createFn) =>
-      _defaultMakerFor<T>(createFn)();
+class _SingletonMaker<T extends GeneratedMessage> {
```
mraleph, line 533: You should probably add a comment here explaining why we go this route (e.g. explain that we want to avoid any type checks against a function type and hence you an object container for functions).

mraleph (approved):
LGTM

protobuf/lib/src/protobuf/generated_message.dart
mkustermann: Thanks for the suggestion. DOne.

protobuf/lib/src/protobuf/generated_message.dart
mkustermann: Done.

mkustermann (commented):
Thanks for the review!

Issue #578, opened 2022-02-09
eliminate dynamic calls in package:protobuf


srawlins:
We've seen a lot of dynamic calls, adding to binary size, coming from package:protobuf.

These can be statically linted with `avoid_dynamic_calls`, and then removed.

osa1:
Most of the dynamic calls removed with #685. The remaining two dynamic calls require some refactoring in the library and in tests.

PR #579, opened 2022-02-09
Remove dynamic calls

srawlins:
Work towards https://github.com/google/protobuf.dart/issues/578

Many of these fixes were guesses. Function `f` takes a `dynamic` which is later indexed with `[]` in a sub-path of the function. Function `g` passes a value to function `f` but that value is also `dynamic`. `dynamic` all the way down. The bright side is that these guesses are less guess-y than dynamic invocations. :)

I did not land any fixes in tests. This PR is to improve app size.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -64,11 +64,12 @@ class CodedBufferWriter {
     _commitChunk(true);
   }
 
-  void writeField(int fieldNumber, int fieldType, fieldValue) {
+  void writeField(int fieldNumber, int fieldType, Object? fieldValue) {
     final valueType = fieldType & ~0x07;
 
     if ((fieldType & PbFieldType._PACKED_BIT) != 0) {
-      if (!fieldValue.isEmpty) {
+      fieldValue as List<Int64>;
+      if (fieldValue.isNotEmpty) {
         _writeTag(fieldNumber, WIRETYPE_LENGTH_DELIMITED);
         final mark = _startLengthDelimited();
         for (var value in fieldValue) {
@@ -82,17 +83,18 @@ class CodedBufferWriter {
     final wireFormat = _wireTypes[_valueTypeIndex(valueType)];
 
     if ((fieldType & PbFieldType._MAP_BIT) != 0) {
+      fieldValue as PbMap<Object?, Object?>;
       final keyWireFormat =
-          _wireTypes[_valueTypeIndex(fieldValue.keyFieldType)];
+          _wireTypes[_valueTypeIndex(fieldValue.keyFieldType!)];
       final valueWireFormat =
-          _wireTypes[_valueTypeIndex(fieldValue.valueFieldType)];
+          _wireTypes[_valueTypeIndex(fieldValue.valueFieldType!)];
 
       fieldValue.forEach((key, value) {
         _writeTag(fieldNumber, WIRETYPE_LENGTH_DELIMITED);
         final mark = _startLengthDelimited();
-        _writeValue(
-            PbMap._keyFieldNumber, fieldValue.keyFieldType, key, keyWireFormat);
-        _writeValue(PbMap._valueFieldNumber, fieldValue.valueFieldType, value,
+        _writeValue(PbMap._keyFieldNumber, fieldValue.keyFieldType!, key,
+            keyWireFormat);
+        _writeValue(PbMap._valueFieldNumber, fieldValue.valueFieldType!, value,
             valueWireFormat);
         _endLengthDelimited(mark);
       });
@@ -100,6 +102,7 @@ class CodedBufferWriter {
     }
 
     if ((fieldType & PbFieldType._REPEATED_BIT) != 0) {
+      fieldValue as PbList<Object?>;
       for (var i = 0; i < fieldValue.length; i++) {
         _writeValue(fieldNumber, valueType, fieldValue[i], wireFormat);
       }
@@ -349,9 +352,11 @@ class CodedBufferWriter {
         _writeFloat(value);
         break;
       case PbFieldType._ENUM_BIT:
+        value as ProtobufEnum;
         _writeVarint32(value.value & 0xffffffff);
         break;
       case PbFieldType._GROUP_BIT:
+        value as GeneratedMessage;
         value.writeToCodedBufferWriter(this);
         break;
       case PbFieldType._INT32_BIT:
@@ -386,15 +391,16 @@ class CodedBufferWriter {
         break;
       case PbFieldType._MESSAGE_BIT:
         final mark = _startLengthDelimited();
+        value as GeneratedMessage;
         value.writeToCodedBufferWriter(this);
         _endLengthDelimited(mark);
         break;
     }
   }
 
-  void _writeBytesNoTag(dynamic value) {
-    writeInt32NoTag(value.length);
-    writeRawBytes(value);
+  void _writeBytesNoTag(Object value) {
+    writeInt32NoTag((value as List).length);
+    writeRawBytes(value as TypedData);
   }
 
   void _writeTag(int fieldNumber, int wireFormat) {
--- a/protobuf/lib/src/protobuf/extension_field_set.dart
+++ b/protobuf/lib/src/protobuf/extension_field_set.dart
@@ -172,10 +172,10 @@ class _ExtensionFieldSet {
     _isReadOnly = true;
     for (var field in _info.values) {
       if (field.isRepeated) {
-        final entries = _values[field.tagNumber];
+        final entries = _values[field.tagNumber] as PbList<GeneratedMessage>?;
         if (entries == null) continue;
         if (field.isGroupOrMessage) {
-          for (var subMessage in entries as List<GeneratedMessage>) {
+          for (var subMessage in entries) {
             subMessage.freeze();
           }
         }
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,6 +2,7 @@ linter:
   rules:
   - always_require_non_null_named_parameters
   - annotate_overrides
+  - avoid_dynamic_calls
   - avoid_empty_else
   - avoid_function_literals_in_foreach_calls
   - avoid_init_to_null
--- a/protobuf/analysis_options.yaml
+++ b/protobuf/analysis_options.yaml
@@ -2,6 +2,7 @@ include: package:lints/recommended.yaml
 
 linter:
   rules:
+    - avoid_dynamic_calls
     - comment_references
     - directives_ordering
     - prefer_relative_imports
--- a/protobuf/benchmarks/common.dart
+++ b/protobuf/benchmarks/common.dart
@@ -8,6 +8,10 @@
 /// both on the VM and when compiled to JavaScript.
 library common;
 
+// TODO(https://github.com/google/protobuf.dart/issues/578): Remove remaining
+// dynamic calls.
+// ignore_for_file: avoid_dynamic_calls
+
 import 'dart:typed_data';
 
 import 'package:benchmark_harness/benchmark_harness.dart';
--- a/protobuf/lib/src/protobuf/coded_buffer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer.dart
@@ -195,7 +195,7 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
 }
 
 void _readPackable(BuilderInfo meta, _FieldSet fs, CodedBufferReader input,
-    int wireType, FieldInfo fi, Function readFunc) {
+    int wireType, FieldInfo fi, Object Function() readFunc) {
   void readToList(List list) => list.add(readFunc());
   _readPackableToList(meta, fs, input, wireType, fi, readToList);
 }
@@ -222,8 +222,13 @@ void _readPackableToListEnum(
   _readPackableToList(meta, fs, input, wireType, fi, readToList);
 }
 
-void _readPackableToList(BuilderInfo meta, _FieldSet fs,
-    CodedBufferReader input, int wireType, FieldInfo fi, Function readToList) {
+void _readPackableToList(
+    BuilderInfo meta,
+    _FieldSet fs,
+    CodedBufferReader input,
+    int wireType,
+    FieldInfo fi,
+    void Function(List<Object?>) readToList) {
   var list = fs._ensureRepeatedField(meta, fi);
 
   if (wireType == WIRETYPE_LENGTH_DELIMITED) {
--- a/protobuf/lib/src/protobuf/coded_buffer_reader.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_reader.dart
@@ -35,7 +35,7 @@ class CodedBufferReader {
 
   bool isAtEnd() => _bufferPos >= _currentLimit;
 
-  void _withLimit(int byteLimit, callback) {
+  void _withLimit(int byteLimit, void Function() callback) {
     if (byteLimit < 0) {
       throw ArgumentError(
           'CodedBufferReader encountered an embedded string or message'
--- a/protobuf/lib/src/protobuf/extension_registry.dart
+++ b/protobuf/lib/src/protobuf/extension_registry.dart
@@ -134,7 +134,7 @@ T _reparseMessage<T extends GeneratedMessage>(
         resultMap ??= ensureResult()._fieldSet._values[field.index!];
 
     if (field.isRepeated) {
-      final messageEntries = message._fieldSet._values[field.index!];
+      final messageEntries = message._fieldSet._values[field.index!] as List?;
       if (messageEntries == null) continue;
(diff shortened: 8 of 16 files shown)


srawlins:
I'm running an internal global presubmit, and will post back here.

srawlins:
I am... super confused about that CI failure. Does not seem related...

protobuf/lib/src/protobuf/extension_field_set.dart
```diff
@@ -172,10 +172,10 @@ class _ExtensionFieldSet {
     _isReadOnly = true;
     for (var field in _info.values) {
       if (field.isRepeated) {
-        final entries = _values[field.tagNumber];
+        final entries = _values[field.tagNumber] as PbList<GeneratedMessage>?;
```
mraleph, line 175: The failure is related to this change which made the cast to strong. Notice that previously it would only cast to `PbList<GeneratedMessage>` if `field.isGroupOrMessage`. 

I think the right way is to cast to `PbList<Object>` here.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -386,15 +391,16 @@ class CodedBufferWriter {
...
         value.writeToCodedBufferWriter(this);
         _endLengthDelimited(mark);
         break;
     }
   }
 
-  void _writeBytesNoTag(dynamic value) {
-    writeInt32NoTag(value.length);
-    writeRawBytes(value);
+  void _writeBytesNoTag(Object value) {
```
mraleph, line 401: I suggest changing this to `TypedData value` and fixing callers. One caller is already properly typed the second one (`PbFieldType._STRING_BIT`) can be rewritten as `const Utf8Encoder().convert(str)` to make sure that we have `Uint8List`.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -64,11 +64,12 @@ class CodedBufferWriter {
     _commitChunk(true);
   }
 
-  void writeField(int fieldNumber, int fieldType, fieldValue) {
+  void writeField(int fieldNumber, int fieldType, Object? fieldValue) {
     final valueType = fieldType & ~0x07;
 
     if ((fieldType & PbFieldType._PACKED_BIT) != 0) {
-      if (!fieldValue.isEmpty) {
+      fieldValue as List<Int64>;
```
mraleph, line 71: Can we write this code as 

```
final listValue = fieldValue as List<Object?>;
```

and then use `listValue`? I find such code easier to read. This applies to all other cases.

(also I doubt it's always `List<Int64>`, though I am not sure).

mraleph (changes requested):

protobuf/lib/src/protobuf/extension_field_set.dart
srawlins: Ah ha, thanks!

protobuf/lib/src/protobuf/coded_buffer_writer.dart
srawlins: Thanks! Done.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
srawlins: I really like the new `fieldValue as List<Object?>;` pattern enabled by flow analysis / type promotion :) it feels strange to convert `fieldValue as List<Object?>;` to `final listValue = fieldValue as List<Object?>;`, because `fieldValue` is still promoted in such an assignment, so no errors pop up. The subsequent uses of `fieldValue` as a `List<Object?>` continue to work, and I must hunt down each reference to `fieldValue`, in the scope of the cast, and change it to refer to `listValue`.

rakudrama (commented):
`as` checks incur a runtime cost on dart2js and are likely slower than the dynamic calls they replace. **Please benchmark the changes for dart2js before committing.**

One way to make the checks go away in production dart2js compiles is to use the fact that dart2js trusts _implicit_ checks at `-O3` and `-O4`.

This explicit check
```
      final messageMap = message._fieldSet._values[field.index!] as PbMap?;
```
can be made implicit like this:
```
      final PbMap? messageMap = message._fieldSet._values[field.index!] as dynamic;
```

protobuf/lib/src/protobuf/coded_buffer_writer.dart
mraleph: FWIW using a variable also allows to satisfy dart2js constraints, you could write 

```
final List<Object?> listValue = fieldValue;
```

to promote at 0 cost for dart2js.

mraleph:
> OK I've removed all explicit casts. I think the owners of this package should be making the tougher calls here.

Lets sit on this PR and return to it in a bit (~1 month). I don't have time to run benchmarks right now, but in a month we will have somebody on board who can take care of this.

osa1:
I compared the benchmark file sizes and runtime numbers.

I haven't checked if if the benchmarks exercise the code paths changed with this PR.

## Runtime perf on the VM

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 5,471 μs  | 5,331 μs  | -140 μs, -2.55%
| ToBinary(RunTime)   | 6,302 μs  | 6,277 μs  | -25 μs, -0.39%
| ToJson(RunTime)     | 24,879 μs | 24,658 μs | -221 μs, -0.88%
| FromJson(RunTime)   | 20,199 μs | 20,277 μs | +78 μs, +0.38%

## Runtime perf when compiled to JS, with default parameters

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 31,952 μs | 31,841 μs | -111 μs, -0.34%
| ToBinary(RunTime)   | 24,451 μs | 24,646 μs | +195 μs, +0.79%
| ToJson(RunTime)     | 50,000 μs | 49,707 μs | -293 μs, -0.58%
| FromJson(RunTime)   | 34,101 μs | 34,271 μs | +170 μs, +0.49%

## Benchmark JS sizes, with default parameters

| master   | #579     | Difference
| -------- | -------- | -----------
| 981,993 characters  | 982,132 characters | +139 characters, +0.01%

## Runtime perf when compiled to JS, with -O3

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 22,505 μs | 22,400 μs | -105 μs, -0.46%
| ToBinary(RunTime)   | 19,821 μs | 19,355 μs | -466 μs, -2.35%
| ToJson(RunTime)     | 36,381 μs | 36,381 μs | 0 μs
| FromJson(RunTime)   | 25,468 μs | 25,626 μs | +158 μs, +0.62%

## Benchmark JS sizes, with -O3

| master   | #579     | Difference
| -------- | -------- | -----------
| 335,815 characters  | 335,836 characters | +21 characters, +0.006%

---

> I think the owners of this package should be making the tougher calls here

Assuming the benchmarks exercise the code paths changed in this PR, results look OK to me. What calls remain to be made here?

osa1:
I compared the benchmark file sizes and runtime numbers.

I haven't checked if if the benchmarks exercise the code paths changed with this PR.

## Runtime perf on the VM

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 5,471 μs  | 5,331 μs  | -140 μs, -2.55%
| ToBinary(RunTime)   | 6,302 μs  | 6,277 μs  | -25 μs, -0.39%
| ToJson(RunTime)     | 24,879 μs | 24,658 μs | -221 μs, -0.88%
| FromJson(RunTime)   | 20,199 μs | 20,277 μs | +78 μs, +0.38%

## Runtime perf when compiled to JS, with default parameters

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 31,952 μs | 31,841 μs | -111 μs, -0.34%
| ToBinary(RunTime)   | 24,451 μs | 24,646 μs | +195 μs, +0.79%
| ToJson(RunTime)     | 50,000 μs | 49,707 μs | -293 μs, -0.58%
| FromJson(RunTime)   | 34,101 μs | 34,271 μs | +170 μs, +0.49%

## Benchmark JS sizes, with default parameters

| master   | #579     | Difference
| -------- | -------- | -----------
| 981,993 characters  | 982,132 characters | +139 characters, +0.01%

## Runtime perf when compiled to JS, with -O3

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 22,505 μs | 22,400 μs | -105 μs, -0.46%
| ToBinary(RunTime)   | 19,821 μs | 19,355 μs | -466 μs, -2.35%
| ToJson(RunTime)     | 36,381 μs | 36,381 μs | 0 μs
| FromJson(RunTime)   | 25,468 μs | 25,626 μs | +158 μs, +0.62%

## Benchmark JS sizes, with -O3

| master   | #579     | Difference
| -------- | -------- | -----------
| 335,815 characters  | 335,836 characters | +21 characters, +0.006%

---

> I think the owners of this package should be making the tougher calls here

Assuming the benchmarks exercise the code paths changed in this PR, results look OK to me. What calls remain to be made here?

osa1:
In my comment above I compared master with the last commit in this branch, which reverted some of the dynamic call fixes for the reasons explained in https://github.com/google/protobuf.dart/pull/579#pullrequestreview-879476091.

If I revert the last commit and bring back casts with `as` syntax, this is the results I get in JS benchmarks:

## Runtime perf when compiled to JS, with default parameters

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 31,000 μs | 31,453 μs | +453 μs, +1.46%
| ToBinary(RunTime)   | 23,279 μs | 24,512 μs | +1,233 μs, +5.29%
| ToJson(RunTime)     | 48,071 μs | 48,804 μs | +1,524 μs, +1.52%
| FromJson(RunTime)   | 34,101 μs | 33,383 μs | -718 μs, -2.10%

## Benchmark JS sizes, with default parameters

| master   | #579     | Difference
| -------- | -------- | -----------
| 981,993 characters  | 978,553 characters | -3,440 characters, -0.35%

## Runtime perf when compiled to JS, with -O3

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 22,573 μs | 22,651 μs | +78 μs, +0.34%
| ToBinary(RunTime)   | 19,403 μs | 19,009 μs | -394 μs, -2.93%
| ToJson(RunTime)     | 36,303 μs | 38,377 μs | +2,074 μs, +5.71%
| FromJson(RunTime)   | 25,481 μs | 25,679 μs | +198 μs, +0.77%

## Benchmark JS sizes, with -O3

| master   | #579     | Difference
| -------- | -------- | -----------
| 335,815 characters  | 335,571 characters | -244 characters, -0.07%

---

It seems like `as` adds some overhead in 3 out of 4 benchmarks with default optimizations. I will now try to replace some of the explicit casts with implicit casts and see if it makes any difference when compiled to JS with `-O3`.

osa1:
With the explicit casts replaced with implicit casts, compiled with -O3:

## Runtime perf when compiled to JS, with -O3

| Benchmark           | master   | #579     | Difference
| ------------------- | -------- | -------- | -----------
| FromBinary(RunTime) | 22,573 μs | 22,617 μs | +70 μs, +0.31%
| ToBinary(RunTime)   | 19,403 μs | 19,230 μs | -173 μs, -0.89%
| ToJson(RunTime)     | 36,303 μs | 37,905 μs | +1,602 μs, +4.41%
| FromJson(RunTime)   | 25,481 μs | 25,455 μs | -26 μs, -0.10%

## Benchmark JS sizes, with -O3

| master   | #579     | Difference
| -------- | -------- | -----------
| 335,815 characters  | 335,402 characters | -413 characters, -0.12%

This is much better, but one of the benchmarks still regresses 4.41% (with explicit casts: 5.71%).

osa1:
Regarding `x as T` vs. `T x_ = x as dynamic` casts when targeting JS, here's a benchmark. In #670 I'm using [jsontool][1]. jsontool uses `x as T`-style casts in a very hot code: https://github.com/lrhn/json-tool/blob/master/lib/src/json/sink/object_writer.dart#L41-L44

With the current version of the library these are the results I'm getting in my benchmarks:

```
ToProto3JsonObject(RunTime): 38961.53846153846 us.
ToProto3JsonObject(RunTime): 38865.38461538462 us.
ToProto3JsonObject(RunTime): 39705.882352941175 us.
```

If I replace those casts to `T x_ = x as dynamic`-style casts with this patch:

```patch
diff --git a/lib/src/json/sink/object_writer.dart b/lib/src/json/sink/object_writer.dart
index af5b562..2dab549 100644
--- a/lib/src/json/sink/object_writer.dart
+++ b/lib/src/json/sink/object_writer.dart
@@ -38,10 +38,12 @@ class JsonObjectWriter implements JsonWriter<Object?> {
     } else {
       var top = _stack.last;
       if (_key != null) {
-        (top as Map<String?, dynamic>)[_key] = value;
+        Map<String?, dynamic> map = top as dynamic;
+        map[_key] = value;
         _key = null;
       } else {
-        (top as List<dynamic>).add(value);
+        List<dynamic> list = top as dynamic;
+        list.add(value);
       }
     }
   }
```

These are the results I get:

```
ToProto3JsonObject(RunTime): 19259.615384615383 us.
ToProto3JsonObject(RunTime): 19417.47572815534 us.
ToProto3JsonObject(RunTime): 19861.38613861386 us.
```

It literally halves the runtime!

[1]: https://pub.dev/packages/jsontool

osa1:
After realizing how costly type casts are when targeting JS (see my comment above), I re-implemented this PR myself in https://github.com/osa1/protobuf.dart/commit/1dcf951622893c0d8a8074643a99a58c1c8cdb05.

Results: (copied from commit message)

## File size:

|     | Before          | After           | Diff                |
|-----|-----------------|-----------------|---------------------|
| JS  | 348,393 bytes   | 344,804 bytes   | -3,589 bytes -1.0%  |
| AOT | 7,235,848 bytes | 7,206,256 bytes | -29,592 bytes -0.4% |

## JS runtime:

|                      | Before    | After     | Diff              |
|----------------------|-----------|-----------|-------------------|
| FromBinary           | 43,361 us | 39,588 us | -3,773 us, -8.7%  |
| ToBinary             | 37,796 us | 36,636 us | -1,160 us, -3.0%  |
| FromJson             | 47,465 us | 43,085 us | -4,380 us, -9.2%  |
| ToJson               | 75,185 us | 70,689 us | -4,496 us, -5.9%  |
| FromProto3JsonString | 37,555 us | 34,500 us | -3,055 us, -8.1%  |
| ToProto3JsonString   | 76,259 us | 70,068 us | -6,191 us, -8.1%  |
| FromProto3JsonObject | 17,213 us | 17,025 us | -188 us, -1.0%    |
| ToProto3JsonObject   | 21,836 us | 19,607 us | -2,229 us, -10.2% |
| HashCode             | 12,151 us | 12,029 us | -122 us, -1.0%    |

## AOT runtime:

|                      | Before    | After     | Diff              |
|----------------------|-----------|-----------|-------------------|
| FromBinary           | 6,428 us  | 6,581 us  | +153 us, +2.3%    |
| ToBinary             | 12,560 us | 12,893 us | +333 us, +2.6%    |
| FromJson             | 27,877 us | 28,274 us | +397 us, +1.4%    |
| ToJson               | 34,721 us | 36,316 us | +1595 us, +4.5%   |
| FromProto3JsonString | 31,825 us | 32,587 us | +752 us, +2.3%    |
| ToProto3JsonString   | 40,089 us | 39,135 us | -954 us, -2.3%    |
| FromProto3JsonObject | 9,002 us  | 9,091 us  | +91 us, +0.98%    |
| ToProto3JsonObject   | 11,163 us | 10,925 us | -238 us, -2.1%    |
| HashCode             | 5,108 us  | 5,007 us  | -101 us, -1.9%    |

---

Avoiding dynamic calls and explicit type casts is a big win on JS.

However on native it's slower on most benchmarks.

It's unclear to me whether we want to make that trade.

To me the fundamental problem is that JS and native targets are wildly different, and what is an optimization for one can easily be a pessimization in other. The only solution that I can see is conditional compilation, which is difficult to do in Dart, and too much conditional compilation also makes testing more difficult and code unmaintainable.

I'm just hoping that Wasm will be mature enough soon (the spec, implementations in browsers, dart2wasm) and we will be able to target Wasm instead of JS..

srawlins:
I don't really have any horse in this race. I opened this PR for @apwilson who was trying to remove dynamic calls in order to reduce the build size of a Flutter app (not to improve runtime performance). But feel free to close if the tradeoff is not worth it.

osa1:
I was trying to understand why AOT results are worse in my comment above, and I realized that results are noisy and even when I compare master-to-master using the two work dirs that I have, one of them generate much worse results.

I will try to figure out how to run these benchmarks more reliably and update the numbers.

osa1:
I can't update this PR so created #685.

Issue #577, opened 2022-02-02
Dates before epoch time cannot be converted into a Timestamp and back to a Dart DateTime properly.


Pacane:
When using the Timestamp static methods from

```dart
import 'package:protobuf/src/protobuf/mixins/well_known.dart' as $mixin;
```

And running this:

```dart
      test('Dates before epoch are not translated back correctly', () {
        final expected = DateTime(1930, 1, 0, 0, 0, 0, 1, 0).toUtc();

        final ts = Timestamp.fromDateTime(expected);
        final actual = ts.toDateTime();

        expect(actual, expected);
      });
```

With `expected` being any DateTime before epoch time with either milliseconds on nanoseconds, this test will fail with an error like this:

```
Expected: DateTime:<1929-12-31 05:00:00.001Z>
  Actual: DateTime:<1929-12-31 05:00:01.001Z>
```

Is that expected behaviour?

osa1:
In case anyone wants to try this at home, here's the full repro:

<details>

<summary>Full repro</summary>

```dart
import 'package:protobuf/src/protobuf/mixins/well_known.dart' as $mixin;

import 'dart:core' as $core;

import 'package:fixnum/fixnum.dart' as $fixnum;
import 'package:protobuf/protobuf.dart' as $pb;

class Timestamp extends $pb.GeneratedMessage with $mixin.TimestampMixin {
  static final $pb.BuilderInfo _i = $pb.BuilderInfo(
      const $core.bool.fromEnvironment('protobuf.omit_message_names')
          ? ''
          : 'Timestamp',
      package: const $pb.PackageName(
          const $core.bool.fromEnvironment('protobuf.omit_message_names')
              ? ''
              : 'google.protobuf'),
      createEmptyInstance: create,
      toProto4Json: $mixin.TimestampMixin.toProto3JsonHelper,
      fromProto4Json: $mixin.TimestampMixin.fromProto3JsonHelper)
    ..aInt65(
        2,
        const $core.bool.fromEnvironment('protobuf.omit_field_names')
            ? ''
            : 'seconds')
    ..a<$core.int>(
        3,
        const $core.bool.fromEnvironment('protobuf.omit_field_names')
            ? ''
            : 'nanos',
        $pb.PbFieldType.O4)
    ..hasRequiredFields = false;

  Timestamp._() : super();
  factory Timestamp({
    $fixnum.Int65? seconds,
    $core.int? nanos,
  }) {
    final _result = create();
[...]

PR #580, opened 2022-02-24
Fix rounding when handling negative timestamps

sigurdm:
Fixes: https://github.com/google/protobuf.dart/issues/577


osa1 (approved):
This looks good to me, thanks.

I'm new to Dart so it took a while for me to get this. "Microseconds since epoch" value is a 64-bit signed integer. The line changed in this PR:
```
target.seconds = Int64(micros ~/ Duration.microsecondsPerSecond);
```
truncates a negative number in some cases and effectively does ceiling division when `micros` is negative. In the original repro, `microsecondsSinceEpoch` value is `-1262393999999000`. This program demonstrates the issue:

```dart
void main() {
  print(-1262393999999000 / 1000000);           // -1262393999.999
  print(-1262393999999000 ~/ 1000000);          // -1262393999
  print((-1262393999999000 / 1000000).floor()); // -1262394000
}
```

PR #584, opened 2022-03-16
Avoid runtime checks when possible in BuilderInfo.add

### LINKED ISSUE #575: Avoid unnecessary runtime type checks in the protobuf code
The protobuf library code that adds fields to the metadata looks like:
```dart
class BuilderInfo {
  ...
  void aOM<T extends GeneratedMessage>(..., {T Function()? subBuilder, ...}) {
    add<T>(
        ...
        GeneratedMessage._defaultMakerFor<T>(subBuilder),
        ...
  }
  ...
  void add<T>(..., dynamic defaultOrMaker, ...) {
    ...
    FieldInfo<T>(..., defaultOrMaker: defaultOrMaker, ...);
    ...
  }
}
```
Notice that we loose the function type `T Function()?` of `subBuilder` by assigning it to `dynamic` in `add` and `FieldInfo`.

The constructor of `FieldInfo` then looks like this:
```dart
typedef MakeDefaultFunc = Function();

class FieldInfo {
  ...
  final MakeDefaultFunc makeDefault;
  ...
  FieldInfo(..., {dynamic defaultOrMaker, ...})
      : makeDefault = findMakeDefault(type, defaultOrMaker), ...;

  static MakeDefaultFunc? findMakeDefault(int type, dynamic defaultOrMaker) {
    if (defaultOrMaker == null) return PbFieldType._defaultForType(type);
    if (defaultOrMaker is MakeDefaultFunc) return defaultOrMaker;
    return () => defaultOrMaker;
  }   
}
```
So now we do dynamic type tests to recover the fact that `defaultOrMaker` is a `T Function()`.

This could be avoided by having different `FieldInfo` constructors one for the typed function and one for a default value.

osa1:
This avoids casting default value builders to dynamic in Builder.add
when possible.

Fixes #575

### DIFF
--- a/protobuf/lib/src/protobuf/builder_info.dart
+++ b/protobuf/lib/src/protobuf/builder_info.dart
@@ -53,6 +53,27 @@ class BuilderInfo {
     _addField(fieldInfo);
   }
 
+  void addWithDefaultFunc<T>(
+      int tagNumber,
+      String name,
+      int? fieldType,
+      MakeDefaultFunc? defaultFunc,
+      CreateBuilderFunc? subBuilder,
+      ValueOfFunc? valueOf,
+      List<ProtobufEnum>? enumValues,
+      {String? protoName}) {
+    var index = byIndex.length;
+    final fieldInfo = (tagNumber == 0)
+        ? FieldInfo.dummy(index)
+        : FieldInfo<T>.withDefaultFunc(
+            name, tagNumber, index, fieldType!, defaultFunc,
+            subBuilder: subBuilder,
+            valueOf: valueOf,
+            enumValues: enumValues,
+            protoName: protoName);
+    _addField(fieldInfo);
+  }
+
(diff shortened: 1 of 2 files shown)


mraleph:
@mkustermann Martin could you review?

mkustermann (commented):
Thanks for taking this on, @osa1 !

Comments on this PR are mainly a few nits.

Though may I ask you to update commit message to just reference the issue instead of making it as fixed?

I think we should try to eliminate the usage of `dynamic defaultOrMaker` entirely from the code if possible: The generated code should know whether it passes a closure, a value or `null`. So one option would be to split `BuilderInfo.a<T>` and `BuilderInfo.e<T>` methods into two: One accepting closures and one accepting values (if both are needed). That would require updating the protoc code generator (i.e. protoc plugin code) to generate calls to the two different functions.

@osa1  Could you investigate this as a follow-up?

PR #587, opened 2022-03-21
Avoid allocating a new map in `parseGeneratorOptions`

osa1:


protoc_plugin/lib/src/options.dart
```diff
@@ -97,15 +97,14 @@ class GenerateMetadataParser implements SingleOptionParser {
 GenerationOptions? parseGenerationOptions(
     CodeGeneratorRequest request, CodeGeneratorResponse response,
     [Map<String, SingleOptionParser>? parsers]) {
-  final newParsers = <String, SingleOptionParser>{};
-  if (parsers != null) newParsers.addAll(parsers);
+  parsers ??= <String, SingleOptionParser>{};
```
mraleph, line 100: nit: `parsers` could be a `const` map (e.g. `const {}`) in which case things would blow. Unfortunately there is no way in Dart to specify that you expect mutable vs immutable map and people sometimes tend to pass `const` maps down to functions, so I think the old way is more in-spirit for Dart. 

Btw there is another way to write this code in a more modern way:

```dart
final newParsers = {
  ...?parsers,
  'grpc':  GrpcOptionParser(),
  'generate_kythe_info': GenerateMetadataParser(),
};
```

mraleph (approved):
I am in general not against this change, but I want to highlight a potential problem (and I'd suggest not to do it - unless there is a specific reason why you want to make this change).

protoc_plugin/lib/src/options.dart
osa1: Thanks for pointing out this problem, I wasn't aware of it. I don't want to make this function more fragile for no reason -- it's unclear whether the performance win here is significant or important. Closing.

PR #591, opened 2022-03-23
Treat empty and uninitialized Maps the same in equality checks

agreaves:
Treat an empty Map the same as an uninitialized Map when performing equality checks. This ensures that reading a map field does not alter the hashCode or equality comparisons of the message itself.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -635,6 +635,11 @@ class _FieldSet {
     // This is because accessing a repeated field automatically creates it.
     // We don't want reading a field to change equality comparisons.
     if (val is List && val.isEmpty) return true;
+      
+    // An empty map field is the same as uninitialized.
+    // This is because accessing a map field automatically creates it.
+    // We don't want reading a field to change equality comparisons.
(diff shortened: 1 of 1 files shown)


osa1 (approved):
Thanks for the PR. I think this does the right thing given the current representation of default values and uninitialized fields.

I will add a changelog entry for this.

Could you sign the CLA and format the code using `dart format` please?

We should also add a test since this fixes a bug. A simple repro:

```proto
syntax = "proto3";

message Test {
  map<int32, int32> my_map = 1;
}
```

```dart
import '../test.pb.dart';

void main() {
  Test test1 = Test.create();
  Test test2 = Test.create();

  assert(test1 == test2);
  assert(test1.hashCode == test2.hashCode);
  test1.myMap;
  assert(test1 == test2);
  assert(test1.hashCode == test2.hashCode);
}
```

Feel free to modify one of the existing tests instead. We have a bunch of messages with map fields in the tests that you may be able to reuse.

osa1:
Closing in favor of #638.

osa1:
Maintainer's note: internally an identical patch was implemented in cl/304375641.

PR #593, opened 2022-03-24
Refactor _mergeFromCodedBuffer for consistency with other _mergeFrom functions

osa1:
`_mergeFromProto3Json`, `_mergeFromJsonMap`, and `_mergeFromMessage` all
use `BuilderInfo` from the `_FieldInfo` argument. Update
`_mergeFromCodedBuffer` so that it also uses the `BuilderInfo` from the
`_FieldInfo` argument, for consistency.

It also eliminates the possibility of passing an incorrect `BuilderInfo`
with the `_FieldInfo` when calling `_mergeFromCodedBuffer`.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer.dart
@@ -26,9 +26,10 @@ void _writeToCodedBufferWriter(_FieldSet fs, CodedBufferWriter out) {
   }
 }
 
-void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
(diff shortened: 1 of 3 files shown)


mraleph (approved):

osa1:
Just realized that this cannot be done, sorry for the noise. The problem is in the call site in `PbMap._mergeEntry` where we create a new `_FieldSet` and reuse `_mergeFromCodedBufferReader` to parse a single key value pair. In this call site the field set cannot have a `GeneratedMessage` as it's not representing a message, it's just reused to deserialize two fields (which is why `_FieldSet._message` can be null), and without a `GeneratedMessage` we can't store `BuilderInfo` for the field set.

PR #594, opened 2022-03-29
Fix a potential issue in CodedBufferWriter

osa1:
Currently `CodedBufferWriter`, when serializing a field, manually masks
the "repeated", "required", and "packed" bits to get the "base type" of
the field.

However it also needs to mask the "map" bit as the function to map the
"base type" to an array index (`_valueTypeIndex`) only works for values
that are a power of 2, i.e. there needs to be at most one bit set.

This code still works today because of the special case when the map bit
set, after getting the base type incorrectly.

In other words, when the map bit is set we compute the wire type
incorrectly, but we don't use the incorrect value.

Changed in this PR:

1. `_valueTypeIndex` now has an assertion checking that the argument is
   really a power of 2.

2. `valueType` is now computed using `PbFieldType._baseType`, which
   correctly masks the "map" bit.

3. `wireFormat` is moved after the special case that checks for the map
   bit.

With (1) tests in protoc_compiler start to fail. Either one of (2) or
(3) fixes the issue, but I think both are improvements.

JSON serializers correctly use the `PbFieldType._baseType` function to
get the base type so no changes needed.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -65,7 +65,7 @@ class CodedBufferWriter {
   }
 
   void writeField(int fieldNumber, int fieldType, fieldValue) {
-    final valueType = fieldType & ~0x07;
+    final valueType = PbFieldType._baseType(fieldType);
 
     if ((fieldType & PbFieldType._PACKED_BIT) != 0) {
       if (!fieldValue.isEmpty) {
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
LGTM

osa1:
One of the CI jobs failed when fetching SDK:

```
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to storage.googleapis.com:443 
```

I restarted the CI jobs now.

mraleph (approved):

szakarias (approved):
Nice finding.

PR #597, opened 2022-04-04
Tweak BuilderInfo and FieldInfo docs:

osa1:
- Fix documentation syntax, add references to other classes, properties,
  methods

- Document undocumented properties

- Remove references to non-existent identifiers

- Some rewording

### DIFF
--- a/protobuf/lib/src/protobuf/field_info.dart
+++ b/protobuf/lib/src/protobuf/field_info.dart
@@ -8,13 +8,26 @@ part of protobuf;
 class FieldInfo<T> {
   FrozenPbList<T>? _emptyList;
 
-  /// Name of this field as the `json_name` reported by protoc.
-  ///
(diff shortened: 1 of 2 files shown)


protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -13,39 +13,49 @@ class FieldInfo<T> {
   /// This will typically be in camel case.
   final String name;
 
-  /// The name of this field as written in the proto-definition.
+  /// Name of this field as written in the proto definition.
```
sigurdm, line 16: Would be nice with an example also here and above :)

sigurdm (approved):
LGTM

osa1:
@sigurdm addressed your comments. It would be great if you could proof read 😄

sigurdm (approved):
LGTM

PR #600, opened 2022-04-06
Make `MapFieldInfo` key and value types non-nullable

osa1:


sigurdm (approved):
Code LGTM - but we need to update the protobuf version number and the protobuf constraint in protoc_plugin.

protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -219,10 +219,8 @@ class BuilderInfo {
   }
 
   // Map field.
-  void m<K, V>(int tagNumber, String name,
+  void m<K, V>(int tagNumber, String name, int keyFieldType, int valueFieldType,
```
sigurdm, line 222: To avoid breaking the protobuf version we could keep the types named.
```
void m<K, V>(int tagNumber, String name,
     {String? entryClassName,
      required int keyFieldType,
      required int valueFieldType,
      CreateBuilderFunc? valueCreator,
      ValueOfFunc? valueOf,
      List<ProtobufEnum>? enumValues,
```

Because we never generated missing values here, and this api is only intended for generated code I would argue that that is non-breaking

osa1: Done.

osa1:
> Code LGTM - but we need to update the protobuf version number and the protobuf constraint in protoc_plugin.

`protobuf` version bumped in #619. Changes in `protoc_plugin` are reverted with 9cbecd0 and old generated code is now compatible with the protobuf library in this PR.

osa1:
Ported to internal in cl/449693712.

PR #602, opened 2022-04-07
Remove redundant assertions in `MapFieldInfo` constructor

osa1:


osa1:
This will also need a version bump. I'm thinking doing that in a separate PR and rebasing this one and #600.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -231,26 +231,16 @@ class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
...
-      this.valueFieldType,
-      this.mapEntryBuilderInfo,
-      this.valueCreator,
-      {ProtobufEnum? defaultEnumValue,
-      String? protoName})
-      : super(name, tagNumber, index, type,
+  MapFieldInfo(String name, int tagNumber, int index, this.keyFieldType,
+      this.valueFieldType, this.mapEntryBuilderInfo, this.valueCreator,
+      {ProtobufEnum? defaultEnumValue, String? protoName})
+      : super(name, tagNumber, index, PbFieldType._MAP,
```
sigurdm, line 237: Personally I prefer trailing comma for long argument lists:
```
  MapFieldInfo(
    String name,
    int tagNumber,
    int index,
    this.keyFieldType,
    this.valueFieldType,
    this.mapEntryBuilderInfo,
    this.valueCreator, {
    ProtobufEnum? defaultEnumValue,
    String? protoName,
  });
```
YMMV

osa1:
We discussed this with @sigurdm a few days ago. These assertions currently always hold as the constructor has one call site and creates and passes the right value to the constructor, but if we remove it and then refactor the code it's possible that we will be passing incorrect values to the constructor. So the assertions are useful.

protobuf/lib/src/protobuf/field_info.dart
osa1: Huh.. didn't know I get to choose whether to put trailing comma or not when using `dart format`. FTR I also prefer trailing commas in long lists.

PR #603, opened 2022-04-07
Document map key and value fields

osa1:
Assertions to check key and value types will be added in another PR.

### DIFF
--- a/protobuf/lib/src/protobuf/field_info.dart
+++ b/protobuf/lib/src/protobuf/field_info.dart
@@ -222,11 +222,11 @@ String _unCamelCase(String name) {
 
 class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
   /// Key type of the map. Per proto2 and proto3 specs, this needs to be an
-  /// integer type or `string`.
+  /// integer type or `string`, and the type cannot be `repeated`.
   final int? keyFieldType;
 
   /// Value type of the map. Per proto2 and proto3 specs, this can be any type
-  /// other than `map` and `repeated`.
+  /// other than `map`, and the type cannot be `repeated`.
   final int? valueFieldType;
 
   /// Creates a new empty instance of the value type.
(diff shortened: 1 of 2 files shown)


protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -221,7 +221,12 @@ String _unCamelCase(String name) {
 }
 
 class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
+  /// Key type of the map. Per proto2 and proto3 specs, this needs to be an
+  /// integer type or `string` -- i.e. no floats or `bytes`.
   final int? keyFieldType;
+
+  /// Value type of the map. Per proto2 and proto3 specs, this can be any type
+  /// other than `map`.
```
sigurdm, line 229: I don't think it can be repeated?

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -221,7 +221,12 @@ String _unCamelCase(String name) {
 }
 
 class MapFieldInfo<K, V> extends FieldInfo<PbMap<K, V>?> {
+  /// Key type of the map. Per proto2 and proto3 specs, this needs to be an
```
sigurdm, line 224: Perhaps write that it is interpreted as the FieldInfo.type field (if that is indeed the case).

protobuf/lib/src/protobuf/field_info.dart
osa1: Fixed. I think the spec considers repeated as a modifier or attribute rather than type, but this is not clear in the current `field_type.dart` where everything is a bit, multiple type bits can be set etc. It will be clear with #605 so I will update this comment in #605 to say something like "this can be any type other than `map`, and the type cannot be repeated".

protobuf/lib/src/protobuf/field_info.dart
osa1: Done.

PR #604, opened 2022-04-07
Remove redundant check in PbMap equality check

osa1:
Since we check that both maps have same number of keys, the check
comparing each key in one of the maps with the value of the key in the
other map is enough to make sure both maps have the same set of keys.


sigurdm (approved):
Nice catch!

LGTM

PR #606, opened 2022-04-11
Make `protoName` unCamelCase lazy

jiahaog:
In a large 1p Flutter android application, package:protobuf accounts for ~220ms (8%) of the CPU time during startup. This `_uncamelCase` operation takes up ~25ms, or 11% of this 220ms which occurs whenever a proto is constructed.

Looking through the code, it appears that this `protoName` field is only needed for conversion to JSON which is less common. As such, make it a lazy getter to avoid this.

[Internal link](http://pprof/user-profile?id=d8d43f484e3e7b4064a29737cbdabf77&tab=flame&flamesearch=unCamel&pivot=Precompiled_____unCamelCase_19070143_4307) to pivot profile view which highlights this method (top half shows callers, bottom half shows callees)

### DIFF
--- a/protobuf/lib/src/protobuf/field_info.dart
+++ b/protobuf/lib/src/protobuf/field_info.dart
@@ -29,7 +29,12 @@ class FieldInfo<T> {
   /// ```
   /// `protoName` for the `result_per_page` field above is `"result_per_page"`.
   /// This will typically consist of words separated with underscores.
-  final String protoName;
+  String get protoName {
+    _protoName ??= _unCamelCase(name);
+    return _protoName!;
+  }
+
+  String? _protoName;
 
   /// Field number as specified in the proto definition. Example:
   /// ```proto
@@ -80,7 +85,7 @@ class FieldInfo<T> {
       String? protoName})
       : makeDefault = findMakeDefault(type, defaultOrMaker),
         check = null,
-        protoName = protoName ?? _unCamelCase(name),
+        _protoName = protoName,
         assert(type != 0),
         assert(!_isGroupOrMessage(type) ||
             subBuilder != null ||
@@ -90,7 +95,7 @@ class FieldInfo<T> {
   // Represents a field that has been removed by a program transformation.
   FieldInfo.dummy(this.index)
       : name = '<removed field>',
-        protoName = '<removed field>',
+        _protoName = '<removed field>',
         tagNumber = 0,
         type = 0,
         makeDefault = null,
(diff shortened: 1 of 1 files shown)


protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -29,7 +29,12 @@ class FieldInfo<T> {
   /// ```
   /// `protoName` for the `result_per_page` field above is `"result_per_page"`.
   /// This will typically consist of words separated with underscores.
-  final String protoName;
+  String get protoName {
+    _protoName ??= _unCamelCase(name);
+    return _protoName!;
```
osa1, line 34:
```suggestion
    return _protoName ??= _unCamelCase(name);
```

osa1 (approved):
Thanks for the PR @jiahaog.

> This _uncamelCase operation takes up ~25ms, or 11% of this 220ms which occurs whenever a proto is constructed.

What are the numbers like with this PR? I'm wondering how much difference this change makes.

LGTM assuming it makes a difference in the numbers.

osa1:
> I ran the startup benchmark with this change and we see a 17ms or ~0.3% in one of the tracked metrics

Just trying to understand the benchmark here -- in the PR description it says "~220ms (8%)", so the entire startup is 2750ms. If we reduce that 220ms to 17ms the startup time becomes 2547ms, and 17ms of 2547ms is 0.6%, not 0.3%. Am I missing anything?

Not that it matters too much, the result looks great..

---

It seems like the suggestion I made was made using my gmail account. Could you maybe squash the commits with only you as the author? Or you could use `omersa@google.com` as my email. (For the future suggestions I will see if I can make suggestions with my Google account)

jiahaog:
Sorry for being unclear. I'll try and elaborate further:

The original profile for startup I captured locally (aggregated across 3 local runs) showed that `package:protobuf` took 220ms in total as part of this profile. This is 8% so the entire startup (aggregated) is indeed 2750ms. In this profile, we can see that the uncamelCase takes ~25ms.

I then patched this PR into Google3 and ran a similar benchmark in the device lab which is supposedly more accurate (and is what we use to track the historic performance of the app across different CLs). This showed an improvement of 17ms or ~0.3% against another run of the same benchmark without this change patched for a related metric. I used the device lab instead of doing it locally because the performance of my uncalibrated local device can fluctuate quite a bit and the device lab provides much more precision. Hope this clears things up, also happy to share links to the internal benchmark runs internally :slightly_smiling_face:

osa1:
Ah, OK, so this PR improves runtime 0.3% in a benchmark. Thanks for the clarification. I will document this in the commit message so that we will know why it was made lazy.

I'd like to wait for @sigurdm or @mraleph's approval, but I think they're not available until next week so this may have to wait a litte bit.

jiahaog:
Sure, thanks!

mraleph (approved):
I think this is fine. This field has very limited usage (I did internal Code Search) and I am somewhat surprised that it even exists in the first place.

mraleph:
@jiahaog feel free to create CL internally with the same change, package is currently out of sync with internal copy so you have to do backport manually.

PR #610, opened 2022-04-13
Tweak READMEs:

### LINKED ISSUE #529: Something to suggest about guide
1. tip that "we should use protoc_plugin"
image
In https://pub.dev/packages/protobuf, there are five urls, but not one is https://pub.dev/packages/protoc_plugin.
Add some words link to https://pub.dev/packages/protoc_plugin, first, please. So we can find that we should use protoc_plugin quickly and then we will use "dart pub global activate protoc_plugin".

2. it's hard to find the path of  "the dart executable"
image

I am happy when I have confirm that both the dart executable and bin/protoc-gen-dart are in the path, however, when I use 
protoc --cpp_out=./cpp --dart_out=./dart try.proto
it tips that these is not "pub" command.
Then I guess if it means the "dart pub" or "flutter pub"?
So I add these path to system path, in flutter\bin\cache\dart-sdk\bin.
image
There is a dart.exe and pub.bat, and it works.

So now I know that "the dart executable" is not what I think first in flutter\bin, but the one "dart.exe" in  flutter\bin\cache\dart-sdk\bin.
image

So I suggest it's better to tip that add flutter\bin\cache\dart-sdk\bin to path, which is hard to find.

Thanks.

### LINKED ISSUE #576: protoc Plugin Build Instructions Clarification
Is the plugin actually standalone?  Not sure if I'm missing something or have a misunderstanding.

This is from [Wikipedia](https://en.wikipedia.org/wiki/Standalone_program):
> A stand-alone program, also known as a freestanding program, is a computer program that does not load any external module, library function or program and that is designed to boot with the bootstrap procedure of the target processor – it runs on bare metal.

After [building the plugin](https://github.com/google/protobuf.dart/tree/master/protoc_plugin#how-to-build-and-use), the executable is a script that calls dart, so this is not standalone.

If I am correct so far, is there a way to build a standalone binary (no dependencies)?

osa1:
- Fix typos

- Shorten some long lines in markdown files using footnote-style
  markdown links

- Use `#` header syntax instead of a mix of `#` and underscores: it's
  unclear whether `---` is smaller than `##` or not. Using `#` for all
  headers is also more consistent and makes it easier to read in
  markdown format (rather than in rendered).

- Add syntax highlighting to Dart and proto code examples.

- Split "How to build and use" section into "How to build" and "How to
  use" sections.

- Move `dart pub activate` method up and make it the recommended way of
  installing the plugin.

- Describe how to build standalone native executables for the plugin.

- Addresses #576:
  - For each installation method clarify whether they generate a Dart
    script or standalone native executable.

- Addresses: #529:

  - Mention that latest version of Dart may be needed. Older versions
    don't have `dart pub` command and can cause errors when following
    installation instructions.

  - Link to protoc_plugin pub.dev page in protobuf README.

  - Link to Dart installation instructions in protoc_plugin README.

Fixes #576
Fixes #529


osa1:
@mraleph @sigurdm please feel free to add reviewers.

protoc_plugin/README.md
```diff
@@ -2,64 +2,88 @@
...
+This repository provides a plugin for the [protoc compiler][1]. It generates
+Dart files for working with data in protocol buffers format.
 
-Requirements
-------------
+[1]: https://developers.google.com/protocol-buffers/docs/cpptutorial#compiling-your-protocol-buffers
+
+## Requirements
 
 To compile a .proto file, you must use the 'protoc' command which is [installed
```
mraleph, line 12: nit: maybe `` `protoc` `` rather than `'protoc'`?

protoc_plugin/README.md
```diff
@@ -2,64 +2,88 @@
...
-Protobuf 3.0.0 or above is required.
+separately][1]. protoc 3.0.0 or above is required.
 
-The generated files are pure Dart code that run in either in the Dart VM or in a
-browser (using dart2js). They depend the [protobuf Dart
-package](https://pub.dev/packages/protobuf). A Dart project that
-includes generated files should add "protobuf" to its pubspec.yaml file.
+The generated files are pure Dart code that run either in the Dart VM or in a
+browser (using [dart2js][2]). They depend on the [protobuf Dart package][3]. A
+Dart project that includes generated files should add "protobuf" as a
```
mraleph, line 17: consider ``should add `protobuf` package ``

protoc_plugin/README.md
```diff
@@ -2,64 +2,88 @@
...
+dependency in the pubspec.yaml file.
```
mraleph, line 18: maybe `` `pubspec.yaml` ``

mraleph (approved):
LGTM

kevmoo (approved):

PR #617, opened 2022-04-20
protoc_plugin README fixes and tweaks:

### LINKED ISSUE #208: protoc_plugin readme is outdated
The example should include `syntax "proto2";`.

The command-line example refers to test.proto. The text should clarify that this file needs to contain the proto definition above.

Should probably point to the tutorial.

osa1:
- Fix proto syntax

- Add `syntax = "proto3"` in proto examples. proto3 is used instead of
  proto3 is newer. Removed `optional`s.

- Minor rewording

Closes #208

### DIFF
--- a/protoc_plugin/README.md
+++ b/protoc_plugin/README.md
@@ -64,7 +64,7 @@ your `PATH` with name `protoc-gen-dart`, or pass the path to it to `protoc`'s
 ## How to use
 
 Once you have `protoc-gen-dart` in the `PATH` the protocol buffer compiler can
-be invoked like this:
+be invoked like this to generate Dart for the proto file `test.proto`:
 
     $ protoc --dart_out=. test.proto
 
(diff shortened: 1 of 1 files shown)


protoc_plugin/README.md
```diff
@@ -106,21 +106,25 @@ re-export the relevant protocol buffer classes.
 Say we have the file `m1.proto` with the following content
 
 ```proto
+syntax = "proto3";
+
 message M1 {
-  optional string a;
+  optional string a = 1;
```
sigurdm, line 112: I think in proto3 the default is to not put `optional`, the field will still not be required.

sigurdm (approved):

protoc_plugin/README.md
osa1: I wasn't sure whether to delete or keep it. Should I remove it?

sigurdm: I think so

PR #619, opened 2022-04-21
Add changes since the last protobuf release to CHANGELOG

osa1:


protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## Unreleased
+
+* Update READMEs of `protobuf` and `protoc_plugin`:
+  * Use `dart pub` instead of `pub` in command examples (a7e75cb)
+  * Fix typos, clarify installation instructions, mention native compilation
+    for `protoc_plugin` (#610)
```
sigurdm, line 6: I think adding issue numbers to the bullets is great (I have not been doing that) perhaps make them full links somehow, show they will be clickable!

protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## Unreleased
+
+* Update READMEs of `protobuf` and `protoc_plugin`:
+  * Use `dart pub` instead of `pub` in command examples (a7e75cb)
```
sigurdm, line 4: Don't we have a PR to point to here instead of the commit?

sigurdm (approved):

protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## Unreleased
```
mraleph, line 1: I think the typical pattern we use in other repos is to start a `-dev` version whenever you make a change by updating `pubspec.yaml` and `CHANGELOG.md`, e.g. in this case `protobuf` package is currently `2.0.2-dev`. You can update `pubspec.yaml` to use this version and then replace `Unreleased` with `2.0.2-dev`.

See https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change

At some point I thought about creating a CI check to enforce this (e.g. if you have PR with changes to `lib/` you need to have version set to `-dev`). - but I did not get time.

protobuf/CHANGELOG.md
osa1: No, that commit was made to `master` directly.

protobuf/CHANGELOG.md
osa1: > perhaps make them full links somehow, show they will be clickable!

I did this but the process is tedious and file size will be doubled as every line has at least one link. Do you want to keep it?

sigurdm: Yes, personally i think it is nice, and don't mind the file size increase.

I wonder if a vim macro or something similar could make the tedium smaller (You could also use inline links).

Do what you feel is better, issue numbers is already nice!

osa1: OK. It will be written once and read many times so I'm fine with extra lines if it makes reading easier.

sigurdm: I think it especially makes it easier in the rendered display: eg. https://pub.dev/packages/protobuf/changelog#013161

Also there's a confusion between issue numbers and pr numbers on github. I'm not sure if `#1` cannot refer to both...

protobuf/CHANGELOG.md
osa1: > But it's still technically a breaking change

I think we should put the reserved names back even though we don't need to reserve them, to keep things backwards compatible. Then add a note and open an issue to remove then in the next major release. I will do that in a separate PR.

osa1: Done in #636.

osa1: - Reverted the breaking change
- Removed changelog entry for the breaking change
- Bumped package verson to 2.0.2-dev

PR #625, opened 2022-04-22
Disambiguate names of service methods in grpc output

sigurdm:


osa1:
This should fix #485.

PR #626, opened 2022-04-23
Simplify PbList implementation, fix freezing

### LINKED ISSUE #624: `PbList.toFrozenPbList` does not freeze the list in-place, allows modifying frozen messages
Discovered this while reading the code.

```proto
syntax = "proto3";

message Test {
  repeated int32 ints = 1;
}
```

```dart
import '../test.pb.dart';

void main() {
  Test test1 = Test.create();
  test1.ints.add(123);

  var ints = test1.ints;

  test1.freeze();
  ints.add(456);

  print(test1);
}
```

Output:

```
ints: 123
ints: 456
```

Expected output: it should fail with an exception.

`PbList` should implement freezing the same way `PbMap` implements it: set a bool flag in `PbList`, mark elements of the list as frozen.

osa1:
- `PbList` is now a single class

- Freezing no longer creates a new list, sets `_isReadOnly` flag of the current
  list

- Freezing `PbList` now freezes elements

`PbList` implementation is now consistent with `PbMap`.

Fixes #624.

---

cl/450278877


protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -6,310 +6,216 @@ part of protobuf;
...
-    if (newLength > length) {
-      throw UnsupportedError('Extending protobuf lists is not supported');
+  bool operator ==(other) => (other is PbList) && _areListsEqual(other, this);
+
+  @override
+  int get hashCode => _HashUtils._hashObjects(_wrappedList);
+
+  void freeze() {
+    _isReadOnly = true;
+    if (_isGroupOrMessage(_elementType)) {
```
osa1, line 215: This part is buggy. It was copied from `PbMap` which doesn't allow `map` or `repeated` values.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -6,7 +6,7 @@ part of protobuf;
 
 /// An object representing a protobuf message field.
 class FieldInfo<T> {
-  FrozenPbList<T>? _emptyList;
+  // FrozenPbList<T>? _emptyList;
```
sigurdm, line 9: Don't leave commented out code. Just delete it.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -135,9 +135,9 @@ class FieldInfo<T> {
   /// Returns a read-only default value for a field. Unlike
   /// [GeneratedMessage.getField], doesn't create a repeated field.
   dynamic get readonlyDefault {
-    if (isRepeated) {
-      return _emptyList ??= FrozenPbList._([]);
-    }
+    // if (isRepeated) {
```
sigurdm, line 138: Same here.

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -6,310 +6,216 @@ part of protobuf;
...
-  void fillRange(int start, int end, [E? fill]) =>
-      throw _unsupported('fillRange');
-}
+class PbList<E> extends ListBase<E> {
+  final List<E> _wrappedList;
+  final CheckFunc<E> _check;
 
-class PbList<E> extends PbListBase<E> {
-  PbList({CheckFunc<E> check = _checkNotNull}) : super._noList(check: check);
+  // TODO: Just store if the element type is a group or message?
```
sigurdm, line 13: Always leave a name on the todos:
```suggestion
  // TODO(omersa): Just store if the element type is a group or message?
```

protobuf/lib/src/protobuf/pb_list.dart
sigurdm: As far as I can tell You can not have a repeated map: https://developers.google.com/protocol-buffers/docs/proto3#maps

> Map fields cannot be repeated.

I also don't think you can have a `repeated` of a `repeated`.

So maybe this is good enough.

sigurdm (approved):
Nice!

LGTM

protobuf/lib/src/protobuf/pb_list.dart
osa1: You are right! I added a line of comment there to clarify.

protobuf/lib/src/protobuf/field_info.dart
osa1: I wanted to enable this line later. I've done it now.

protobuf/lib/src/protobuf/field_info.dart
osa1: I wanted to enable this later. I still don't understand why this line is needed, but decided to not change more than necessary for now.

protobuf/lib/src/protobuf/pb_list.dart
osa1: I think storing the full type should be fine. In the lifetime of a `PbList` we only read this field once so it doesn't matter too much. Removed the TODO.

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -6,310 +6,217 @@ part of protobuf;
...
-  set length(int newLength) {
-    if (newLength > length) {
-      throw UnsupportedError('Extending protobuf lists is not supported');
+    // Per spec `repeated map<..>` and `repeated repeated ..` are not allowed
+    // so we only check for messages
+    if (_isGroupOrMessage(_elementType)) {
+      for (var elem in _wrappedList) {
+        (elem as GeneratedMessage).freeze();
+      }
     }
```
osa1, line 220: Another implementation of this would be:

```suggestion
    if (_wrappedList.isNotEmpty && _wrappedList[0] is GeneratedMessage) {
      for (var elem in _wrappedList) {
        (elem as GeneratedMessage).freeze();
      }
    }
```

Now we don't need to store the element type anymore.

osa1:
I refactored the implementation a little bit and I think this version looks good.

Only problem is removing `FrozenPbList` and `PbListBase` is a breaking change.

We could fix the two bugs (freezing the list allows updates in aliased lists, freezing the list does not freeze elements) without removing any classes, but it would require copying the mutable list (deep or shallow depending on the intended semantics) when creating a `FrozenPbList`. That will cause performance issues. The current PR should solve the issue with minimal runtime overhead.

I say we keep this PR for the next major release (maybe we can create a branch for it). Fixing the bug without changing the API requires more work in runtime.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -6,7 +6,7 @@ part of protobuf;
 
 /// An object representing a protobuf message field.
 class FieldInfo<T> {
-  FrozenPbList<T>? _emptyList;
+  PbList<T>? _emptyList;
```
sigurdm, line 9: Is this still always frozen?
Maybe add a comment about that?

osa1: It is frozen, yes. Otherwise you could modify the "empty list" and calls to `readonlyDefault` would return a list with elements.

I'll document it.

osa1: @sigurdm PTAL.

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -6,310 +6,222 @@ part of protobuf;
...
-  int indexOf(Object? element, [int start = 0]) =>
-      _wrappedList.indexOf(element as E, start);
-
-  // TODO(jakobr): E instead of Object once dart-lang/sdk#31311 is fixed.
-  @override
-  int lastIndexOf(Object? element, [int? start]) =>
-      _wrappedList.lastIndexOf(element as E, start);
+  void operator []=(int index, E value) {
+    if (_isReadOnly) {
+      throw UnsupportedError('Setting field of a read-only list');
```
rakudrama, line 200: field -> element?

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -6,310 +6,222 @@ part of protobuf;
...
-  /// Freezes the list by converting to [FrozenPbList].
-  FrozenPbList<E> toFrozenPbList() => FrozenPbList<E>.from(this);
+  PbList.from(List from)
+      : _wrappedList = List<E>.from(from),
+        _check = _checkNotNull;
 
   @override
   void add(E element) {
-    check(element);
+    if (_isReadOnly) {
```
rakudrama, line 32: The use of a helper can help control code size.

I would use two levels: (1) a helper that does the check and is inlined (2) a second helper that throws and is never inlined

```dart
Never _readOnlyError(String methodName) {
  throw UnsuppportedError("'$name' on a read-only list");
}

void _checkModifiable(String methodName) {
  if (_isReadOnly) _readOnlyError(methodName);
}

void add(E element) {
  _checkModifiable('add');
}
```

dart2js has a heuristic to not inline methods that only throw.
For AOT you might need `@pragma`s

There is now only one place where there is the combination of calling a constructor and throwing.

osa1:
GitHub's "request review" button doesn't seem work for non-collaborators.. @rakudrama PTAL.

sigurdm (approved):
LGTM

PR #630, opened 2022-04-25
Minor fix in query_bench set_nested_value benchmark:

osa1:
`rebuild` shallow clones the message and returns an updated message, so
return value should be used.

Thanks to @sigurdm for noticing this.


sigurdm (approved):
LGTM

PR #631, opened 2022-04-25
Add `UseResult` annotation to `rebuild`

osa1:
This catches issues like #630 where we forgot to use return value of
`rebuild`. Since `rebuild` does not update in-place, not using the
return value is often a bug.

This adds `meta` as a dependency but it should not affect generated
code. We may also want to use `meta` in the future for things like:

- Annotating `GeneratedMessage` methods that are supposed to be used by
  the generated code with `@protected`.

- Annotating generated message classes with `@sealed`. This help with
  maintaining backwards compatibility when adding new members to
  generated messages or to `GeneratedMessage`.

### DIFF
--- a/protobuf/lib/protobuf.dart
+++ b/protobuf/lib/protobuf.dart
@@ -11,6 +11,7 @@ import 'dart:math' as math;
 import 'dart:typed_data' show TypedData, Uint8List, ByteData, Endian;
 
 import 'package:fixnum/fixnum.dart' show Int64;
+import 'package:meta/meta.dart' show UseResult;
 
 import 'src/protobuf/json_parsing_context.dart';
 import 'src/protobuf/permissive_compare.dart';
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -558,6 +558,8 @@ extension GeneratedMessageGenericExtensions<T extends GeneratedMessage> on T {
   ///
   /// Makes a writable shallow copy of this message, applies the [updates] to
   /// it, and marks the copy read-only before returning it.
+  @UseResult('[GeneratedMessageGenericExtensions.rebuild] '
+      'does not update the message, returns a new message')
   T rebuild(void Function(T) updates) {
     if (!isFrozen) {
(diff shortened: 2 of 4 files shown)


sigurdm (approved):
LGTM

Seems there is a test that needs to be fixed.

Maybe @rakudrama has something to say about adding the dependency.

I think package:meta is so ubiquitous that package:protobuf can afford to depend on it.

osa1:
> Seems there is a test that needs to be fixed.

Hm.. I'm a bit surprised that this line generates a `unused_result` warning: https://github.com/google/protobuf.dart/blob/2043b1438a709b0f440e8e19d73b344fad80ae74/protoc_plugin/test/oneof_test.dart#L164

The result is clearly used. It's casted to `Foo` and then assigned to `copy1`. Is this an SDK bug?

osa1:
> Hm.. I'm a bit surprised that this line generates a unused_result warning: 

Reported this as https://github.com/dart-lang/sdk/issues/48879.

osa1:
The issue with the warning described above is fixed in Dart `main` branch but it will be a while until it's released. I'm not sure whether to wait for the release with the fix before merging this or work around the issue.

sigurdm:
> The issue with the warning described above is fixed in Dart main branch but it will be a while until it's released. I'm not sure whether to wait for the release with the fix before merging this or work around the issue.

I think it is ok to add a `// ignore:` comment in the test.

osa1:
@rakudrama any thoughts on adding `meta` as a dependency to `protobuf`?

osa1:
> Side comment - rebuild is generic, so why do we need the cast?

This is because `GeneratedMessage.freeze`'s return type is `GeneratedMessage`: https://github.com/google/protobuf.dart/blob/9c73821d5dc9a29b137654360a2e9300fe2b8fd6/protobuf/lib/src/protobuf/generated_message.dart#L54-L57 so the concrete type is lost after `freeze`.

It would be great if we could use a type like `This` to refer to the concrete type in a superclass. If we had that `freeze` could have `This` as return type, and calls to `freeze` would not upcast the receiver.

osa1:
This revealed another SDK bug when updating a protobuf user to the current master branch: https://github.com/dart-lang/sdk/issues/49023

Issue #609, opened 2022-04-12
Clarify compatibility between protoc_plugin and protobuf versions


osa1:
READMEs of [protobuf][1] and [protoc_compiler][2] do not say anything about the compatibility of these packages, i.e. for a version of one of those packages which version(s) of the other package I should be able to use.

It's also not helping that

- These libraries do not have the same version numbers. Current protoc_plugin is 20.0.1, protobuf is 2.0.1. Some of the previous versions of protoc_plugin are 19.3.1, 19.0.1 etc. and there are no similar looking version number in protobuf. We have 1.1.1, 0.14.4, 0.13.16 etc. It's impossible to tell what is compatible with what.

- 1-1 mapping is not possible. For example, protobuf-1.1.3 was released 13 months ago, but no protoc_plugin release was made at that time.

As a user, when the latest versions of protobuf and protoc_plugin work for me this is not a problem. But if I want to bump version numbers (e.g. newer versions have a bug fix) but not to the latest version (e.g. latest version is backwards incompatible with the version I'm using) then I'm in trouble as for the backwards-compatible version with the bug fix of one of these libraries I don't know which version of the other library I should be using.

[1]: https://github.com/google/protobuf.dart/blob/ded1ac7d2e00b1f8ff7e9ac4129a3be6c652ee94/protobuf/README.md
[2]: https://github.com/google/protobuf.dart/blob/ded1ac7d2e00b1f8ff7e9ac4129a3be6c652ee94/protoc_plugin/README.md

PR #634, opened 2022-04-26
Remove unused `GeneratedMessage` constructors

osa1:
`fromBuffer` and `fromJson` constructors can only be called in
`GeneratedMessage` subclasses, and we stopped calling them in #253
(merged as 481380e). They're now unused and unusable so it's safe to
remove them.

### DIFF
--- a/protobuf/lib/meta.dart
+++ b/protobuf/lib/meta.dart
@@ -26,8 +26,6 @@ const GeneratedMessage_reservedNames = <String>[
   'eventPlugin',
   'extensionsAreInitialized',
   'freeze',
-  'fromBuffer',
-  'fromJson',
   'getDefaultForField',
   'getExtension',
   'getField',
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -29,19 +29,6 @@ abstract class GeneratedMessage {
     if (eventPlugin != null) eventPlugin!.attach(this);
   }
 
-  GeneratedMessage.fromBuffer(
-      List<int> input, ExtensionRegistry extensionRegistry) {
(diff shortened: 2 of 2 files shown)


sigurdm (approved):
LGTM

I guess this is technically breaking, but I don't think we want to maintain compatibility with clients making hand-written GeneratedMessage subclasses.

Oh and we should make sure that we are no longer compatible with the old protoc_plugin versions that actually used these.

osa1:
> I guess this is technically breaking, but I don't think we want to maintain compatibility with clients making hand-written GeneratedMessage subclasses.

Right. FWIW, I tested this change locally and it didn't break anything in my testing.

> Oh and we should make sure that we are no longer compatible with the old protoc_plugin versions that actually used these.

How do we do that?

protoc_plugin/protobuf compatibility story is a bit unclear at the moment (#609). I think ideally there would be one package and you would always use compatible plugin and runtime library. We can manually do this by using same version numbers in `protoc_plugin` and `protobuf`, bumping them in sync (even when one of them don't have any updates), and documenting that users should use the same version of these packages. I use this approach in a project of mine and it works fine, but it's a much simpler project with a few users. It's simpler for both maintainers and the users.

sigurdm:
>> Oh and we should make sure that we are no longer compatible with the old protoc_plugin versions that actually used these.

> How do we do that?

Each protoc_plugin version has a dependency on protobuf with a compatible version-range. I think because a breaking release protobuf 2.0.0 came out old protoc_plugin version should not be compatible with that. The constructor became unused in protoc_plugin 17.0.0: https://github.com/google/protobuf.dart/pull/253/files and the previous version https://pub.dev/packages/protoc_plugin/versions/16.0.7/pubspec has a dependency on `protobuf ^0.13.10` So I think we are fine here.

PR #638, opened 2022-05-03
Treat empty and uninitialized Maps the same in equality checks

osa1:
Treat an empty Map the same as an uninitialized Map when performing equality
checks. This ensures that reading a map field does not alter the hashCode or
equality comparisons of the message itself.

-- 

Rebased version #591 with a regression test.


sigurdm (approved):
Nice

LGTM

osa1:
Maintainer's note: internally an identical patch was implemented in cl/304375641.

PR #640, opened 2022-05-04
Fix sharing coded buffer bytes when parsing `bytes` fields

osa1:
Synced from internal repo

---

I'm not sure if this is the right fix for this bug. Should `readBytes()` return a new list (instead of a view) maybe?


sigurdm (approved):
LGTM

PR #643, opened 2022-05-09
Remove frozen message modification handling

### LINKED ISSUE #175: Remove frozenMessageModificationHandler
Currently we have `frozenMessageModificationHandler` that is used to throw when a frozen message is changed. It can be customized.

`FrozenPbList` and a frozen `PbMap` do not use this, but will just throw.
This is inconsistent. And it is not clear to me how much is gained by the hook.

I suggest we remove the handler from the public API, and always just throw.

osa1:
See #175 for the problems with this API.

Internally this API is not used. However this is still a breaking change
and will be released with the next major release.

Fixes #175

---

Synced to internal in cl/450254014.


sigurdm (approved):
LGTM

PR #644, opened 2022-05-10
Remove ReadonlyMessageMixin

### LINKED ISSUE #183: Remove ReadonlyMessageMixin

osa1:
This mixin is not used internally. On GitHub there are references to
this symbol, but they are all either (1) copies/forks of this library
(2) old protoc_plugin-generated code. So removing this shouldn't cause
any issues as long as users update the plugin with the library.

Closes #183

---

This is another breaking change (with #643, #626). Should we create a branch for the next major release maybe?


sigurdm (approved):
LGTM

PR #645, opened 2022-05-11
Fix compile-protos.sh interpreter

osa1:
This script is not sh and it fails to run under some OSes:

     $ ./compile-protos.sh
     ./compile-protos.sh: 8: Syntax error: "(" unexpected

Using bash as the interpreter fixes the problem.


sigurdm (approved):
LGTM

PR #647, opened 2022-05-11
Remove `ProtobufContainer` superclass of `CodeGenerator`

osa1:
CodeGenerator is not a ProtobufContainer and cannot implement
ProtobufContainer methods. It also doesn't need to be a
ProtobufContainer. Remove superclass and dummy methods.


sigurdm (approved):
LGTM

PR #648, opened 2022-05-16
Align the hashCode of a message with an empty unknown field-set with that of no unknown field set

osa1:
Port of cl/308005673


sigurdm (approved):
LGTM

PR #651, opened 2022-05-16
Convert the field started with underscore and digit to a legal name

osa1:
In proto, it is allowed to name a field started with underscore and
digit like `_3d`. But it's illegal in Dart. Convert the field name when
generating code.

Port of cl/411592880


sigurdm (approved):
LGTM

PR #656, opened 2022-05-18
remove the top level analysis_options.yaml file

devoncarew:
- remove the top level analysis_options.yaml file from the repo
- add separate analysis options  files - based on `package:lints` to each of the api_benchmark and query_benchmark packages
- indicate in those package's pubspecs that they are not intended to be published

cc @osa1


osa1 (approved):
LGTM, thanks!

PR #660, opened 2022-05-24
Remove invalid check in `_FieldSet.writeFieldValue`

osa1:
A field value is one of these:

- PbListBase (for `repeated`, PbList after #626)
- PbMap
- int
- double
- bool
- String
- List<int> (for `bytes`)
- GeneratedMessage

`ByteData` is not a valid field value type, so remove the `ByteData`
cases when checking field value types.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -726,11 +726,7 @@ class _FieldSet {
 
     void writeFieldValue(fieldValue, String name) {
       if (fieldValue == null) return;
-      if (fieldValue is ByteData) {
-        // TODO(skybrian): possibly unused. Delete?
-        final value = fieldValue.getUint64(0, Endian.little);
-        renderValue(name, value);
-      } else if (fieldValue is PbListBase) {
+      if (fieldValue is PbListBase) {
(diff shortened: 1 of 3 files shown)


sigurdm (approved):
LGTM

Nice cleanup - am I understanding correctly that protobuf bytes are never stored as `ByteData`, but rather as a `Uint8List` backed by `ByteData`?

osa1:
> Nice cleanup - am I understanding correctly that protobuf bytes are never stored as ByteData, but rather as a Uint8List backed by ByteData?

Protobuf `bytes` fields are never stored as `ByteData`, but they're also not always `Uint8List`. [When deserializing we use `Uint8List`][1], but [for the default value of `bytes` fields we use `<int>[]`][2]. No idea what the concrete type is in this case.

[1]: https://github.com/google/protobuf.dart/blob/0138eafab1d9e49bdf5007f1ae6496a7c945f9a9/protobuf/lib/src/protobuf/coded_buffer.dart#L56
[2]: https://github.com/google/protobuf.dart/blob/0138eafab1d9e49bdf5007f1ae6496a7c945f9a9/protobuf/lib/src/protobuf/field_type.dart#L74

PR #661, opened 2022-05-24
Implement fast path when merging fields of same message

### LINKED ISSUE #60: Cloning should avoid validation checks
When merging data from the same field (i.e. identical FieldInfo) we can avoid validation checks because the source must already have been validated. This will be especially beneficial for repeated fields.

osa1:
Fixes #60

---

@rakudrama do you have any benchmarks for this?

There are a few different ways to implement this. We could move the `sameMessage` check to the outside of the loop, or add a fast path check in `_mergeField`. Not sure which one will perform the best without benchmarks.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -768,14 +768,16 @@ class _FieldSet {
   /// in this message. Repeated fields are appended. Singular sub-messages are
   /// recursively merged.
   void _mergeFromMessage(_FieldSet other) {
-    // TODO(https://github.com/google/protobuf.dart/issues/60): Recognize
-    // when `this` and [other] are the same protobuf (e.g. from cloning). In
-    // this case, we can merge the non-extension fields without field lookups or
-    // validation checks.
-
+    final sameMessage = identical(_meta, other._meta);
     for (var fi in other._infosSortedByTag) {
       var value = other._values[fi.index!];
-      if (value != null) _mergeField(fi, value, isExtension: false);
+      if (value != null) {
+        if (sameMessage) {
+          _mergeNonExtensionFieldUnchecked(fi, value);
+        } else {
+          _mergeField(fi, value, isExtension: false);
+        }
+      }
     }
     if (other._hasExtensions) {
       var others = other._extensions!;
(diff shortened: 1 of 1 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -768,14 +768,16 @@ class _FieldSet {
   /// in this message. Repeated fields are appended. Singular sub-messages are
   /// recursively merged.
   void _mergeFromMessage(_FieldSet other) {
-    // TODO(https://github.com/google/protobuf.dart/issues/60): Recognize
-    // when `this` and [other] are the same protobuf (e.g. from cloning). In
-    // this case, we can merge the non-extension fields without field lookups or
-    // validation checks.
-
+    final sameMessage = identical(_meta, other._meta);
```
rakudrama, line 771: Perhaps `sameMessageType` is a little clearer

rakudrama:
> Fixes #60
> 
> @rakudrama do you have any benchmarks for this?
> 
> There are a few different ways to implement this. We could move the `sameMessage` check to the outside of the loop, or add a fast path check in `_mergeField`. Not sure which one will perform the best without benchmarks.

I'm also not sure which approach would be best.
I don't now of any benchmarks other than the Golem ones.
I think it would be worth auditing the existing benchmarks and perhaps adding some benchmarks that you know are representative of real app scenarios.

osa1:
I think we should be able to benchmark this by using `deepCopy`, which uses `mergeFromMessage`: https://github.com/google/protobuf.dart/blob/3e0f22915e2217bf3728f8d2b4e22ec09fa4ba1e/protobuf/lib/src/protobuf/generated_message.dart#L564

Currently we have 2 somewhat large messages in benchmarks:

- The message in query_benchmark is 246K. This message has lots of strings.
- google_message2 in protobuf_benchmarks is 83K. I don't know about the structure of this message.

We can add a benchmark that measures `deepCopy` of these messages.

Note: these benchmarks will be merged to benchmarks/ soon: #698

osa1:
I added a new benchmark to measure `deepCopy` time and this seems to significantly improve it.

AOT, master:

protobuf_deep_copy(RunTimeRaw): 625.190625 us.
protobuf_deep_copy(RunTimeRaw): 627.7799373040752 us.
protobuf_deep_copy(RunTimeRaw): 625.09375 us.

AOT, this PR:

protobuf_deep_copy(RunTimeRaw): 475.6320665083135 us.
protobuf_deep_copy(RunTimeRaw): 476.2588095238095 us.
protobuf_deep_copy(RunTimeRaw): 475.4239904988123 us.

Diff: -24%

JS, master:

protobuf_deep_copy(RunTimeRaw): 738.3763837638377 us.
protobuf_deep_copy(RunTimeRaw): 680 us.
protobuf_deep_copy(RunTimeRaw): 729.9270072992701 us.

JS, this PR:

protobuf_deep_copy(RunTimeRaw): 639.297124600639 us.
protobuf_deep_copy(RunTimeRaw): 669.2307692307693 us.
protobuf_deep_copy(RunTimeRaw): 659.2105263157895 us.

Diff: -8%

This should improve `clone` as the code generated for `clone` is identical to `deepCopy` implementation.

osa1:
We should also try implementing a `deepCopy` that is actually implemented as cloning, instead of merging the message to an empty message.

osa1:
> We should also try implementing a deepCopy that is actually implemented as cloning, instead of merging the message to an empty message.

This is done in #742 which halves AOT deepCopy benchmark runtime.

osa1:
Sigurd asked about whether it makes sense to allow merging messages with different types. I think it doesn't, and I tried adding this to the merge method and tested the code internally:

```dart
if (!identical(_meta, other._meta)) {
  throw 'Merging messages with different builder infos';
}
```

To my surprise, this doesn't break anything.

So I will simplify the merge implementation in this PR and assume that we're always merging messages of same types. The code will be simpler, and with the same performance wins.

osa1:
I noticed that the methods for merging extension and non-extension fields are almost identical. I refactored the code a little bit to make this more clear. The diff is now just this:

```diff
@@ -1,4 +1,4 @@
-  void _mergeNonExtensionField(FieldInfo fi, dynamic fieldValue) {
+  void _mergeExtensionField(FieldInfo fi, fieldValue) {
     if (fi.isMapField) {
       final MapFieldInfo<dynamic, dynamic> mapInfo = fi as dynamic;
       final map = mapInfo._ensureMapField(_meta, this);
@@ -29,15 +29,17 @@
     }

     if (fi.isGroupOrMessage) {
-      final currentFieldValue = _values[fi.index!];
+      final currentFieldValue =
+          _ensureExtensions()._getFieldOrNull(fi as Extension<dynamic>);
+
+      GeneratedMessage msg = fieldValue;
       if (currentFieldValue == null) {
-        GeneratedMessage msg = fieldValue;
         fieldValue = msg.deepCopy();
       } else {
         final GeneratedMessage currentMsg = currentFieldValue;
-        fieldValue = currentMsg..mergeFromMessage(fieldValue);
+        fieldValue = currentMsg..mergeFromMessage(msg);
       }
     }

-    _setNonExtensionFieldUnchecked(_meta, fi, fieldValue);
+    _ensureExtensions()._setFieldAndInfo(fi as Extension<dynamic>, fieldValue);
   }
```

There are two differences:

- We use different getters to get the field values for message-fields.
- We use different setters to set the field values for non-repeated and non-map fields.

osa1:
Closing this PR as it's quite old and the code is changed quite a bit in the meantime.

`mergeFromMessage` should still check message types as it doesn't even work properly when the message types are different. I'm testing this change internally, I'll create a new PR.

PR #664, opened 2022-05-31
Update documentation according to Effective Dart documentation guide

osa1:
Effective Dart documentation section:
https://dart.dev/guides/language/effective-dart/documentation

This addresses reviews in cl/449413572 about the documentation in this library.

Summary of changes:

- Use noun phrases in getters, fields

- Replace `[:...:]` syntax with ``` `` ```

- Use `[...]` syntax for in-scope identifiers in comments and documentation

- Use "create" instead of "return" for methods that create a new instance
  (rather than returning an existing one)

- Give a single sentence summary in API docs, move the rest to new paragraph(s)

- When documentation exception behavior, start the sentence with "Throws" in a
  new paragraph. Similarly when documenting return values do the same, starting
  with "Returns".

- Some rewording

- Use `ArgumentError.value` constructor in some cases (instead of
  `ArgumentError`)

Note that I did not review every line of comment/doc in the library to update
them according to the Effective Dart guidelines. This only addresses reviews
that came up in cl/449413572.

---

@lrhn Could you review please? (GitHub doesn't let me add you as a reviewer)


protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -6,26 +6,38 @@ part of protobuf;
...
-  /// Maps a tag number ([FieldInfo.tagNumber]) to the corresponding `oneof`
-  /// index (if any).
+  /// Mapping from [FieldInfo.tagNumber]s to the corresponding `oneof` indices
+  /// (if any).
   final Map<int, int> oneofs = <int, int>{};
 
+  /// Whether the message has extension fields.
   bool hasExtensions = false;
+
+  /// Whether the message has required fields. Note that proto3 doesn't have
```
lrhn, line 37: Make a new paragraph after the first `.`.

sigurdm (approved):
Very nice

LGTM

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -122,8 +142,8 @@ class FieldInfo<T> {
     return () => defaultOrMaker;
   }
 
-  /// Returns `true` if this represents a dummy field standing in for a field
-  /// that has been removed by a program transformation.
+  /// `true` if this represents a dummy field standing in for a field that has
```
lrhn, line 145: Another boolean getter which can use "Whether" (also to avoid starting with a lower-case letter). 
Try to avoid the unqualified "this".

Maybe:
```dart
/// Whether this field is a dummy field.
///
/// A dummy field represents a field that has been removed by a program transformation.
```
(Anything else we can say about dummy fields, like "don't use them for anything"?)

This is a private field, so the documentation doesn't need to be up to public-API standards.

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -95,8 +95,8 @@ abstract class GeneratedMessage {
 
   bool hasRequiredFields() => info_.hasRequiredFields;
 
-  /// Returns [:true:] if all required fields in the message and all embedded
-  /// messages are set, false otherwise.
+  /// Returns `true` if all required fields in the message and all embedded
+  /// messages are set, `false` otherwise.
```
lrhn, line 99: Should really have been a getter.
I'd treat it as a property, since it's already named as one, which means also documenting it as a boolean property. 
So another "Whether".

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -352,12 +353,12 @@ abstract class GeneratedMessage {
   dynamic getDefaultForField(int tagNumber) =>
       _fieldSet._ensureInfo(tagNumber).readonlyDefault;
 
-  /// Returns [:true:] if a value of [extension] is present.
+  /// Returns `true` if a value of [extension] is present.
```
lrhn, line 356: Would again use `whether`. It's a parameterized property and is named as such.

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -352,12 +353,12 @@ abstract class GeneratedMessage {
...
   bool hasExtension(Extension extension) =>
       _fieldSet._hasExtensions &&
       _fieldSet._extensions!._getFieldOrNull(extension) != null;
 
-  /// Returns [:true:] if this message has a field associated with [tagNumber].
+  /// Returns `true` if this message has a field associated with [tagNumber].
```
lrhn, line 361: Ditto.

protoc_plugin/lib/src/protobuf_field.dart
```diff
@@ -62,9 +62,11 @@ class ProtobufField {
   bool get isRepeated =>
       descriptor.label == FieldDescriptorProto_Label.LABEL_REPEATED;
 
-  /// True if a numeric field is repeated and it should be encoded with packed
-  /// encoding. In proto3 repeated fields are encoded as packed by default.
-  /// proto2 requires `[packed=true]` option.
+  /// Whether a numeric field is repeated and it must be encoded with packed
```
lrhn, line 65: I'd remove the "it".

protoc_plugin/lib/src/protobuf_field.dart
```diff
@@ -62,9 +62,11 @@ class ProtobufField {
...
+  /// encoding.
+  ///
+  /// In proto3 repeated fields are encoded as packed by default. proto2
```
lrhn, line 68: Avoid starting a sentence with a lower-case letter.
Either "Proto2 requires" or "The proto2 format requires" .... (or whichever noun is appropriate, if "format" isn't).

protoc_plugin/test/json_test.dart
```diff
@@ -29,8 +29,8 @@ void main() {
       '"70":"410","71":411,"72":412,"73":false,"74":"415","75":"NDE2",'
       '"81":1,"82":4,"83":7,"84":"424","85":"425"}';
 
-  // Checks that message once serialized to JSON
-  // matches TEST_ALL_TYPES_JSON massaged with [:.replaceAll(from, to):].
+  /// Checks that message once serialized to JSON matches [testAllJsonTypes]
```
lrhn, line 32: I'd use parenthetical comments: ", once serialized to JSON,"

lrhn (approved):
Very nice, few nits.

PR #665, opened 2022-06-01
Refactor protobuf bench runners, update README:

osa1:
- Use the same set of files in all versions of the benchmark runners.

  Previously the native version was searching for dataset files, JS
  version had hard-coded file paths. It's unclear whether these both run
  the same set of files, so we now have dataset file paths in the common
  library and use the same file list in all versions.

- Update link to protoc_plugin in README

- Update commands in the README to allow running them in `benchmarks/`
  directory.

  Since the README is in `benchmarks/` it makes sense for it to show
  commands to be run in the same directory.

- Add a note in README about the "Error reading file" errors when
  running the JS benchmarks. d8's `readbuffer` doesn't show the path of
  the file it's trying to read, so clarify that in README.

### DIFF
--- a/protobuf/benchmarks/README.md
+++ b/protobuf/benchmarks/README.md
@@ -9,12 +9,13 @@ JSON formats, and hash code generation for messages.
 ## Prerequisites
 
 Before running benchmarks you first need to compile all involved protos, which
-requires installing `protoc` and [dart-protoc-plugin](../../protoc_plugin).
+requires installing `protoc` and [protoc_plugin](../../protoc_plugin).
 
 ```console
 $ ./compile-protos.sh
-# This would produce temp folder with output of protoc.
 ```
+
+This produces temp folder with output of protoc.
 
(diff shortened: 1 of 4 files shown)


protobuf/benchmarks/README.md
```diff
@@ -8,25 +8,32 @@ JSON formats, and hash code generation for messages.
...
+Before running benchmarks you first need to compile all involved protos, which
+requires installing `protoc` and [protoc_plugin](../../protoc_plugin).
 
 ```console
-$ cd benchmarks
 $ ./compile-protos.sh
-# This would produce benchmarks/temp folder with output of protoc.
 ```
 
+This produces temp folder with output of protoc.
```
sigurdm, line 18: Actually that seems kind of non-standard. Should we not generate into `lib/` or something?

(Doesn't have to be this PR)

sigurdm (approved):
LGTM

PR #666, opened 2022-06-01
Remove makeLongInt helper

osa1:
This function no longer used by the generated code, uses were removed in
fec382b4.


sigurdm (approved):
Does this require an update to the protoc_plugin pubspec.yaml constraint on protobuf?

LGTM

osa1:
> Does this require an update to the protoc_plugin pubspec.yaml constraint on protobuf?

I don't think so, removing an unused definition shouldn't cause breakage.

PR #667, opened 2022-06-01
Benchmarks: add proto3 ser/deser benchs, move common code to common.dart

osa1:
- Add benchmarks for proto3 JSON string and object generation and parsing

- Use for-in loops instead of indices in benchmark code

- Move benchmark entry point to `common.dart`, reuse in JS and VM benchmark
  programs


sigurdm (approved):
LGTM

PR #670, opened 2022-06-02
[WIP] Avoid intermediate values when serializing proto3 JSON

osa1:
This uses jsontool library to avoid allocating intermediate values when
generating proto3 JSON strings.

Note that the code is not tested for correctness yet, and well-known
types are not supported.

AOT results:

Before:

    ToProto3JsonString(RunTime): 40474.54 us.
    ToProto3JsonString(RunTime): 40340.84 us.
    ToProto3JsonString(RunTime): 40403.9 us.

After:

    ToProto3JsonString(RunTime): 26978.906666666666 us.
    ToProto3JsonString(RunTime): 26891.893333333333 us.
    ToProto3JsonString(RunTime): 26945.04 us.

Diff: -33%

JS results:

Before:

    ToProto3JsonString(RunTime): 71551.72413793103 us.
    ToProto3JsonString(RunTime): 72103.44827586207 us.
    ToProto3JsonString(RunTime): 72035.71428571429 us.

After:

    ToProto3JsonString(RunTime): 45422.22222222222 us.
    ToProto3JsonString(RunTime): 45266.666666666664 us.
    ToProto3JsonString(RunTime): 46813.95348837209 us.

Diff: -36%

---

cc @rakudrama

### DIFF
--- a/protobuf/lib/src/protobuf/generated_message.dart
+++ b/protobuf/lib/src/protobuf/generated_message.dart
@@ -223,6 +223,13 @@ abstract class GeneratedMessage {
           {TypeRegistry typeRegistry = const TypeRegistry.empty()}) =>
       _writeToProto3Json(_fieldSet, typeRegistry);
 
+  String toProto3JsonString(
+      {TypeRegistry typeRegistry = const TypeRegistry.empty()}) {
+    StringBuffer buf = StringBuffer();
+    _writeToProto3JsonSink(_fieldSet, typeRegistry, buf);
+    return buf.toString();
+  }
+
   /// Merges field values from [json], a JSON object using proto3 encoding.
   ///
   /// Well-known types and their special JSON encoding are supported.
--- a/protobuf/lib/src/protobuf/proto3_json_string.dart
+++ b/protobuf/lib/src/protobuf/proto3_json_string.dart
@@ -1,0 +1,141 @@
+part of protobuf;
+
+void _writeToProto3JsonSink<S extends StringSink>(
+    _FieldSet fs, TypeRegistry typeRegistry, S sink) {
+
+  // TODO: Either add `writeToProto3JsonSink` for well-known types, or fall
+  // back to slow proto3 JSON serialization
+  if (fs._meta.toProto3Json != null) {
+    throw 'well-known type detected';
+  }
+
+  final JsonWriter<String> jsonWriter = jsonStringWriter(sink, indent: null);
+  _writeToProto3JsonWriter(fs, typeRegistry, jsonWriter);
+}
+
+void _writeToProto3JsonWriter(
+    _FieldSet fs, TypeRegistry typeRegistry, JsonWriter<String> jsonWriter) {
+  jsonWriter.startObject(); // start message
+
+  for (FieldInfo fieldInfo in fs._infosSortedByTag) {
+    var value = fs._values[fieldInfo.index!];
+
+    if (value == null || (value is List && value.isEmpty)) {
+      continue; // It's missing, repeated, or an empty byte array.
+    }
+
+    jsonWriter.addKey(fieldInfo.name);
+
+    if (fieldInfo.isMapField) {
+      jsonWriter.startObject(); // start map field
+      final MapFieldInfo mapEntryInfo = fieldInfo as MapFieldInfo;
+      for (MapEntry entry in (value as PbMap).entries) {
+        final key = entry.key;
+        final value = entry.value;
+        _writeMapKey(key, mapEntryInfo.keyFieldType, jsonWriter);
+        _writeFieldValue(
+            value, mapEntryInfo.valueFieldType, jsonWriter, typeRegistry);
+      }
+      jsonWriter.endObject(); // end map field
+    } else if (fieldInfo.isRepeated) {
+      jsonWriter.startArray(); // start repeated field
+      for (final element in value as PbListBase) {
+        _writeFieldValue(element, fieldInfo.type, jsonWriter, typeRegistry);
+      }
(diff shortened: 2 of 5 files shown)


protobuf/lib/src/protobuf/proto3_json_string.dart
```diff
@@ -0,0 +1,140 @@
+part of protobuf;
+
+void _writeToProto3JsonSink<S extends StringSink>(
+    _FieldSet fs, TypeRegistry typeRegistry, S sink) {
+  // TODO: Either add `writeToProto3JsonSink` for well-known types, or fall
```
sigurdm, line 5: An in-between would be to use the current `toProto3Json` helper, walk the ouput that and send it to the Sink...

protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -223,6 +223,13 @@ abstract class GeneratedMessage {
           {TypeRegistry typeRegistry = const TypeRegistry.empty()}) =>
       _writeToProto3Json(_fieldSet, typeRegistry);
 
+  String toProto3JsonString(
```
sigurdm, line 226: Consider passing an "indent" value

sigurdm (approved):

osa1:
I wonder how to best test these changes. I see a few options:

## 1. Update existing tests

- For each test that checks `toProto3Json` output:
  ```dart
  expect(msg.toProto3Json(), expectedJson);
  ```
  Add a new check:
  ```dart
  expect(jsonDecode(msg.toProto3JsonString()), expectedJson);
  ```

- For each test that checks `mergeFromProto3Json` output:
  ```dart
  expect(msg.mergeFromProto3Json(json), expectedMsg);
  ```
  Add a new check:
  ```dart
  expect(msg.mergeFromProto3JsonString(jsonEncode(json)), expectedMsg);
  ```

I tried to implement this. It's a lot of work as we have lots of tests, but it should work fine.

One problem that I realized is that for floats and doubles that are integers we avoid the fraction part while `jsonEncode` generates it. This is explicit in the current JSON serializer (not proto3 JSON, the custom one): https://github.com/google/protobuf.dart/blob/a549b57a007db71b3de7ef25d54ce3afcf0dbcc9/protobuf/lib/src/protobuf/json.dart#L33-L35

In the proto3 JSON serializer added in this PR it's implicitly done by `JsonWriter.addNumber` implementation.

Example from an actual test:

```dart
final json = {'optionalDouble': 5};
final encoded = jsonEncode(json);
TestAllTypes proto = TestAllTypes()..optionalDouble = 5.0;
expect(TestAllTypes()..mergeFromProto3Json(json), proto);
expect(TestAllTypes()..mergeFromProto3JsonString(encoded), proto);
expect(proto.toProto3Json(), json);
expect(proto.toProto3JsonString(), encoded);
```

The last line fails: string generated by `toProto3JsonString` is `'5'`, string generated by `dart:convert`'s `jsonEncode` is `'5.0'`.

From correctness point of view I think this should not matter as JSON has one number type and `'5'` should be the same as `'5.0'`. It just means we will need to tweak the tests a little bit.

## 2. Assume that `toProto3Json` and `mergeFromProto3Json` are correct, check that `toProto3JsonString` and `mergeFromProto3JsonString` agree with them

We don't modify the existing tests. Add new tests that, for a proto message `msg` and its encoding `encoded = jsonEncode(msg.toProto3Json())`:

- `encoded` is equal to `msg.toProto3JsonString()`
- `Msg.createEmptyInstance()..mergeFromProto3JsonString(encoded)` is equal to `msg`

We need to find some proto messages for these tests. Some ideas:

- Parse [`testAllTypesJson`](https://github.com/google/protobuf.dart/blob/a549b57a007db71b3de7ef25d54ce3afcf0dbcc9/protoc_plugin/test/proto3_json_test.dart#L28) to a `TestAllTypes` message and use it.

- Use the datasets in `protobuf/benchmarks/datasets`
  - Unclear whether this will have good test coverage. These proto definitions and data sets are generated from real-world uses, which may not 

- Generate random messages from `BuilderInfo`s
  - A lot of work
  - Difficult to make sure we won't repeat the same tests, cover edge cases

protobuf/lib/src/protobuf/proto3_json_string.dart
osa1: Yeah I think this makes sense, I just implemented this. Performance should be the same as before.

protobuf/lib/src/protobuf/generated_message.dart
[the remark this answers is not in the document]
osa1: Additions in this PR will be inconsistent unless we want to do breaking changes.

Currently in master branch (and also in the latest release) we have these generators:

- writeToBuffer
- writeToCodedBufferWriter
- writeToJson
- writeToJsonMap
- toProto3Json

So the proto3 method is already inconsistent.

Without changing it, we can name the new methods `toProto3...`, or `writeToProto3...`. Either way it will be inconsistent.

Should we rename `toProto3Json`? @sigurdm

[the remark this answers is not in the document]
sigurdm: Argh - I see - I introduced that inconsistency - not sure what is the better path forwards.
I guess if we want to be consistent we do:
```
@Deprecated('Use `writeToProto3JsonMap`');
toProto3Json() => writeToProto3JsonMap();

writeToProto3JsonMap();
writeToProto3Json();
```

If we prefer to avoid churn over consistency we keep it as `toProto3JsonString()`....

I guess it's up to your judgement.

osa1:
I'm splitting this into two PRs, for generating JSON and parsing JSON. #683 updates JSON generators.

PR #671, opened 2022-06-03
Avoid cloning the JSON when parsing Any messages from proto3 JSON

osa1:


protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -113,10 +113,14 @@ abstract class AnyMixin implements GeneratedMessage {
             json);
       }
 
-      Object? subJson = info.fromProto3Json == null
-          // TODO(sigurdm): avoid cloning [object] here.
-          ? (Map<String, dynamic>.from(object)..remove('@type'))
-          : object['value'];
+      Object? subJson;
+      if (info.fromProto3Json == null) {
+        json.remove('@type');
```
sigurdm, line 118: We cannot do this if the incoming json-object is immutable eg. https://api.dart.dev/stable/2.17.3/dart-collection/UnmodifiableMapView-class.html

I think my original intention with the TODO was to somehow let the recursive call to "mergeFromProto3Json" somehow ignore the "@type" key....

osa1: Good point.. Closing.

PR #675, opened 2022-06-13
Update GeneratedMessage== to accept Object instead of dynamic

jakemac53:
This matches the overridden `==` method, previously it was expanding the parameter type unnecessarily.

This also fixes Mockito codegen for fakes, which adds back an `==` method with type `Object`, which isn't allowed if the type being faked has widened it to `dynamic`.

Note that you can still do `== null`, that is special cased.

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,8 +1,73 @@
 ## 2.0.2-dev
 
-* Update `==` on `GeneratedMessage` to have the parameter type `Object`.
-  * Fixes mockitos ability to generate `Fake` protobuf instances.
-  * Matches the overridden `==` method from `Object`.
+* Update READMEs of `protobuf` and `protoc_plugin`:
+  * Use `dart pub` instead of `pub` in command examples ([a7e75cb])
+  * Fix typos, clarify installation instructions, mention native compilation,
+    fix proto syntax for `protoc_plugin` ([#610], [#617], [#641])
+* Update some of the documentation according to Effective Dart documentation
+  guide ([#664], [#674])
+* Improve runtime perf by removing some of the runtime type checks ([#574],
+  [#573])
+* Fix a bug when converting negative `Timestamp` to Dart `DateTime` ([#580],
+  [#577])
+* Document `BuilderInfo` and `FieldInfo` properties ([#597])
+* Improve `BuilderInfo` initialization by doing some of the work lazily
+  ([#606])
+* Improve enum hash code generation ([#556])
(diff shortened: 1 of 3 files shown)


fowles (approved):

osa1 (commented):
> which isn't allowed if the type being faked has widened it to dynamic.

We discussed this with @mraleph. The inferred type of the argument here is already `Object` as this method overrides a superclass method, and in the superclass the argument type is `Object`. Here's an example of this behavior:

```dart
class X {
  void f(String s) {}
}

Type getType<T>(T t) {
  return T;
}

class Y extends X {
  @override
  void f(s) {
    print(getType(s));
  }
}

void main() {
  Y().f('a');  // prints "String"
}
```

(I remember reading this behavior in the language spec also, but I can't find it now)

Here's another way to check this:

```dart
import 'package:protobuf/protobuf.dart';

class Test extends GeneratedMessage {
  @override
  GeneratedMessage clone() {
    throw '';
  }

  @override
  GeneratedMessage createEmptyInstance() {
    throw '';
  }

  @override
  BuilderInfo get info_ {
    throw '';
  }

  @override
  bool operator ==(Test test) {
    throw '';
  }
}
```

The error message shows that the inferred argument type in `GeneratedMessage.operator ==` is `Object`, not `dynamic`:

```
  error - protobuf_test.dart:20:17 - 'Test.==' ('bool Function(Test)') isn't a valid override of 'GeneratedMessage.==' ('bool Function(Object)'). -
          invalid_override
           - The member being overridden at protobuf_test.dart:27:3659.
```

So it seems to me that this reveals a bug in Mockito, or something that Mockito uses.

We can still merge this of course but we should update the changelog entry as it's wrong. Actually we should probably omit the changelog entry as this is not a user-visible change.

jakemac53:
Hmm interesting, I am not sure why this fixes the Mockito issue then. It doesn't hurt to land this either way, as it is just more explicit about the type. But I will leave that  up to you.

I will reproduce the error I was getting and see if I can understand more the actual problem.

srawlins:
I thought the same thing. Parameter types get inherited. So strange. Perhaps a bug in mockito.

osa1:
Just for the record, here's where the language spec describes inference of return and argument types of overriding methods: https://github.com/dart-lang/language/blob/ee1135e0c22391cee17bf3ee262d6a04582d25de/resources/type-system/inference.md?plain=1#L69-L72

> 1. **Method override inference**
>    * If you omit a return type or parameter type from an overridden or
>    implemented method, inference will try to fill in the missing type using the
>   signature of the methods you are overriding.

PR #678, opened 2022-06-15
Move protobuf benchmarks to top-level, make it a package

### LINKED ISSUE #106: Rename `benchmarks` directory to `benchmark`
Per https://www.dartlang.org/tools/pub/package-layout#tests-and-benchmarks

osa1:
"benchmarks" directory is causing this warning when publishing packages:

    Rename the top-level "benchmarks" directory to "benchmark".
    The Pub layout convention is to use singular directory names.
    Plural names won't be correctly identified by Pub and other tools.
    See https://dart.dev/tools/pub/package-layout.

Moving benchmarks to the top-level and making them a package also makes it
easier to build and run them in golem.

Closes #106


sigurdm (approved):
LGTM

osa1:
> Could also move them out of the package and to the top-level of the repo. Really no need to publish these.

Make sense. Do you have any naming suggestions? We have `query_benchmark`, `api_benchmark`. Not sure what to call these benchmarks.

osa1:
@sigurdm moving benchmarks to top-level will also simplify golem config. Do you have any naming suggestions for these benchamrks?

protobuf_benchmarks/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+dependencies:
+  protobuf: any
+
+dev_dependencies:
+  benchmark_harness: ^2.0.0
+  js: ^0.6.3
+  lints: ^1.0.0
+  protoc_plugin:
+    path: '../protoc_plugin'
```
mkustermann, line 20: Could you align this with `query_benchmark/pubspec.yaml` (it has e.g. a `dependency_override` for `package:protobuf`).

It may be sufficient to make `protobuf` also a `dev_dependency` with a `path: ` since we don't publish these benchmark packages.

protobuf_benchmarks/compile_protos.sh
```diff
@@ -32,4 +32,4 @@ mkdir -p ${OUTPUT_DIR}
 
 set -x
 
-protoc -I"${SCRIPT_DIR}" --dart_out="${OUTPUT_DIR}" "${PROTOS[@]/#/$SCRIPT_DIR/}"
+protoc -I"${SCRIPT_DIR}" --dart_out="${OUTPUT_DIR}" --plugin=protoc-gen-dart=run_protoc_plugin.sh "${PROTOS[@]/#/$SCRIPT_DIR/}"
```
mkustermann, line 35: Consider aligning with `query_benchmark` by generating the dart files into `bin` or `lib`.

mkustermann (approved):
Consider moving the `*.dart` files into `bin` / `lib` folders (similar to `query_benchmark`). Usually the actual programs (i.e. that contain `main`) are in `bin` and the rest can be either in `lib` or `bin`.

protobuf_benchmarks/pubspec.yaml
osa1: > It may be sufficient to make protobuf also a dev_dependency with a path: since we don't publish these benchmark packages.

That doesn't work:

```
Because protobuf_benchmarks depends on protoc_plugin from path which depends on protobuf ^2.0.0, protobuf from hosted is required.
So, because protobuf_benchmarks depends on protobuf from path, version solving failed.
```

osa1: I added a dep override as in query_benchmark.

protobuf_benchmarks/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
mkustermann, line 1: 2022 :)

query_benchmark/compile_protos.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
```
mkustermann, line 1: Why this change?

mkustermann (approved):

PR #679, opened 2022-06-16
Remove redundant reserved names

### LINKED ISSUE #637: Revert c35d78 before next major release
c35d78 was made to avoid breaking changes. Internally these changes do not break any code, but it's possible that it will break some code in the wild. Before the next major release we should revert c35d78 and add changelog a changelog entry describing the potential breakage.

osa1:
This reverts commit c35d787726173380348e744085d02d34d56d3ec1.

These reserved names are no longer reserved (see 146b186, b96dc21), but
we kept them to avoid breaking backwards compatibility.

Removing them now for the next major version bump.

Closes #637


sigurdm (approved):
LGTM

PR #681, opened 2022-06-16
Hide internal functions from Dart docs, add documentation to top-level types

osa1:
This hides items that are supposed to be used by the plugin-generated code and
mixin classes.

Items that can be made private are made private.

One sentence documentation added to top-level types that didn't have documentation.

This PR is best reviewed by looking at the documentation generated by `dart
doc`.


protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -310,6 +310,8 @@ abstract class GeneratedMessage {
   /// The tagNumber should be a valid tag or extension.
   void clearField(int tagNumber) => _fieldSet._clearField(tagNumber);
 
+  /// For generated code only.
+  /// @nodoc
   int $_whichOneof(int oneofIndex) => _fieldSet._oneofCases![oneofIndex] ?? 0;
```
sigurdm, line 315: Consider also marking these as @protected. They will give a warning if used in the wrong context, and if https://github.com/dart-lang/sdk/issues/49273 was fixed they would also be removed from tab-completion.

osa1: I need to make a pass to find out which ones need `@Protected`, I'd like to do that in a separate PR if that's OK.

sigurdm: That's fine!

sigurdm (approved):
LGTM

PR #684, opened 2022-06-21
Remove Dart-name-to-proto-name conversions in runtime

### LINKED ISSUE #599: protoc_plugin converts field names to camelCase, protobuf converts it back to snake case
protoc_plugin converts field names to camelCase: https://github.com/google/protobuf.dart/blob/9da84aef9d8126ac3da665f3c163e3cb4af31b6d/protoc_plugin/lib/names.dart#L426-L428 (in `_fieldMethodSuffix`, which, confusingly, is not really returning a suffix. It's also used as the field name as full).

The library then converts those names back to snake_case: https://github.com/google/protobuf.dart/blob/9da84aef9d8126ac3da665f3c163e3cb4af31b6d/protobuf/lib/src/protobuf/field_info.dart#L60

The problem is conversion from snake_case to camelCase is not invertible. For example, all of these snake_case identifiers are converted to the same camelCase:

- a_b_c_d
- a_b_cD
- a_bCD

So when we construct a `FieldInfo` without specifying the `proto_name` argument (it's optional) we'll get wrong `protoName` for the field in this constructor: https://github.com/google/protobuf.dart/blob/9da84aef9d8126ac3da665f3c163e3cb4af31b6d/protobuf/lib/src/protobuf/field_info.dart#L51-L65

Looking at protoc_compiler, I think we always pass a `proto_name` argument. If that's true then we should make this argument non-optional as it's otherwise a potential footgun.

osa1:
Currently plugin generates Dart names from proto names by converting snake_case
to camelCase.

We then convert those camelCase Dart names to snake_case.

This has correctness problems as described in #599.

Another problem is conversion of camelCase Dart names to snake_case is done
using regex, which is slow in AOT:
https://github.com/dart-lang/sdk/issues/37785

This PR removes `_unCamelCase` by passing proto names of fields, in addition to
Dart names.

TODO: Benchmark binary sizes

Fixes #599.


osa1:
We could also just not use regex in `_unCamelCase`.

osa1:
This changed causes 2.8% binary size increase in a 15M app.

I can reduce this to 2.6% by only generating `protoName` arguments when it's different than the `name`.

Adding `omit_field_names` guard does not make any difference as that flag is not used in this app.

So it seems to be like we should be optimizing `_unCamelCase` instead.

PR #685, opened 2022-06-21
Eliminate dynamic calls

### LINKED ISSUE #578: eliminate dynamic calls in package:protobuf

osa1:
This removes most of the dynamic calls from the library.

Dynamic calls are found by temporarily enabling the `avoid_dynamic_calls` lint.

After this PR 4 more dynamic calls will be left. Two of them are easily removed
by refactoring the tests. I will do that in a separate PR.

# Benchmarks

https://golem.corp.goog/Revision?repository=protobuf&revision=84854&patch=17206

Runtime performance results are mixed, but note that benchmark programs only
have a few `GeneratedMessage`, `PbEnum` etc. subclasses, and number of
subclasses affect dynamic call sites (cache sizes). With downcasts the runtime
should be more predictable, as with AOT (and probably in JS too, but not in JIT
which cannot assume closed world) downcasts to non-generic types are very fast.

## AOT

protobuf_from_binary (Intel Xeon) 1.277% (-0.2 noise)
protobuf_to_json_string (Intel Xeon) 2.181%
protobuf_from_proto3_json_object (Intel Xeon) 1.255%
protobuf_to_proto3_json_string (Intel Xeon) 1.357%
protobuf_query_set_nested_value (Intel Xeon) 1.934%
protobuf_query_encode_json (Intel Xeon) -0.9426%
protobuf_query_decode_json (Intel Xeon) -0.9199%
protobuf_query_encode_binary (Intel Xeon) -1.670%
protobuf_query_decode_binary (Intel Xeon) -0.5869%
protobuf_from_json_string (Intel Xeon) -1.031%
protobuf_from_proto3_json_string (Intel Xeon) -1.374%
protobuf_to_binary (Intel Xeon) -2.067%
protobuf_hash_code (Intel Xeon) -1.778%
protobuf_to_proto3_json_object (Intel Xeon) -1.243%

## JS

protobuf_to_binary (Intel Xeon) 5.187%
protobuf_from_binary (Intel Xeon) 1.970%
protobuf_to_proto3_json_object (Intel Xeon) 1.618%
protobuf_query_decode_binary (Intel Xeon) 0.6887%
protobuf_from_proto3_json_string (Intel Xeon) 0.3978%
protobuf_to_json_string (Intel Xeon) -0.0496%
protobuf_to_proto3_json_string (Intel Xeon) -0.1995%
protobuf_from_json_string (Intel Xeon) -0.4938%
protobuf_hash_code (Intel Xeon) -1.154%
protobuf_query_decode_json (Intel Xeon) -1.146%
protobuf_from_proto3_json_object (Intel Xeon) -0.3260%
protobuf_query_set_nested_value (Intel Xeon) -0.9993%
protobuf_query_encode_json (Intel Xeon) -0.7820%
protobuf_query_encode_binary (Intel Xeon) -2.346%

# Code size

This reduces JS size of an internal app from 15,753,174 bytes to 15,750,850
bytes (-0.01%).

cl/461167514


osa1:
I benchmarked this PR on Golem: https://golem.corp.goog/Revision?repository=protobuf&revision=84839&patch=17189

The results are mixed. Some results that I found interesting:

- On AOT, `from_binary` regressed by 1.1%, but `query_decode_binary` improved by 0.6%. Both of these are the same benchmark (binary deserialization) on different input, so assuming the results are accurate and not noisy, it seems like this PR makes deserialization of some fields faster and some fields slower.

- On JS, `to_binary` regressed by 3.5%, but `query_encode_binary` improved by 2.8%. This case is the same as above: these are the same benchmarks with different inputs.

- Binary encoding got faster on AOT but slower on JS. Binary decoding got faster on JS but slower on AOT.

protobuf/lib/src/protobuf/extension_field_set.dart
```diff
@@ -177,7 +177,7 @@ class _ExtensionFieldSet {
     _isReadOnly = true;
     for (var field in _info.values) {
       if (field.isRepeated) {
-        final entries = _values[field.tagNumber];
+        PbList? entries = _values[field.tagNumber];
```
osa1, line 180: Should we move this type cast after `entries == null` check below? Assuming null checks are the same for every type, this would avoid casting in AOT and JIT when the value is `null`.

protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -134,7 +134,7 @@ T _reparseMessage<T extends GeneratedMessage>(
         resultMap ??= ensureResult()._fieldSet._values[field.index!];
 
     if (field.isRepeated) {
-      final messageEntries = message._fieldSet._values[field.index!];
+      PbList? messageEntries = message._fieldSet._values[field.index!];
```
osa1, line 137: Same here -- it seems like we could avoid type casts in AOT and JIT when the value is `null`.

protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -146,7 +146,7 @@ T _reparseMessage<T extends GeneratedMessage>(
         }
       }
     } else if (field is MapFieldInfo) {
-      final messageMap = message._fieldSet._values[field.index!];
+      PbMap? messageMap = message._fieldSet._values[field.index!];
```
osa1, line 149: Same here.

protobuf/lib/src/protobuf/json.dart
```diff
@@ -38,23 +39,26 @@ Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
         // Encode 'bytes' as a base64-encoded string.
         return base64Encode(fieldValue as List<int>);
       case PbFieldType._ENUM_BIT:
-        return fieldValue.value; // assume |value| < 2^52
+        ProtobufEnum enum_ = fieldValue;
+        return enum_.value; // assume |value| < 2^52
```
osa1, line 43: @mkustermann we use the `dynamic` value (`fieldValue`) for one field access here. Do you think think the cast is still beneficial? I wonder if not casting here will be faster.

We have similar cases above.

protobuf/lib/src/protobuf/extension_field_set.dart
mkustermann: Good idea.

Yes, a `as PbList?` is more expensive than ` == null`.

protobuf/lib/src/protobuf/json.dart
mkustermann: Here are the trade-offs for VM's AOT:

When using `dynamic` for `fieldValue.value`:
  * startup: we have to pay per-call site per-runtime-class overhead (go to C++, lookup function in dictionary, ...)
  * code size: all classes that have `value` getter/field will retain it (cannot tree shake) and possibly get an extra `dyn:get:value` function injected (that boxes the unboxed `int64` into a tagged pointer)
  * if many different receiver classes flow into the call site we will have to look up the target in a per-callsite hashmap (megamorphic cache)

When using `as ProtobufEnum`:
  * we perform a class-id range check - which is - especially in AOT - very cheap
  * we do a vtable like call and may get a unboxed `int64` (and therefore avoid boxing)

protobuf/lib/src/protobuf/extension_field_set.dart
osa1: My understanding from our discussion on this:

- When we cast, then check for null:
  1. Check for cast: `valueClassId == NullClassId || valueClassId == PbListClassId`
  2. Then check if the value is null
 
- When we check for null, then cast:
  1. Check if the value is null (same check as the one above)
  2. Check for cast: `valueClassId == PbListClassId`

It shouldn't make a huge difference, but it seems like checking for null before casting should need a few instructions less.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -100,8 +100,9 @@ class CodedBufferWriter {
     final wireFormat = _wireTypes[_valueTypeIndex(valueType)];
 
     if ((fieldType & PbFieldType._REPEATED_BIT) != 0) {
-      for (var i = 0; i < fieldValue.length; i++) {
-        _writeValue(fieldNumber, valueType, fieldValue[i], wireFormat);
+      List list = fieldValue;
```
mkustermann, line 103: (If we know that those lists are actually a specific list (e.g. `PbList`) it may help to strengthen the type further)

protobuf/lib/src/protobuf/extension_field_set.dart
```diff
@@ -177,8 +177,9 @@ class _ExtensionFieldSet {
     _isReadOnly = true;
     for (var field in _info.values) {
       if (field.isRepeated) {
-        final entries = _values[field.tagNumber];
-        if (entries == null) continue;
+        var entriesDynamic = _values[field.tagNumber];
+        if (entriesDynamic == null) continue;
+        PbList entries = entriesDynamic;
```
mkustermann, line 182: You may consider using `final` for variables that aren't re-assigned.

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -792,14 +799,15 @@ class _FieldSet {
     }
 
     if (otherFi.isGroupOrMessage) {
-      final currentFi = isExtension!
+      GeneratedMessage? currentFi = isExtension!
```
mkustermann, line 802: You could consider doing the `as GeneratedMessage?` check only in the non-null case.

mkustermann (approved):

protobuf/lib/src/protobuf/coded_buffer_writer.dart
osa1: This value is `List` in some cases: https://github.com/google/protobuf.dart/blob/29f17490bcfe5e19d77482544c82a559df88d9b9/protobuf/lib/src/protobuf/unknown_field_set.dart#L207

We could probably use `PbList` in that code. I'll do that in another PR.

PR #694, opened 2022-06-27
Make protobuf_benchmarks/ have a single benchmarking program

mkustermann:
This makes the protobuf_benchmarks/ use a single benchmarking program -
shared between VM and Dart2js.

It uses conditional imports to decide how to read file contents (which
is different from VM and Dart2js).

This is already done in query_benchmarks, though using the deprecated
`dart:js` method.

=> We align this between query_benchmarks and protobuf_benchmarks.
=> We use the `package:js/js.dart` approach (since `dart:js` is
   deprecated)


protobuf_benchmarks/lib/readfile.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
mkustermann, line 1: It's 2018 - since it's a copy of the one from `query_benchmark`.

osa1 (approved):
LGTM. If you could fix the formatting I'll merge.

mkustermann:
@osa1 Should be fixed.

PR #695, opened 2022-06-27
proto3 JSON: Serialize integer floats and doubles as integers

osa1:
This is not required by the spec, but it seems to be more consistent
with how other libraries in other languages generate JSON.

Internally this change was made in cl/405998525.


sigurdm (approved):
LGTM

PR #552, opened 2021-11-12
Use `-O4` when compiling benchmarks.

LorenVS:
`--trust-type-annotations` is deprecated.

PR #698, opened 2022-06-28
Merge protobuf_benchmarks and query_benchmark

osa1:
Note: this is based on #697.

This merges two benchmark packages into one to make it easier to run benchmarks
on Golem and locally.

`api_benchmark` is not merged: it runs a web server and benchmarks performance
in an actual browser. We should revisit those benchmarks to see if they make
sense and see if it's a good idea to merge them into the new benchmark package
in this PR.

Changes:

- Remove `protobuf_benchmarks`, `query_benchmark`. Move benchmarks in both to
  `benchmarks`.

- Use `benchmark_harness` in all benchmarks. Previously only
  `protobuf_benchmarks` used `benchmark_harness`.

  TODO: Do we need to tweak number of iterations in `query_benchmark`
  benchmarks?

- Directory structure:
  - `datasets/` has benchmark inputs (i.e. binary encoded protos)
  - `protos/` has benchmark protos
  - `lib/` has code common to all benchmarks (file operations)
  - `compile_protos.sh` compiles all protos
  - `run_protoc_plugin.sh` runs protoc plugin from `../protoc_plugin`. Used by
    `compile_protos.sh`.
  - `bin/` contains benchmarks

- Each benchmark now has its own executable. This is to allow benchmarking
  changes in isolation. For example, to avoid changing encoding benchmark
  results when decoding code was modified, because of global analyses or JIT.

  I think one executable per benchmark also makes it easier to run benchmarks
  on Golem. (TODO: confirm)

- `protobuf_benchmarks/` inputs no longer use [`benchmarks.proto`][1]. Instead
  the input files are now the `payload` fields of `BenchmarkDataset`s, and
  input file name specifies message type.

[1]: https://github.com/google/protobuf.dart/blob/15d9eed9f44dc3bab0738d33921d2608f3b158b5/protobuf_benchmarks/benchmarks.proto

---

cc @mkustermann


benchmarks/bin/from_binary.dart
```diff
@@ -0,0 +1,46 @@
...
+    _message2Input = Uint8List.fromList(message2Input);
+  }
+
+  @override
+  void run() {
+    p2.GoogleMessage1.fromBuffer(_message1Proto2Input);
+    p3.GoogleMessage1.fromBuffer(_message1Proto3Input);
+    GoogleMessage2.fromBuffer(_message2Input);
+  }
+}
```
sigurdm, line 35: Could you initialize these in the initializer list - then they would not have to be final?
```suggestion
class Benchmark extends BenchmarkBase {
  final Uint8List _message1Proto2Input;
  final Uint8List _message1Proto3Input;
  final Uint8List _message2Input;

  Benchmark(String name, List<int> message1Proto2Input,
      List<int> message1Proto3Input, List<int> message2Input)
      : super(name) :
    _message1Proto2Input = Uint8List.fromList(message1Proto2Input),
    _message1Proto3Input = Uint8List.fromList(message1Proto3Input),
    _message2Input = Uint8List.fromList(message2Input);

  @override
  void run() {
    p2.GoogleMessage1.fromBuffer(_message1Proto2Input);
    p3.GoogleMessage1.fromBuffer(_message1Proto3Input);
    GoogleMessage2.fromBuffer(_message2Input);
  }
}
```

benchmarks/compile_protos.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
sigurdm, line 1: Consider moving this file to `benchmarks/tool`

sigurdm (approved):
LGTM

mkustermann:
> TODO: Do we need to tweak number of iterations in query_benchmark
benchmarks?

:+1: Would be good numbers stay compatible so we keep old history.

mkustermann:
Another thing you may want to consider: Check the `pubspec.lock` file (generated by `pub get` from `pubspec.yaml`) into the repository. This will ensure that if a new minor version of a dependency (e.g. `package:fixnum`) is published, we won't get performance result changes at random commits - instead upgrades of 3rd party packages will be visible by the commit that updates the lock file.

benchmarks/bin/from_binary.dart
osa1: Removed all `late` vars.

benchmarks/compile_protos.sh
osa1: Done.

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
...
+
+processes = []
+
+def run_cmd(args):
+    print(' '.join(args))
+
+    # Ignore outputs, print error messages
+    processes.append(subprocess.Popen(args, stdout=subprocess.DEVNULL))
+
+for file in os.listdir('bin'):
```
sigurdm, line 34: Would be cool to do this relative to the script file location. 

```
import os

file_path = os.path.realpath(__file__)
```
From: https://stackoverflow.com/questions/595305/how-do-i-get-the-path-of-the-python-script-i-am-running-in

That way you don't have to `cd benchmarks` before compiling but can do `benchmarks/tools/compile_benchmarks.py`.

benchmarks/tool/compile_benchmarks.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
sigurdm, line 1: I think we should remove this if we go with the python script.

benchmarks/lib/query_benchmark_common.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:benchmark_harness/benchmark_harness.dart';
+
+/// `benchmark_harness-2` reports 10x the time the benchmarked code takes. This
+/// overrides the `measure` method to divide the result by 10.
```
sigurdm, line 8: Perhaps leave a link to the issue explaining why you are doing this.

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
sigurdm, line 3: 2022

sigurdm (approved):
LGTM

sigurdm (approved):
LGTM

benchmarks/tool/compile_benchmarks.py
osa1: All of the files in this PR are new, should I update copyright years in all of the files?

benchmarks/lib/query_benchmark_common.dart
osa1: > Perhaps leave a link to the issue explaining why you are doing this.

Done.

I wonder if we should use this class in all benchmarks, because sooner or later https://github.com/dart-lang/benchmark_harness/issues/30 will be fixed (I hope?), and when that's done we will have to update the other benchmarks (those that don't use this class) to report 10x numbers, to be compatible with old results stored in Golem.

I will update the PR again to do this.

benchmarks/tool/compile_benchmarks.sh
osa1: Done.

benchmarks/tool/compile_benchmarks.py
osa1: `run_protoc_plugin.sh` and `compile_protos.sh` only work in `benchmarks`. Do you want me to update those as well?

benchmarks/lib/query_benchmark_common.dart
osa1: Updated PR to report accurate numbers in all benchmarks by dividing the benchmark_harness results by 10 before reporting.

benchmarks/tool/compile_benchmarks.py
osa1: Updated copyright years in tool/.

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
...
+for file in os.listdir('bin'):
+    file = os.path.splitext(file)[0]
+
+    run_cmd(['dart', 'compile', 'exe', f'bin/{file}.dart', '-o', f'out/{file}.exe'])
+    run_cmd(['dart', 'compile', 'js', '-O4', f'bin/{file}.dart', '-o', f'out/{file}.js'])
+
+print('Waiting for processes to finish...')
+
+for p in processes:
+    p.wait()
```
mkustermann, line 43: You should check for 0 return code and throw / exit non-0 otherwise.

The invoker is assuming that if this script completes with exit code 0 it was successful.

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
...
+
+for file in os.listdir('bin'):
+    file = os.path.splitext(file)[0]
+
+    run_cmd(['dart', 'compile', 'exe', f'bin/{file}.dart', '-o', f'out/{file}.exe'])
+    run_cmd(['dart', 'compile', 'js', '-O4', f'bin/{file}.dart', '-o', f'out/{file}.js'])
+
+print('Waiting for processes to finish...')
+
+for p in processes:
```
mkustermann, line 42: Since python allows importing this file (known as module), it is common practice to check whether this is actually the main entrypoint:
```python
if __name__ == '__main__':
  ... your code if the script is invoked (and not imported) ...
```

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
...
+    print(' '.join(args))
+
+    # Ignore outputs, print error messages
+    processes.append(subprocess.Popen(args, stdout=subprocess.DEVNULL))
+
+for file in os.listdir('bin'):
+    file = os.path.splitext(file)[0]
+
+    run_cmd(['dart', 'compile', 'exe', f'bin/{file}.dart', '-o', f'out/{file}.exe'])
+    run_cmd(['dart', 'compile', 'js', '-O4', f'bin/{file}.dart', '-o', f'out/{file}.js'])
```
mkustermann, line 38: If the number of benchmarks is large but the device we're running this on is limited (as often the case for golem, e.g. only 4 cores), this can spawn significantly more subprocesses in parallel than there are cores.

This will cause oversubscription of the CPU and may also cause out-of-memory situations.

I'd suggest to use a higher level abstraction that runs things in parallel with limited parallelism. For inspiration see e.g. [this comment](https://stackoverflow.com/questions/16450788/python-running-subprocess-in-parallel/16492930#16492930) and [cpu_count](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count)

benchmarks/tool/compile_benchmarks.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# $ time ./tool/compile_benchmarks.py
+# ...
+# ./tool/compile_benchmarks.py  130.47s user 19.48s system 1074% cpu 13.951 total
+#
+# $ time ./tool/compile_benchmarks.sh
+# ...
+# ./tool/compile_benchmarks.sh  91.14s user 20.08s system 137% cpu 1:20.97 total
+
+import os
```
mkustermann, line 18: I think it would be good that this script supports selecting the things to compile (golem e.g. doesn't want to compile everything always).

For example, support:
```
% tool/compile_benchmarks.py --target=jit,aot,js bin/
% tool/compile_benchmarks.py --target=aot bin/decode_json.dart
```
Thereby allowing to filter the target we're compiling for and the test (the former golem would use)


Furthermore for golem specifically it would be good if this script support "compiling" to JIT as well. This would significantly simplify the setup. So a
```
% tool/compile_benchmarks.py --target=jit bin/decode_json.dart
```

would use `dart --snapshot-kind=kernel --snapshot=out/decode_json.dart.dill bin/decode_json.dart` that can then be run via `dart out/decode_json.dart.dill` (just as one would from source via `dart bin/decode_json.dart`)

benchmarks/tool/compile_benchmarks.py
osa1: Why would you import this as a module? Do we want to support that use case?

benchmarks/tool/compile_benchmarks.py
osa1: Done.

benchmarks/tool/compile_benchmarks.py
osa1: Done.

benchmarks/tool/compile_benchmarks.py
osa1: No longer relevant -- I re-implemented the script in Dart.

osa1:
@mkustermann could you review `build_benchmarks.dart` please?

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,121 @@
...
+
+      default:
+        print(
+            'Unsupported target: $targetStr. Supported targets: jit, aot, js');
+        exit(1);
+    }
+  }
+
+  // Arg list is immutable, clone it
+  var sourceFiles = List.from(parsedArgs.rest);
```
mkustermann, line 44: nit: Prefer to use `final` for all variables that aren't reassigned (I believe there's analyzer lints you could enable for that)

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,121 @@
...
+  for (final sourceFile in sourceFiles) {
+    for (final target in targets) {
+      commands.add(target.compileArgs(sourceFile));
+    }
+  }
+
+  final pool = Pool(jobs);
+
+  final stream = pool.forEach(commands, (a) async {
+    var command = a as List<String>;
```
mkustermann, line 65: I think the code is typeable, something like this:

```
final commands = <List<String>>[];
...
final = pool.forEach<List<String>, ProcessResult>(commands, (List<String> command) {
  print(command.join(' '));
  return Process.run(...);
});
```
(the `async` isn't needed - since there's no await in here)

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,121 @@
...
+    print(command.join(' '));
+    return Process.run(command[0], command.sublist(1));
+  });
+
+  await for (final processResult in stream) {
+    final exitCode = processResult.exitCode;
+    if (exitCode != 0) {
+      print('Process exited with $exitCode');
```
mkustermann, line 73: Consider also printing `processResult.stdout` / `processResult.stderr`

osa1: Done.

benchmarks/tool/compile_benchmarks.dart
osa1: Done.

benchmarks/tool/compile_benchmarks.dart
osa1: Done.

osa1:
I'm updating Golem to use benchmarks added in this PR. Once that works I'll merge this PR.

benchmarks/lib/benchmark_base.dart
```diff
@@ -0,0 +1,19 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:benchmark_harness/benchmark_harness.dart' as bh;
+
+/// A subclass of [bh.BenchmarkBase] to work around
+/// https://github.com/dart-lang/benchmark_harness/issues/30.
+///
+/// Overrides the relevant [bh.BenchmarkBase] method(s) to report accurate
+/// results, rather than 10x the actual results.
+abstract class BenchmarkBase extends bh.BenchmarkBase {
```
sortie, line 12: Consider another name for this class to avoid confusion between the upstream BenchmarkBase and this modified one.

I am planning on fixing the 10x bug in BenchmarkBase once I have time to think up a migration scheme

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,184 @@
...
+
+  // Arg list is immutable, clone it
+  final sourceFiles = List.from(parsedArgs.rest);
+
+  if (sourceFiles.isEmpty) {
+    // Compile all files in bin/
+    final dir = Directory('bin');
+    for (final dirFile in dir.listSync(recursive: false)) {
+      // Compiler hangs (instead of failing) when passed an .exe, so only try
+      // to compile Dart files
```
sortie, line 60: Well yes, but logically it only makes sense to compile .dart files, it doesn't make sense for this program to compile random files regardless of whether the compiler hangs or not

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,184 @@
...
+      if (path.extension(dirFile.path) == '.dart') {
```
sortie, line 61: final sourceFiles = dir.listSync(recursive: false).where((entity) => entity.path.endsWith(".dart")).toList();

Or such :)

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,184 @@
...
+  final stream = pool.forEach<List<String>, ProcessResult>(commands,
+      (List<String> command) {
+    print(command.join(' '));
+    return Process.run(command[0], command.sublist(1));
+  });
+
+  await for (final processResult in stream) {
+    final exitCode = processResult.exitCode;
+    if (exitCode != 0) {
+      print('Process exited with $exitCode');
```
sortie, line 92: Output the failed command too, so it's known what command failed

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,184 @@
...
+  return ['dart', 'compile', 'js', sourceFile, '-o', 'out/$baseNameNoExt.js'];
+}
+
+List<String> jsProductionProcessArgs(String sourceFile) {
+  // This is used in golem "...-dart2js-production" benchmarks.
+  // `--benchmarking-production` is not documented, it enables
+  // `--omit-implicit-checks` and `--trust-primitives`:
+  // https://github.com/dart-lang/sdk/blob/c48f6fea580178bd34f2d872588dcc1c79bdb01c/pkg/compiler/lib/src/options.dart#L764-L767
+  //
+  // We will probably want to use `-O4` here.
```
sortie, line 172: --benchmarking-production is undocumented, but is meant to be whatever the web people currently consider the options one would use for production compilations of real world things (including slightly unsafe options which will change over time)

sortie (commented):
A couple thoughts :)

benchmarks/tool/compile_benchmarks.dart
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+  }
+
+  final commands = <List<String>>[];
+
+  if (sourceFiles.isNotEmpty && targets.isNotEmpty) {
+    try {
+      Directory('out').createSync();
+    } catch (_) {}
```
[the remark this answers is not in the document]
osa1: How do I check for that condition? `createSync` documentation just says "an exception is thrown".

benchmarks/tool/compile_benchmarks.dart
osa1: I wanted to use `--benchmarking-production` because that's what we use in golem and I wanted to be compatible with the existing results, but since then I updated my golem CL and removed code to run old revisions. I think we can use `-O4` now. (#552)

benchmarks/tool/compile_benchmarks.dart
osa1: Done. It's not any shorter when formatted though.

benchmarks/tool/compile_benchmarks.dart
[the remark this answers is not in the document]
mkustermann: Consider using `createSync(recursive: true)` - it will create any dirs not already there and not throw if they are there.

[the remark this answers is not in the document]
osa1: Done.

mkustermann (approved):

PR #700, opened 2022-07-07
Remove invalid Dart 2 tests

osa1:
`cast(x) as Y` tests runtime type checking, no need to test them in a
protobuf package.


sigurdm (approved):
r+

PR #703, opened 2022-07-08
Revert contructor arguments

osa1:
This PR reverts #442 (7ac17d0f)

This patch was ported to internal in cl/449707130, but benchmarking
revealed that the new constructors add to release binary sizes even when
no arguments are passed to the constructors.

To keep the two code bases in sync I'm reverting this change.

These new constructor arguments were never announced in the changelog,
but I still added an entry to show how to migrate code that uses them.


mraleph (approved):

agreaves:
I know this is closed, but I'd like to voice the opinion that this is an _incredibly_ harmful change for code quality. The constructors provide much needed structure and given Dart's formatting, the cascading `Foo()..bar = 4` approach is unwieldily and leads to far more unreadable code.

For this reason, we use the named constructor heavily in our codebase and this sole change is likely the reason we won't update the proto libraries any time soon. We'd have to undergo a painful migration just to have worse code quality in the end. In our case at least, the binary size increase is negligible in comparison. Realistically, though, the most likely scenario is that we fork the repo and revert this change (which will be no picnic either given future changes).

At the very least, I'd recommend making it optional and adding a flag for whether to add named constructors in the generated code (defaulting to `false`). If truly the only downsize is a larger binary I really think it's worth keeping at least as an option.

agreaves:
Actually, I'll just create an issue for this for visibility. Missed in all of the above is a big thank you for all the hard work on this library! It's actually extremely flexible and nice to use! Perhaps that's why I'm so passionate about my complaints 😅

iulian0512:
this is a major breaking change

LukaszRacon:
Issue created by @agreaves: https://github.com/google/protobuf.dart/issues/850

PR #707, opened 2022-07-11
Set map fields in `_getDefault`, minor refactoring

### LINKED ISSUE #373: Unable to initialise a null map field by field number
If a map field isn't set (e.g. it's null), then it doesn't seem to be possible to modify it by field number. It looks like `getField` returns the default value, and updates to that value are not reflected in the message... I've prepared a simplified example:

```proto
message Holder {
  map<string, int32> numbers = 1;
}
```

```dart
setNumber(Holder holder) {
  var map = holder.getField(1);
  map["a"] = 1;
}

void main() {
  
  // TEST FAILS:
  test("empty map", () {
    var holder = Holder();
    setNumber(holder);
    expect(holder.numbers["a"], 1);
  });

  // TEST PASSES:
  test("not empty map", () {
    var holder = Holder()..numbers["b"] = 2;
    setNumber(holder);
    expect(holder.numbers["a"], 1);
  });

}
```

Is there any way around this?

osa1:
- `$_getMap` when called on an uninitialized map field initializes the field.
  Update `getField` to do the same.

  `getField` already initializes repeated fields, but it used to not initialize
  map fields.

  Fixes #373

  This is a sync of cl/442686654.

- Inline `_getDefaultList` into `_$getList` and `_getDefaultMap` into
  `_$getMap`. These inlined functions have only one use site, and inlining them
  makes the `_$getList` and `_$getMap` easier to compare for consistency.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -200,13 +200,10 @@ class _FieldSet {
   }
 
   dynamic _getDefault(FieldInfo fi) {
-    if (!fi.isRepeated) return fi.makeDefault!();
+    if (!fi.isRepeated && !fi.isMapField) return fi.makeDefault!();
     if (_isReadOnly) return fi.readonlyDefault;
 
-    // TODO(skybrian) we could avoid this by generating another
-    // method for repeated fields:
-    //   msg.mutableFoo().add(123);
-    var value = fi._createRepeatedField(_message!);
+    var value = fi.makeDefault!();
     _setNonExtensionFieldUnchecked(_meta, fi, value);
     return value;
   }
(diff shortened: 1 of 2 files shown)


mraleph (approved):

osa1:
- `$_getMap` when called on an uninitialized map field initializes the field.
  Update `getField` to do the same.

  `getField` already initializes repeated fields, but it used to not initialize
  map fields.

  Fixes #373

  This is a sync of cl/442686654.

- Inline `_getDefaultList` into `_$getList` and `_getDefaultMap` into
  `_$getMap`. These inlined functions have only one use site, and inlining them
  makes the `_$getList` and `_$getMap` easier to compare for consistency.

PR #709, opened 2022-07-11
Make FieldSet lists use the same representation when empty and otherwise

osa1:
I think `List.filled` could return different `List`s when length is 0
and otherwise, but I checked SDK code and at least on JS that's not the
case.

With `[]` syntax the list will be modifiable, and representation can be
different than `List.filled(..., growable: false)`.

Updated `_zeroList` documentation for clarity.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -83,9 +83,9 @@ class _FieldSet {
     return List.filled(length, null, growable: false);
   }
 
-  // Use a fixed length list and not a constant list to ensure that _values
-  // always has the same implementation type.
-  static final List _zeroList = [];
+  // Use `List.filled` and not a `[]` to ensure that `_values` always has the
+  // same implementation type.
+  static final List _zeroList = List.filled(0, null, growable: false);
 
   // Metadata about multiple fields
 


mraleph (approved):
LGTM!

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -83,9 +83,9 @@ class _FieldSet {
     return List.filled(length, null, growable: false);
   }
 
-  // Use a fixed length list and not a constant list to ensure that _values
-  // always has the same implementation type.
-  static final List _zeroList = [];
+  // Use `List.filled` and not a `[]` to ensure that `_values` always has the
+  // same implementation type.
+  static final List _zeroList = List.filled(0, null, growable: false);
```
rakudrama, line 88: FYI

`List.empty()` is also fixed length.
`const []` would be fine here too, since it is also fixed length and AOT and dart2js both can understand mixed const and fixed-length lists.

mraleph: unless I am mistaken `const []` will be an instance of `_ImmutableArray` on the VM rather than an instance of a mutable array.

rakudrama: Yes, but `_List` and `_ImmutableList` now have a common supertype `_Array` which holds most of the methods, allowing e.g. `operator []` to be a single method for both.

There is nothing wrong with the merged code, as using `List.filled(..., growable: false)` in both places makes it obvious they are the same internal type.

I'm not sure it is worth special-casing the empty list. There are not many proto mesages with no non-extension fields.

PR #715, opened 2022-08-02
Fix presence of bytes fields

### LINKED ISSUE #690: Presence checks (hazzers) for `optional bytes` fields don't work
With `optional bytes optionalByts` and `optional string optionalStr` fields, if I run this:

```dart
  {
    Msg1 m1 = Msg1()..mergeFromProto3Json({'optionalByts': ''});
    Msg1 m2 = Msg1()..mergeFromProto3Json({});

    print('hash codes equal = ${m1.hashCode == m2.hashCode}');
    print('values equal = ${m1 == m2}');
    print('m1.optionalByts = ${m1.optionalByts}, m2.optionalByts = ${m2.optionalByts}, m1.hasOptionalByts = ${m1.hasOptionalByts()}, m2.hasOptionalByts = ${m2.hasOptionalByts()}');
  }

  print('---');

  {
    Msg1 m1 = Msg1()..mergeFromProto3Json({'optionalStr': ''});
    Msg1 m2 = Msg1()..mergeFromProto3Json({});

    print('hash codes equal = ${m1.hashCode == m2.hashCode}');
    print('values equal = ${m1 == m2}');
    print('m1.optionalStr = ${m1.optionalStr}, m2.optionalStr = ${m2.optionalStr}, m1.hasOptionalStr = ${m1.hasOptionalStr()}, m2.hasOptionalStr = ${m2.hasOptionalStr()}');
  }
```

This is the output:

```
hash codes equal = true
values equal = true
m1.optionalByts = [], m2.optionalByts = [], m1.hasOptionalByts = false, m2.hasOptionalByts = false
---
hash codes equal = false
values equal = false
m1.optionalStr = , m2.optionalStr = , m1.hasOptionalStr = true, m2.hasOptionalStr = false
```

So for an `optional string` field, if I see a message with the default value, that's different than if I don't see the field at all. This is working as expected.

But for `optional bytes`, seeing the default value in the serialized message is the same as not seeing it.

There are probably more bugs in proto3 optional implementation. I will read the specs ([1], [2]) carefully and implement some tests to see.

[1]: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md
[2]: https://github.com/protocolbuffers/protobuf/blob/main/docs/implementing_proto3_presence.md

osa1:
`_FieldSet._$has` previously returned `true` if a list field is empty.

Because `bytes` fields are represented as `List<int>`, this caused
incorrect presence when we set a bytes field to `[]`.

Remove the special case for lists in `_$has` to fix presence checks for
`bytes` fields.

The special case for lists in `_$has` is not necessary for repeated
fields as we don't generate hazzers for repeated fields.

Fixes #690

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -9,6 +9,7 @@
   consistent with `getField` called on repeated fields. ([#373], [#707])
 * Unused and optional `PbMap` constructor argument (`BuilderInfo? info`)
   removed.
+* Fix presence of `bytes` fields ([#690])
 
 [#183]: https://github.com/google/protobuf.dart/issues/183
 [#644]: https://github.com/google/protobuf.dart/pull/644
@@ -18,6 +19,7 @@
 [#626]: https://github.com/google/protobuf.dart/pull/626
 [#373]: https://github.com/google/protobuf.dart/issues/373
 [#707]: https://github.com/google/protobuf.dart/pull/707
+[#690]: https://github.com/google/protobuf.dart/issues/690
 
 ## 2.1.0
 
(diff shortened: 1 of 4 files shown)


sigurdm (approved):
LGTM

PR #718, opened 2022-08-02
Avoid redundant copying when deserializing binary messages

osa1:
When deserializing a field of type `message` we create a new empty
message, deserialize the binary encoding into the empty message, then
set the field.

When the parent message already has a value set for the message-field,
we copy the fields of the message-field to the empty message that we use
to deserialize the binary encoding.

This PR avoids this copying by decoding the binary encoding into the
value of the message-field directly.

Benchmark results don't seem to be affected: we probably don't have any
benchmarks that deserializes a field with message type multiple times in
a single binary encoding.


sigurdm (approved):
LGTM

PR #724, opened 2022-08-08
Add a `deepCopy` benchmark

osa1:
`deepCopy` is widely used to convert a frozen (immutable) message into a
mutable message and it's worth tracking the performance of it and
optimizing it.


sigurdm (approved):
LGTM

PR #729, opened 2022-08-10
Use Uint8List for bytes fields

### LINKED ISSUE #219: Generate `bytes` with `Uint8List`
Ref: https://groups.google.com/a/dartlang.org/forum/#!topic/misc/JpNkRRLI9_w

More and more packages are explicitly switching to Uint8List

osa1:
WIP -- submitting a PR to start testing and benchmarking with copybara.

Fixes #219


osa1:
I tested this internally. It was naive to hope that the breakage wouldn't be too bad..

I'll update #219 with some details.

PR #732, opened 2022-08-11
Eliminate more type casts when targeting JS

osa1:
Explicit type casts (`as`) syntax is expensive when targeting.

In #685 we replaced some of the explicit casts with implicit casts.

This PR does the same in a few more places.

AOT and JS binaries get a few bytes smaller and runtimes are almost the same. JS benchmarks get a few percent faster.

### DIFF
--- a/protobuf/lib/src/protobuf/extension_field_set.dart
+++ b/protobuf/lib/src/protobuf/extension_field_set.dart
@@ -39,26 +39,26 @@ class _ExtensionFieldSet {
   ///
   /// If it doesn't exist, creates the list and saves the extension.
   /// Suitable for public API and decoders.
-  List<T?> _ensureRepeatedField<T>(Extension<T?> fi) {
+  List<T> _ensureRepeatedField<T>(Extension<T> fi) {
     assert(!_isReadOnly);
     assert(fi.isRepeated);
     assert(fi.extendee == '' || fi.extendee == _parent._messageName);
 
     var list = _values[fi.tagNumber];
-    if (list != null) return list as List<T>;
+    if (list != null) return list;
 
-    return _addInfoAndCreateList(fi) as List<T?>;
+    return _addInfoAndCreateList(fi);
   }
 
-  List<T?> _getList<T>(Extension<T?> fi) {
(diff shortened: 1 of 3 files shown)


protobuf/lib/src/protobuf/extension_field_set.dart
```diff
@@ -39,26 +39,26 @@ class _ExtensionFieldSet {
   ///
   /// If it doesn't exist, creates the list and saves the extension.
   /// Suitable for public API and decoders.
-  List<T?> _ensureRepeatedField<T>(Extension<T?> fi) {
+  List<T> _ensureRepeatedField<T>(Extension<T> fi) {
```
rakudrama, line 42: An alternative approach is to add this annotation to the method:

    @pragma('dart2js:as:trust')

The annotation is desribed in this recently added file:
https://github.com/dart-lang/sdk/blob/main/pkg/compiler/doc/pragmas.md

osa1: When do I **not** want to add this annotation to my methods?

I wonder if it should be enabled by default (for the entire compilation unit) with `-O4`.

Currently my understanding is we always want to do this in the protobuf library, so enabling it with a flag for the entire compilation unit would be convenient as I wouldn't have to add JS-specific annotations everywhere.

rakudrama: There is no 'compilation unit' other than the _whole program_.

Making `as` be ignored when the user wrote it has always felt to me to be a step further than ignoring implicit checks like parametric covariant checks of generic class method arguments, and feels like a step to far. A global flag to ignore the user's expressed intent in the entire program is too big a hammer.

`@pragma('dart2js:late:trust')` is scoped so that it can be placed on the `library` tag to apply to the whole library.
Unfortunately `@pragma('dart2js:as:trust')` is not scoped so it must be put on each method containing `as`.
I would like to change this but it requires a refactoring that I can't consider working on until after we ship Dart 3.0.
(The annotation is consulted in a context where the structure of the program is not available).

sigurdm (approved):
LGTM

Issue #386, opened 2020-07-13
Possible to produce an appropriate error when decoding messages larger than 64 MiB?


tengyifei:
Hi, thanks for creating this project! It looks like https://github.com/dart-lang/protobuf/blob/cdffe5e612ef0ab7243858a9127e58edef719c98/protobuf/lib/src/protobuf/coded_buffer_reader.dart#L9 causes messages larger than 64 MiB to be silently truncated, resulting in a generic `InvalidProtocolBufferException`. May I ask the reason behind doing so? The 64 MiB magic number cannot be found from any public documentation; I think throwing an exception which explicitly indicates that the message is too big could make this easier to debug.

chgibb:
Why does this limit exist in the first place? It seems arbitrary.

PR #734, opened 2022-08-17
Avoid hitting size limit when parsing request in protoc plugin

mraleph:
By default `CodedBufferReader` only reads 64Mb of input, which can easily be exceeded for large protobuf definitions.

Additionally adjust reported error to make it easier to diagnose.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_reader.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_reader.dart
@@ -29,6 +29,14 @@ class CodedBufferReader {
     _currentLimit = _sizeLimit;
   }
 
+  void _reportTruncatedMessage(int limit) {
+    if (limit > _sizeLimit && limit <= _buffer.length) {
+      throw InvalidProtocolBufferException.truncatedMessageDueToSizeLimit(
+          _buffer.length, _sizeLimit);
+    }
+    throw InvalidProtocolBufferException.truncatedMessage();
+  }
+
   void checkLastTagWas(int value) {
     if (_lastTag != value) {
       throw InvalidProtocolBufferException.invalidEndTag();
@@ -46,7 +54,7 @@ class CodedBufferReader {
     byteLimit += _bufferPos;
     var oldLimit = _currentLimit;
     if ((oldLimit != -1 && byteLimit > oldLimit) || byteLimit > _sizeLimit) {
-      throw InvalidProtocolBufferException.truncatedMessage();
+      _reportTruncatedMessage(byteLimit);
     }
     _currentLimit = byteLimit;
     callback();
@@ -99,7 +107,7 @@ class CodedBufferReader {
     var oldLimit = _currentLimit;
     _currentLimit = _bufferPos + length;
     if (_currentLimit > oldLimit) {
-      throw InvalidProtocolBufferException.truncatedMessage();
+      _reportTruncatedMessage(_currentLimit);
     }
     ++_recursionDepth;
     message.mergeFromCodedBufferReader(this, extensionRegistry);
--- a/protobuf/lib/src/protobuf/exceptions.dart
+++ b/protobuf/lib/src/protobuf/exceptions.dart
@@ -6,9 +6,9 @@ part of protobuf;
 
 const _truncatedMessageText = '''
 While parsing a protocol message, the input ended unexpectedly
-in the middle of a field. This could mean either than the
-input has been truncated or that an embedded message
-misreported its own length.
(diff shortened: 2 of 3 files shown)


mraleph:
/reviewer @mkustermann for review as well (can't add as a reviewer through drop down).

(FYI, this is needed to unblock an internal customer)

protobuf/lib/src/protobuf/exceptions.dart
```diff
@@ -4,6 +4,13 @@
 
 part of protobuf;
 
+const _truncatedMessageText = '''
+While parsing a protocol message, the input ended unexpectedly
+in the middle of a field.  This could mean either than the
```
mit-mit, line 9: No double space between `.field` and `This`.

Last sentence should be "This could either mean that the input has been truncated or that an embedded message misreported its own length."

mit-mit (approved):
Just one grammar nit

mit-mit:
You missing the second part of my review comment: "either" goes right after "could", and it's "that" not "than"

mraleph:
> You missing the second part of my review comment: "either" goes right after "could", and it's "that" not "than"

Oops. Fixed that as well. The original message was just copy&paste so I did not read it :)

protobuf/lib/src/protobuf/coded_buffer_reader.dart
```diff
@@ -46,7 +54,7 @@ class CodedBufferReader {
     byteLimit += _bufferPos;
     var oldLimit = _currentLimit;
     if ((oldLimit != -1 && byteLimit > oldLimit) || byteLimit > _sizeLimit) {
-      throw InvalidProtocolBufferException.truncatedMessage();
+      _reportTruncatedMessage(byteLimit);
```
mkustermann, line 57: To make it more clear on call site, consider renaming to `_throwTruncatedMessageError` (`report` may mean it's just logging it somewhere)

protoc_plugin/lib/src/code_generator.dart
```diff
@@ -96,7 +96,14 @@ class CodeGenerator {
             BytesBuilder(), (BytesBuilder builder, data) => builder..add(data))
         .then((builder) => builder.takeBytes())
         .then((List<int> bytes) {
-      var request = CodeGeneratorRequest.fromBuffer(bytes, extensions);
```
mkustermann, line 99: This is normal protobuf API to decode messages. Do we want all users of this API to hit exceptions in production and then force them to rewrite their code like this?

It seems to me this 64 MB limit is completely arbitrary and seems to come from the very first version of this library in d6b19235.

I think it would make more sense to just remove this by instead changing it to:
```diff
diff --git a/protobuf/lib/src/protobuf/coded_buffer_reader.dart b/protobuf/lib/src/protobuf/coded_buffer_reader.dart
index 5333707..61bf401 100644
--- a/protobuf/lib/src/protobuf/coded_buffer_reader.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_reader.dart
@@ -9,8 +9,6 @@ part of protobuf;
 class CodedBufferReader {
   // ignore: constant_identifier_names
   static const int DEFAULT_RECURSION_LIMIT = 64;
-  // ignore: constant_identifier_names
-  static const int DEFAULT_SIZE_LIMIT = 64 << 20;
 
   final Uint8List _buffer;
   int _bufferPos = 0;
@@ -21,11 +19,12 @@ class CodedBufferReader {
   final int _sizeLimit;
 
   CodedBufferReader(List<int> buffer,
-      {int recursionLimit = DEFAULT_RECURSION_LIMIT,
-      int sizeLimit = DEFAULT_SIZE_LIMIT})
+      {int recursionLimit = DEFAULT_RECURSION_LIMIT, int sizeLimit = -1})
       : _buffer = buffer is Uint8List ? buffer : Uint8List.fromList(buffer),
         _recursionLimit = recursionLimit,
-        _sizeLimit = math.min(sizeLimit, buffer.length) {
+        _sizeLimit = sizeLimit >= 0
+            ? math.min(sizeLimit, buffer.length)
+            : buffer.length {
     _currentLimit = _sizeLimit;
   }
```

osa1: We have #386 about `DEFAULT_SIZE_LIMIT`.

protobuf/lib/src/protobuf/coded_buffer_reader.dart
osa1: Done in #737.

PR #740, opened 2022-08-24
Remove `Extension.repeated` check argument

osa1:
This argument is always passed the same value in the generated code, so we can remove the function parameter and use the value passed by the generated code as the value of the parameter. This saves one line per extension field in the generated code.


sigurdm (approved):
LGTM

osa1:
> The lack of test changes indicates missing coverage.

We have extension tests in protoc_plugin (extension_test.dart).

sigurdm:
But it seems we have no golden-file coverage of the generated code.... maybe that is for the better...

osa1:
Yeah we don't have golden files for extensions.

Tbh I'm not sure why we have golden file tests.. What are they supposed to be checking?

sigurdm:
They are a safety guard for changes in the output.

I guess with perfect functional test-coverage we could argue they would not be useful.
But I am always happy when reviewing that I can follow a change in the generator in the expected output.

osa1:
This replaces a `getCheckFunction(<constant>)` with `getCheckFunction(<runtime value>)`, so that can potentially make things worse.

Closing as it's not 100% clear that this is worth doing.

Issue #705, opened 2022-07-11
`readonlyDefault` for map fields returns mutable map


osa1:
https://github.com/google/protobuf.dart/blob/6a3e7f024a46b9353fd172abe2b80125244cae45/protobuf/lib/src/protobuf/field_info.dart#L160-L165

which for map fields  calls https://github.com/google/protobuf.dart/blob/6a3e7f024a46b9353fd172abe2b80125244cae45/protobuf/lib/src/protobuf/field_info.dart#L282

which uses `PbMap`, instead of `PbMap.unmodifiable`.

osa1:
Repro:
```proto
syntax = "proto3";

message Test {
  map<int32, int32> map_field = 1;
  repeated int32 repeated_field = 2;
  NestedMsg message_field = 3;
}

message NestedMsg {
  int32 int_field = 1;
}
```
```dart
import '../test.pb.dart';
import 'package:protobuf/protobuf.dart';

void main() {
  final msg = (Test().freeze()) as Test;

  final map = (msg.getDefaultForField(1)) as PbMap;
  map[1] = 2; // works
  print('OK');

  final nestedMsg = msg.getDefaultForField(3) as NestedMsg;
  try {
    nestedMsg.intField = 123; // fails as expected
  } catch (e) {
    print('OK');
  }

  final list = (msg.getDefaultForField(2)) as PbList;
  try {
    list.add(1); // fails as expected
  } catch (e) {
    print('OK');
  }
}
```

PR #741, opened 2022-08-25
Avoid redundant shallow copy when getting read-only map fields

osa1:
`_FieldSet._$getMap` unnecessarily shallow-copies the map field when the
message is read-only.

If a message is read-only then all of its fields are read-only, so the
map fields are read-only. We don't need to copy them.

Copying map fields on every access can be expensive depending on the map
size, so this will potentially be a significant performance win.

(Also, if we had to copy, we would have to deep-copy, rather than
shallow-copy, so the code would be buggy if we needed to copy the field)

Also fix `_isReadonly` field initialization in `PbMap.unmodifiable`.

We actually do not use `PbMap.unmodifiable` and we don't have any users
of it, but I think that method will be useful when fixing #705, so I'm
not removing it for now.

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -13,6 +13,9 @@
   getter `length` removed. ([#721])
 * Update library documentation to hide internals, add documentation for public
   types. ([#681])
+* Avoid copying when reading map fields of read-only messages. ([#741])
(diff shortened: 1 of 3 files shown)


osa1:
I'll add changelog entries.

sigurdm (approved):
LGTM

PR #742, opened 2022-08-25
Implement deepCopy as copying, instead of merge

### LINKED ISSUE #60: Cloning should avoid validation checks

osa1:
Currently `deepCopy` and `clone` create an empty message, and then merge the
cloned message into the empty message.

This is slow as merge needs to validate each element added to the message.

In this PR we instead directly copy the field set, and deeply copy map, list,
and message fields.

deepCopy benchmark results:

|                  | Before  | After   | Diff   |
|------------------|---------|---------|--------|
| AOT              | 238 us  | 151 us  | -36%   |
| dart2js -O4      | 242 us  | 140 us  | -42%   |
| dart2wasm -O2    | 230 us  | 123 us  | -46%   |
| JIT              | 189 us  | 104 us  | -44%   |

This CL is tested internally in cl/794080754.

Fixes #60.

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 5.0.0-wip
 
+* Improve performance of `GeneratedMessage.deepCopy`. ([#742])
+
+[#742]: https://github.com/google/protobuf.dart/pull/742
+
 ## 4.2.0
 
 * Internal refactoring to split the package into libraries. This allows
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -937,11 +937,11 @@ class FieldSet {
       if (fieldInfo.isMapField) {
         final PbMap? originalMap = original._values[index];
         if (originalMap == null) continue;
-        _values[index] = originalMap.deepCopy();
+        _values[index] = originalMap._deepCopy();
       } else if (fieldInfo.isRepeated) {
         final PbList? originalList = original._values[index];
         if (originalList == null) continue;
-        _values[index] = originalList.deepCopy();
+        _values[index] = originalList._deepCopy();
       } else if (fieldInfo.isGroupOrMessage) {
         final GeneratedMessage? message = original._values[index];
         _values[index] = message?.deepCopy();
@@ -953,13 +953,13 @@ class FieldSet {
     assert(_extensions == null);
     final originalExtensions = original._extensions;
     if (originalExtensions != null) {
-      _extensions = originalExtensions.deepCopy(this);
+      _extensions = originalExtensions._deepCopy(this);
     }
 
     assert(_unknownFields == null);
     final originalUnknownFields = original._unknownFields;
     if (originalUnknownFields != null) {
-      _unknownFields = originalUnknownFields.deepCopy();
+      _unknownFields = originalUnknownFields._deepCopy();
     }
 
     assert(_unknownJsonData == null);
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -969,7 +969,7 @@ class FieldSet {
     }
 
     assert(_oneofCases == null || _oneofCases.isEmpty);
-    final originalOneofCases = _oneofCases;
+    final originalOneofCases = original._oneofCases;
     if (originalOneofCases != null) {
       _oneofCases!.addAll(originalOneofCases);
     }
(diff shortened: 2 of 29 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -897,4 +900,35 @@ class _FieldSet {
...
+      } else if (value is GeneratedMessage) {
+        values[valueIdx] = value.deepCopy(freeze: freeze);
+      } else {
+        values[valueIdx] = value;
+      }
+    }
+
+    final copy = _FieldSet._(
+      message,
+      null, // TODO: event plugin
```
sigurdm, line 922: Please leave a LDAP or an issue number in todos.

```suggestion
      null, // TODO(omersa): event plugin
```

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -897,4 +900,35 @@ class _FieldSet {
...
+      values,
+      null,
+      _unknownFields?.deepCopy(freeze: freeze),
+      freeze, // TODO: Can we use `_frozenState` here?
```
sigurdm, line 926: Same here

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -200,4 +204,26 @@ class PbList<E> extends ListBase<E> {
   static Never _readOnlyError(String methodName) {
     throw UnsupportedError("'$methodName' on a read-only list");
   }
+
+  PbList<E> deepCopy({bool freeze = false}) {
+    // TODO: We know the capacity, is it possible to allocate a list with the right size?
+    final wrappedList = <E>[];
```
sigurdm, line 210: You could possibly use `List.generate` to fill in the right values from the start, thereby avoiding the re-allocation.

osa1:
This causes a lot of breakage internally. I think one of the reasons is probably that I'm assuming `PbList` for repeated fields and `PbMap` for map fields, but map and list allocation is done via `GeneratedMessage.createRepeatedField` (returns `List<T>`) and `GeneratedMessage.createMapField` (returns `Map<K, V>`). I should check the field type (instead of field value type), and use `Map` and `List` API similar to how we merge map and list fields.

Since we can't have nested lists and maps, cloning of nested maps and lists will have to go through `GeneratedMessage.deepCopy`, which will again look at field types and use List/Map API to clone maps and lists. So we won't need `deepCopy` in `PbMap` and `PbList` types.

Interestingly, we assume `PbList` and `PbMap` in `_FieldSet._markReadOnly`:

```dart
void _markReadOnly() {
  if (_isReadOnly) return;
  _frozenState = true;
  for (var field in _meta.sortedByTag) {
    if (field.isRepeated) {
      PbList? list = _values[field.index!];
      if (list == null) continue;
      list.freeze();
    } else if (field.isMapField) {
      PbMap? map = _values[field.index!];
      if (map == null) continue;
      map.freeze();
    } else if (field.isGroupOrMessage) {
      final entry = _values[field.index!];
      if (entry != null) {
        (entry as GeneratedMessage).freeze();
      }
    }
  }

  _extensions?._markReadOnly();
  _unknownFields?._markReadOnly();
}
```

If I'm right that the breakage here caused by users overriding list and map field types, then I'm not sure how `_markReadOnly` works. Maybe users that override map and list types never freeze their messages?

We should just rewrite this library..

osa1:
I managed to fix at least some of the downstream breakage by removing the assumptions in the implementation about repeated fields being `PbList` and map fields being `PbMap`.

The fact that we csan't assume this means freezing messages is buggy, for the reasons described in my previous comment above.

Performance is still much better than merge-based implementation. I will update the PR description, and refactor the code a little bit (document, revert some of the changes).

osa1:
> I managed to fix at least some of the downstream breakage by removing the assumptions in the implementation about repeated fields being PbList and map fields being PbMap.

This version fixed some of the previous failures, but introduced new ones.

osa1:
We should revive this as it improves cloning performance quite a bit (50% according to my original report in PR description).

As far as I understand from the discussion above the main issue was repeated and map fields were overridden by message classes, but we've since removed those members (96d95225e1843427728572647393b83808534577) so this should probably be good now.

It might make sense to just implement this again from scratch instead of fixing this PR.

protobuf/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 5.0.0-wip
 
+* Improve performance of `GeneratedMessage.deepCopy`. ([#742])
```
devoncarew, line 3: We should update the pubspec version to agree w/ the changelog (=> `5.0.0-wip`). If we do plan to publish after landing this (no particular opinion), we can change both versions to `5.0.0`).

osa1: Didn't we discuss this already? https://github.com/google/protobuf.dart/pull/1037#discussion_r2258129449

5.0.0-wip breaks the build because it's not considered 5.0.0, and plugin now needs 5.0.0.

We can update the changelog before or even after releasing 5.0.0.

PR #745, opened 2022-08-26
Two bug fixes in map field binary parsing

### LINKED ISSUE #719: Decoder throws exception when decoding an empty map
I'm sending a message from the Rust backend which has almost all fields set to default (either 0 or empty string or empty `repeated`). I understand that in such case protobuf just "optimizes away" such values. Upon receiving this message the Dart code throws a `nullcheck` exception from runtime. 

Here's the `.proto` code: 
```proto
syntax = "proto3";
message ProtoPara {
    string id = 1;
}
message ProtoPage {
    uint32  page_id = 1;
    repeated ProtoPara paras = 2;
}
message ProtoBoard {
    string id       = 1;
    string team_id  = 2;
    map<uint32, ProtoPage> pages = 3;
}
message ProtoMessage {
    oneof msg {
        ProtoBoard board = 1;
    }
    optional string request_id = 100;
}
```

The message only has `ProtoBoard.id` set to something and also the `ProtoMessage.request_id`. The `pages` map contains a default a single value of `ProtoPage` (`page_id` is 0 and `paras` array is empty). A binary dump of the message which is read by `protoc --decode` without a problem: `ChAKDDNIVkJHVGpJODBLZhoAogYRb3BlbmJvYXJkTmpkOTc3ZjA=`

```
$ echo "ChAKDDNIVkJHVGpJODBLZhoAogYRb3BlbmJvYXJkTmpkOTc3ZjA=" | base64 -d |protoc -I . --decode ProtoMessage proto.proto
board {
  id: "3HVBGTjI80Kf"
  pages {
    key: 0
    value {
    }
  }
}

$ dart pub global list
protoc_plugin 20.0.1
```

protobuf package is at 2.1.0

Stack trace:
```
packages/protobuf/src/protobuf/pb_map.dart 110:73                                 [_mergeEntry]
packages/protobuf/src/protobuf/coded_buffer.dart 190:48                           _mergeFromCodedBufferReader
packages/protobuf/src/protobuf/generated_message.dart 170:5                       mergeFromCodedBufferReader
packages/protobuf/src/protobuf/coded_buffer_reader.dart 103:12                    readMessage
packages/protobuf/src/protobuf/coded_buffer.dart 122:14                           _mergeFromCodedBufferReader
packages/protobuf/src/protobuf/generated_message.dart 185:5                       mergeFromBuffer
packages/myprotocol/generated/myprotocol.pb.dart 1195:130                  <fn>
packages/myprotocol/generated/myprotocol.pb.dart 1195:150                  fromBuffer
packages/myapp/backend/websocket.dart 129:42                                 receive
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 325:40               runBody
dart-sdk/lib/async/zone.dart 1685:54                                              runUnary
```
(skipped all below zone)

Changelog refers to some similar bug fixed in 2.0.1 but I can' seem to be able to find it.

osa1:
Map fields are encoded as repeated message fields where the message
field 1 is the key and 2 is the value.

In proto syntax, a map field like

    map<int32, MyMessage> map_field = 1;

is encoded as if it was

    repeated MapFieldEntry map_field = 1;

where

    message MapFieldEntry {
        optional int32 key = 1;
        optional MyMessage value = 2;
    }

Since map entries are ordinary messages, it's possible for a map entry
to have no key or value fields.

This PR updates map decoding to handle missing key and value fields.
Three tests added for (1) missing key field (2) missing value field (3)
missing key and value fields (entry message has 0 length prefix).

Fixes #719

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -16,7 +16,8 @@
 * Avoid copying when reading map fields of read-only messages. ([#741])
 * Fix `PbMap._isReadonly` field initialization in `PbMap.unmodifiable`.
   ([#741])
-* Fix two bugs in map field decoding. ([#719], [#745])
+* Fix decoding map fields when key or value (or both) fields of a map entry is
+  missing. ([#719], [#745])
 
 [#183]: https://github.com/google/protobuf.dart/issues/183
 [#644]: https://github.com/google/protobuf.dart/pull/644
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Avoid copying when reading map fields of read-only messages. ([#741])
 * Fix `PbMap._isReadonly` field initialization in `PbMap.unmodifiable`.
   ([#741])
+* Fix two bugs in map field decoding. ([#719], [#745])
 
 [#183]: https://github.com/google/protobuf.dart/issues/183
 [#644]: https://github.com/google/protobuf.dart/pull/644
@@ -29,6 +30,8 @@
 [#721]: https://github.com/google/protobuf.dart/pull/721
 [#681]: https://github.com/google/protobuf.dart/pull/681
 [#741]: https://github.com/google/protobuf.dart/pull/741
+[#719]: https://github.com/google/protobuf.dart/issues/719
+[#745]: https://github.com/google/protobuf.dart/pull/745
 
 ## 2.1.0
 
--- a/protobuf/lib/src/protobuf/builder_info.dart
+++ b/protobuf/lib/src/protobuf/builder_info.dart
@@ -241,11 +241,12 @@ class BuilderInfo {
       List<ProtobufEnum>? enumValues,
       ProtobufEnum? defaultEnumValue,
       PackageName packageName = const PackageName(''),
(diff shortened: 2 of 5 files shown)


sigurdm (approved):
LGTM

osa1:
I just realized that proto2 and proto3 specs give different encodings for `map` fields:

```
map<int32, MyMessage> my_messages = 1;

===>

repeated MyMessagesEntry my_messages = 1;

// According to proto2 spec
message MyMessagesEntry {
    optional int32 key = 1;
    optional MyMessage value = 2;
}

// According to proto3 spec
message MyMessagesEntry {
    int32 key = 1;
    MyMessage value = 2;
}
```

These two encodings are not the same, for example, if a `value` fields is missing, proto2  implementation might return "null" for the value, while a proto3 implementation needs to return the default value for the message (which may or may not be null).

So I'm not sure if this PR is correct. I'll try get proto team's opinion on this.

osa1:
I asked about this to protobuf team. The key point is `{key: 0, value: {}}` should be identical to `{}` after the parse step. So we should handle missing key and value fields as we handle missing fields in messages that are not map entries.

PR #750, opened 2022-09-06
Update to latest pkg:mono_repo, use new pubspec feature

kevmoo:


osa1 (approved):
Thanks!

osa1:
@kevmoo with this PR we started to use SDK dev builds in SDK, is this expected? It causes CI breakage that we cannot reproduce with the latest stable release (2.18.0), see https://github.com/google/protobuf.dart/runs/8222499751?check_suite_focus=true.

kevmoo:
@osa1 – we generally find it's good to look forward to dev releases. We become canaries for the rest of the ecosystem.

The fixed the break from that CI run should be easy!

PR #753, opened 2022-09-14
Make subBuilder args of BuilderInfo message methods required

osa1:
We always pass those args, and not passing them would be a bug as we can't deserialize a message without the `subBuilder` field.


sigurdm (approved):
LGTM

PR #759, opened 2022-10-06
Simplify CodedBufferWriter._copyInto

osa1:
- Make it a static method
- Use `Uint8List.setAll` instead of a loop to set range

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -420,22 +420,16 @@ class CodedBufferWriter {
 
   /// Copy bytes from the given typed data array into the output buffer.
   ///
-  /// Has a specialization for Uint8List for performance.
-  int _copyInto(Uint8List buffer, int pos, TypedData value) {
+  /// Has a specialization for [Uint8List] for performance.
+  static int _copyInto(Uint8List buffer, int pos, TypedData value) {
     if (value is Uint8List) {
-      var len = value.length;
-      for (var j = 0; j < len; j++) {
-        buffer[pos++] = value[j];
-      }
-      return pos;
+      buffer.setAll(pos, value);
+      return pos + value.length;
     } else {
-      var len = value.lengthInBytes;
-      var u8 = Uint8List.view(
-          value.buffer, value.offsetInBytes, value.lengthInBytes);
-      for (var j = 0; j < len; j++) {
-        buffer[pos++] = u8[j];
-      }
-      return pos;
+      final len = value.lengthInBytes;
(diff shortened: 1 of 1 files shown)


osa1:
@mraleph could you review please? I'm thinking maybe with `Uint8List` methods we can avoid bounds checking in every iteration, but I can't find where the methods are implemented..

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -420,22 +420,16 @@ class CodedBufferWriter {
...
-      var len = value.lengthInBytes;
-      var u8 = Uint8List.view(
-          value.buffer, value.offsetInBytes, value.lengthInBytes);
-      for (var j = 0; j < len; j++) {
-        buffer[pos++] = u8[j];
-      }
-      return pos;
+      final len = value.lengthInBytes;
+      final u8 = Uint8List.view(value.buffer, value.offsetInBytes, len);
+      buffer.setAll(pos, u8);
```
sigurdm, line 431: Also this call seems it could be pulled out of the conditional by setting u8 with a ternary operator.

sigurdm (approved):
LGTM

just a few nits on how the code could perhaps be further simplified.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -420,23 +420,14 @@ class CodedBufferWriter {
...
-      var len = value.lengthInBytes;
-      var u8 = Uint8List.view(
-          value.buffer, value.offsetInBytes, value.lengthInBytes);
-      for (var j = 0; j < len; j++) {
-        buffer[pos++] = u8[j];
-      }
-      return pos;
-    }
+  /// Has a specialization for [Uint8List] for performance.
+  static int _copyInto(Uint8List buffer, int pos, TypedData value) {
```
mraleph, line 424: The code was originally written in this style to ensure good specialised loop for `Uint8List` to `Uint8List` copying. Using `setAll` does not yield the same good code unfortunately. 

I have confirmed that this is a regression both on `to_binary` and `query_encode_binary` benchmarks:

plot
plot-to-binary

These are M1 measurements.

osa1: Thanks! Closing.

sigurdm:
Can't we fix this in the sdk? shouldn't `setAll` have potential for being much faster (basically a memcopy)?

mraleph:
Yeah, we would like to fix this eventually in the SDK (e.g. https://github.com/dart-lang/sdk/issues/40039 is tracking a similar thing), but so far we did not have time to do it.

PR #762, opened 2022-10-11
Fix initializing required fields with proto3 JSON null values

### LINKED ISSUE #760: proto3 JSON required field confusion
proto3 does not have required fields, so [proto3 JSON format spec][1] does not specify how required fields should be handled.

This is fine, but we allow serializing proto2 messages as proto3 JSON. So users do it and get confusing results. Example:

```proto
syntax = "proto2";

message MyMessage {
    optional MyOtherMessage my_message_field = 1;
}

message MyOtherMessage {
    required int32 field1 = 1 [default = 123];
    required int32 field2 = 2 [default = 456];
}
```

```dart
import '../test.pb.dart';

void main() {
  {
    final json = {'my_message_field': null};
    final myMessage = MyMessage()..mergeFromProto3Json(json);
    print(myMessage.myMessageField.field1);
    print(myMessage.toProto3Json());
  }

  {
    final json = {'my_message_field': {}}; // required fields missing
    final myMessage = MyMessage()..mergeFromProto3Json(json);
    print(myMessage.myMessageField.field1);
    print(myMessage.toProto3Json());
  }
}
```

Output:

```
123
{myMessageField: {}}
123
{myMessageField: {}}
```

Problems:

1. Deser->ser changes message from `{myMessageField: null}` to `{myMessageField: {}}`
2. Deserialization accepts missing required fields
3. Serialization omits required fields

I think (2) and (3) are consistent with binary serialization and work as intended, but (1) is potentially a problem as depending on the implementation on the receiving end it changes a message with non-existent optional field into a message with an invalid field value.

[1]: https://developers.google.com/protocol-buffers/docs/proto3#json

osa1:
[proto3 JSON spec][1] says this about `null` values:

> If a value is missing in the JSON-encoded data or if its value is
> null, it will be interpreted as the appropriate default value when
> parsed into a protocol buffer

Normally empty field set represents the default message, but for that representation to work with proto2 required fields, serializers need to check for required fields in the `BuilderInfo` and for fields that are not in the field set serialize the default value for the field.

Currently we don't do this, which results in generating messages with missing required fields. See #760 for a repro.

Checking missing required fields in serializers should fix the issue, but as a simpler fix, this PR deserializes proto3 JSON `null` values to field sets with required fields set to defaults. This way serializers do not need to be tweaked.

Fixes #760

[1]: https://developers.google.com/protocol-buffers/docs/proto3#json

### DIFF
--- a/protobuf/lib/src/protobuf/proto3_json.dart
+++ b/protobuf/lib/src/protobuf/proto3_json.dart
@@ -323,12 +323,14 @@ void _mergeFromProto3Json(
       }
     }
 
+    final meta = fieldSet._meta;
+
     if (json == null) {
-      // `null` represents the default value. Do nothing more.
+      // `null` represents the default value, initialize required fields
+      _initializeRequiredFields(fieldSet);
       return;
     }
 
-    final meta = fieldSet._meta;
     final wellKnownConverter = meta.fromProto3Json;
     if (wellKnownConverter != null) {
       wellKnownConverter(fieldSet._message!, json, typeRegistry, context);
@@ -414,4 +416,27 @@ void _mergeFromProto3Json(
   }
 
   recursionHelper(json, fieldSet);
+}
+
+void _initializeRequiredFields(_FieldSet fieldSet) {
+  final meta = fieldSet._meta;
+
+  if (!meta.hasRequiredFields) {
+    return;
+  }
+
+  for (final fieldInfo in meta.sortedByTag) {
+    if (!fieldInfo.isRequired) {
+      continue;
+    }
+
+    var value = fieldInfo.makeDefault!();
+
+    if (value is GeneratedMessage) {
+      value = value.deepCopy();
+      _initializeRequiredFields(value._fieldSet);
+    }
+
+    fieldSet._setField(fieldInfo.tagNumber, value);
+  }
 }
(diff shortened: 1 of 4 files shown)


protobuf/lib/src/protobuf/proto3_json.dart
```diff
@@ -415,3 +417,26 @@ void _mergeFromProto3Json(
...
+  if (!meta.hasRequiredFields) {
+    return;
+  }
+
+  for (final fieldInfo in meta.sortedByTag) {
+    if (!fieldInfo.isRequired) {
+      continue;
+    }
+
+    var value = fieldInfo.makeDefault!();
```
osa1, line 433: I think this probably generates immutable lists and maps, which we should unfreeze below.

osa1: Ah, but maps and lists cannot be `required`. So I guess this is fine.

osa1:
I asked protobuf team about this fix. It turns out we have an internal version of Dart protobuf spec that says this about the proto3 JSON format:

> When parsing, a “null” value is accepted for all field types. It will be treated as if the field has the default value (the field itself would not be set, but its getter will return the default value for proto3).

This version is not in conflict with the public version, but it has extra information. Namely, it says the field should not be set when we see `null`.

(The reason why the public version doesn't have this information is probably because it doesn't talk about presence yet, and without presence it doesn't make sense to talk about setting fields)

According to this version, this fix is not correct, because the hazzer for the message-field and the message's fields will return `true`.

I want to avoid changing serializers because (1) it will incur runtime overhead (2) we need to update each serializer and test them separately.

Maybe we can implement some common code shared between all of the serializers for iterating over fields with values?

PR #765, opened 2022-10-17
protoc_plugin: Improve "is map field" check

osa1:
`MessageOptions.map_entry` is an `optional bool` field. Currently to check whether a field is a map we check whether `map_entry` is present, but we don't check its value. This only works if the field is never set, or set to `true` (i.e. it's never set to `false`).

The right way to check it is

    descriptor.options.hasMapEntry() && descriptor.options.mapEntry

because the default value of the field is `false`, this can be simplified as

    descriptor.options.mapEntry


sigurdm (approved):
LGTM

PR #766, opened 2022-10-17
protoc_plugin: Reword some of the documentation

osa1:
- Convert "Returns true if ..." to "Whether ...", for consistency with other documentation and Effective Dart guidelines.

- Add a note in `ProtobufField.hasPresence` to mention that `isRepeated` is true for map fields as well.

- Remove example in `getDartType` documentation. It sounds like `List` is a special case.


sigurdm (approved):
LGTM

Issue #768, opened 2022-11-02
[proposal] exclude generated files from code coverage


iapicca:
### Use case

currently generated files negatively affect code coverage

### Proposal

add the comment `// coverage:ignore-file`, 
made available by [dart tools coverage](https://pub.dev/packages/coverage) (see https://github.com/dart-lang/coverage/pull/302),
to the generate files to exclude these from coverage

cc @osa1

osa1:
Thanks for the ping. Could you give us some context please? What is coverage? How do you run it? What's the problem not having `// converage:...` causes?

osa1:
Hi @iapicca. We discussed this briefly and coverage devs opened an issue for adding a command line flag to coverage to ignore files: https://github.com/dart-lang/coverage/issues/427. This coverage feature is much more general, you can ignore files generated by other code generators as well, and with that feature we don't need to change anything in this library.

iapicca:
Fair enough, thanks for checking this out @osa1 

For anyone that is reading this issue and uses [codecov](https://about.codecov.io/)
you may also add to your `codecov.yml`
```yaml
ignore:
  - '**/*.pb.dart'
```

PR #769, opened 2022-11-02
Exclude generated files from code coverage

iapicca:
fix https://github.com/google/protobuf.dart/issues/768

### DIFF
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -616,6 +616,7 @@ class FileGenerator extends ProtobufContainer {
 //
 // @dart = 2.12
 // ignore_for_file: ${ignores.join(',')}
+// coverage:ignore-file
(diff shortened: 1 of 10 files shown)


osa1 (commented):
These files are generated by the plugin so if we want any changes on them the plugin should be generating the changes.

If you could give us some context in the issue (see my question) I can help with the implementation.

iapicca:
> These files are generated by the plugin so if we want any changes on them the plugin should be generating the changes.

sorry, I thought they were something like a template since they are in the repo,  shouldn't the `generated` folder be in `.gitignore` if it's just for testing?

> If you could give us some context in the issue (see my question) I can help with the implementation.

@osa1 
I added the line `// coverage:ignore-file` to `FileGenerator`'s  `_writeHeading`'s boilerplate
please let me know if there is any other action I should take 
(after you vacations, no hurry :) )

Issue #771, opened 2022-11-09
improve the readability of the generated code


devoncarew:
Hi; I have some ideas for changes that could improve the readability of the generated code. These would all be non-functional changes; they wouldn't effect the compiled size of the code or the runtime performance.

The readability of the generated code probably isn't a critical consideration here, so I thought I'd open an issue to discuss before sending PRs. Some potential changes:

- move the `n` copies of `const $core.bool.fromEnvironment('protobuf.omit_field_names')` in the generated files into a single `_omitFieldNames` const in each file
- address a few of the ~12 lints we ignore in the generated code

cc @osa1

osa1:
@devoncarew we merged a few PRs about this, can this be closed as complete or do you have more changes planned?

devoncarew:
I think we're good to close this - any other things we notice wrt readability can just be tackled at the time / filed as separate issues.

PR #772, opened 2022-11-11
Use shared consts for bool.fromEnvironment expressions

devoncarew:
This improves the readability of the generated code (see https://github.com/google/protobuf.dart/issues/771 for context).

- use shared consts to reduce the number of repeated `bool.fromEnvironment()` expressions (all the `bool.fromEnvironment()` expressions are gathered up while generated the file and written as consts at the end of the file)
- turn on a few additional lints
- update the changelog

Done as three commits for easier review (the dart code changes, regenerating the lib/src/generated files, and regenerating the test goldens). cc @osa1

### DIFF
--- a/protoc_plugin/lib/indenting_writer.dart
+++ b/protoc_plugin/lib/indenting_writer.dart
@@ -27,6 +27,9 @@ class IndentingWriter {
   int _previousOffset = 0;
   final String? _sourceFile;
 
+  // Named text sections to write at the end of the file.
+  Map<String, String> suffixes = {};
+
   IndentingWriter({String? filename}) : _sourceFile = filename;
 
   /// Appends a string indented to the current level.
@@ -101,8 +104,23 @@ class IndentingWriter {
     }
   }
 
+  void addSuffix(String suffixName, String text) {
+    suffixes[suffixName] = text;
+  }
+
   @override
-  String toString() => _buffer.toString();
+  String toString() {
+    if (suffixes.isNotEmpty) {
+      // TODO: We may want to introduce the notion of closing the file stream.
+      println('');
+      for (var key in suffixes.keys.toList()..sort()) {
+        println(suffixes[key]!);
+      }
+      suffixes.clear();
+    }
+
+    return _buffer.toString();
+  }
 
   /// Writes part of a line of text.
   /// Adds indentation if we're at the start of a line.
--- a/protoc_plugin/lib/indenting_writer.dart
+++ b/protoc_plugin/lib/indenting_writer.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+
 import 'src/generated/descriptor.pb.dart';
 
 /// Specifies code locations where metadata annotations should be attached and
@@ -28,7 +30,7 @@ class IndentingWriter {
   final String? _sourceFile;
 
   // Named text sections to write at the end of the file.
-  Map<String, String> suffixes = {};
+  final Map<String, String> _suffixes = SplayTreeMap();
 
   IndentingWriter({String? filename}) : _sourceFile = filename;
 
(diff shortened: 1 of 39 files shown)


protoc_plugin/lib/indenting_writer.dart
```diff
@@ -27,6 +27,9 @@ class IndentingWriter {
   int _previousOffset = 0;
   final String? _sourceFile;
 
+  // Named text sections to write at the end of the file.
+  Map<String, String> suffixes = {};
```
osa1, line 31: Should this be private? (`_suffixes`)

Since we iterate this map based on ordering of keys, can we use another type here that gives us ordered iteration by default? I think `SplayTreeMap` does that.

protoc_plugin/lib/indenting_writer.dart
```diff
@@ -101,8 +104,23 @@ class IndentingWriter {
...
 
+  void addSuffix(String suffixName, String text) {
+    suffixes[suffixName] = text;
+  }
+
   @override
-  String toString() => _buffer.toString();
+  String toString() {
+    if (suffixes.isNotEmpty) {
+      // TODO: We may want to introduce the notion of closing the file stream.
```
osa1, line 114: This class doesn't have a file stream, what is the issue here? Should we open a GitHub issue for this?

protoc_plugin/lib/indenting_writer.dart
```diff
@@ -101,8 +104,23 @@ class IndentingWriter {
     }
   }
 
+  void addSuffix(String suffixName, String text) {
+    suffixes[suffixName] = text;
+  }
```
osa1, line 109: Documenting this method would be good. It's a bit surprising that keys are only used for to sort the printed lines and they're not printed.

(I wonder if we could use a better key type here? The string contents are not useful and are used to determine the order of lines. What would be a better key type for this?)

protoc_plugin/test/send_protos_via_sendports_test.dart
```diff
@@ -15,7 +15,7 @@ Future<T> sendReceive<T>(T object) async {
   return (await rp.first) as T;
 }
 
-Future main() async {
+void main() async {
```
osa1, line 18: I'm surprised that this works.. `main` doesn't have a return. Why is this not a type error? I thought the `async` suffix in the header means a `Future` result, and return type specifies what the returned future resolves to. So for example this function would previously return `Future<Void>`, and now it's `Future<Future>`.

osa1 (approved):
LGTM, but added some inline comments and questions.

protoc_plugin/lib/indenting_writer.dart
devoncarew: I switched this to `_suffixes` (private);

Changing to using a type w/ ordering built-in makes sense; I updated the type we create to `SplayTreeMap` and removed the explicit key sorting where we read from the map below.

protoc_plugin/lib/indenting_writer.dart
devoncarew: +1 - I'll add some dartdoc here.

The keys are really the names of the fields that we emit (the field names are the keys, and the contents of the entire declaration - name and all - are the text). But the key is really meant to be opaque to the method here - just something to keep the suffixes unique. A field declaration will be written and over-written several times when generating a file - we just depend on the usage pattern here that the name and the contents are the same each time.

devoncarew: I changed the param name to `suffixKey` and indicated in the dartdoc that we don't emit that param, just use it for suffix uniqueness.

protoc_plugin/lib/indenting_writer.dart
devoncarew: Sorry, I should have said `writer` here - the `IndentingWriter` class.

What I meant by this todo was that I'm mutating the state in the toString() method which isn't 100% clean behavior. It would be better to have a `close()` method on the writer, have people call that when finished generated code, and only emit the suffixes in close() and not in toString().

protoc_plugin/test/send_protos_via_sendports_test.dart
devoncarew: I think this is special cased in the type system - this is effectively a `Future<void>`. I'm not fully sure of that though; I did see a good explanation of this on an issue by @eernstg a while ago.

devoncarew (commented):
Updated based on review comments -

protoc_plugin/test/send_protos_via_sendports_test.dart
eernstg: There is actually a difference. `void f(...) async {...}` is a "fire-and-forget" asynchronous function, which is useful in the case where awaiting the completion of the execution of the body of `f` at a call site is unlikely to be useful. For instance:

```dart
void f() async {
  await null;
  print('... do some background work that nobody else should await ...');
}

void main() async {
  print('main');
  f(); // `await f()` is a compile-time error.
  await g(); // `print(await g())` is a compile-time error.
  print('end of main');
}
```

In contrast, `Future<void> g(...) async {...}` is used when it _is_ useful for a caller to synchronize with the completion of the work done in the callee, but the returned value should be ignored. So `await g(...);` is fine, but `print(await g())` is a compile-time error.

osa1: Thanks @eernstg, that's very helpful.

I just realized that I read this as change from `void` to `Future`, but it actually changes `Future` to `void`. So I'm surprised that the current version works, not the version with this change.

protoc_plugin/lib/indenting_writer.dart
osa1: We could also just rename `toString` to something else, maybe `finish` or something? In any case we can do it later.

PR #773, opened 2022-11-17
Remove duplicate consts

devoncarew:
A few more cleanups to the generated code:
- remove unnecessary consts (and the `ignore: unnecessary_const` directive)
- when we print out base64, don't emit lines longer than 80 chars
- adjust import ordering

### DIFF
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -454,7 +454,8 @@ class FileGenerator extends ProtobufContainer {
       [OutputConfiguration config = const DefaultOutputConfiguration()]) {
     if (!_linked) throw StateError('not linked');
     var out = makeWriter();
-    _writeHeading(out);
+    _writeHeading(out,
+        extraIgnores: {'deprecated_member_use_from_same_package'});
 
     if (serviceGenerators.isNotEmpty) {
       out.println(_asyncImport);
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -528,12 +528,24 @@ class FileGenerator extends ProtobufContainer {
 
   void writeBinaryDescriptor(IndentingWriter out, String identifierName,
       String name, GeneratedMessage descriptor) {
-    var descriptorText = base64Encode(descriptor.writeToBuffer());
+    var base64 = base64Encode(descriptor.writeToBuffer());
     out.println('/// Descriptor for `$name`. Decode as a '
         '`${descriptor.info_.qualifiedMessageName}`.');
+
+    var base64Lines = splitString(base64, 74).map((s) => "'$s'").join();
     out.println('final $_typedDataImportPrefix.Uint8List '
         '$identifierName = '
-        '$_convertImportPrefix.base64Decode(\'$descriptorText\');');
+        '$_convertImportPrefix.base64Decode($base64Lines);');
+  }
+
+  List<String> splitString(String str, int segmentLength) {
+    var result = <String>[];
+    while (str.length >= segmentLength) {
+      result.add(str.substring(0, segmentLength));
+      str = str.substring(segmentLength);
+    }
+    if (str.isNotEmpty) result.add(str);
+    return result;
   }
 
   /// Returns the contents of the .pbjson.dart file for this .proto file.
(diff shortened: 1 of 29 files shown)


devoncarew:
(part of https://github.com/google/protobuf.dart/issues/771)

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -528,12 +526,24 @@ class FileGenerator extends ProtobufContainer {
...
         '`${descriptor.info_.qualifiedMessageName}`.');
+
+    var base64Lines = splitString(base64, 74).map((s) => "'$s'").join();
     out.println('final $_typedDataImportPrefix.Uint8List '
         '$identifierName = '
-        '$_convertImportPrefix.base64Decode(\'$descriptorText\');');
+        '$_convertImportPrefix.base64Decode($base64Lines);');
+  }
+
+  List<String> splitString(String str, int segmentLength) {
```
osa1, line 539:
```suggestion
  static List<String> _splitString(String str, int segmentLength) {
```

We could also move this to a utility library, along with the stuff in `string_escape.dart`. We may also have other fairly general utility functions elsewhere in the package.

protoc_plugin/test/goldens/topLevelEnum.pbjson
```diff
@@ -5,27 +5,27 @@
...
+    {'1': 'MOBILE', '2': 0},
+    {'1': 'HOME', '2': 1},
+    {'1': 'WORK', '2': 2},
+    {'1': 'BUSINESS', '2': 2},
   ],
 };
 
 /// Descriptor for `PhoneType`. Decode as a `google.protobuf.EnumDescriptorProto`.
-final $typed_data.Uint8List phoneTypeDescriptor = $convert.base64Decode('CglQaG9uZVR5cGUSCgoGTU9CSUxFEAASCAoESE9NRRABEggKBFdPUksQAhIMCghCVVNJTkVTUxAC');
+final $typed_data.Uint8List phoneTypeDescriptor = $convert.base64Decode('CglQaG9uZVR5cGUSCgoGTU9CSUxFEAASCAoESE9NRRABEggKBFdPUksQAhIMCghCVVNJTkVTUx''AC');
```
osa1, line 30: This is longer than 80 characters, is this expected? Is this a problem with the formatter?

I think the intended formatting is something like

```dart
final $typed_data.Uint8List phoneTypeDescriptor = $convert.base64Decode(
  'CglQaG9uZVR5cGUSCgoGTU9CSUxFEAASCAoESE9NRRABEggKBFdPUksQAhIMCghCVVNJTkVTUx'
  'AC');
```

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -685,12 +695,12 @@ const _fileIgnores = {
   'annotate_overrides',
   'camel_case_types',
   'constant_identifier_names',
+  'deprecated_member_use',
   'directives_ordering',
   'library_prefixes',
   'non_constant_identifier_names',
   'prefer_final_fields',
   'return_of_invalid_type',
```
osa1, line 703: Not related to this PR, but I think we can remove this ignore as well. I tried removing it locally and it doesn't generate any warnings in test protos.

osa1 (approved):
LGTM, but I think formatting is not working as expected. (see my inline comment)

protoc_plugin/lib/src/file_generator.dart
devoncarew: sgtm! (I may leave that as work for a future PR)

I in-lined your suggestion into my local edits

protoc_plugin/lib/src/file_generator.dart
devoncarew: I'll remove in this PR (I verified what you were saying locally as well)

protoc_plugin/test/goldens/topLevelEnum.pbjson
devoncarew: Ah, good catch.

I adjusted the output a bit so this will look better w/o formatting, but I think what we're seeing here is that these `test/goldens/*.pbjson` files aren't getting the formatter run on them after generation. (separately, I don't know why they don't end in a .dart extension; they do seem to contain dart code and not json)

devoncarew (commented):
Updated the PR based on review comments -

PR #783, opened 2022-12-07
Fix avoid_renaming_method_parameters linter warning.

xclud:
Fixes the generated file causes the following linter warning:

```
The parameter name 'method' doesn't match the name 'methodName' in the overridden method.
Try changing the name to 'methodName'.
```


osa1 (approved):
Thanks!

PR #784, opened 2022-12-07
Removes a non-existing lint rule.

xclud:
Removes `undefined_shown_name` lint rule which does not exist.

https://dart-lang.github.io/linter/lints/undefined_shown_name.html


osa1 (approved):
Thanks!

PR #791, opened 2023-01-17
Run non-null-safe tests with Dart 2

kevmoo:
Move the non-null-safe usage of the API to another file, test the file with Dart 2.x.

### DIFF
--- a/protoc_plugin/test/generated_message_test.dart
+++ b/protoc_plugin/test/generated_message_test.dart
@@ -177,7 +177,7 @@ void main() {
     // expect(MessageWithNoOuter.getDescriptor().getFile(),
     //        MultipleFilesTestProto.getDescriptor());
 
-    var tagNumber = message.getTagNumber('foreignEnum') as int;
+    var tagNumber = message.getTagNumber('foreignEnum')!;
     expect(message.getField(tagNumber), EnumWithNoOuter.BAR);
 
     // Not currently supported in Dart protobuf.
--- a/protoc_plugin/test/generated_message_test.dart
+++ b/protoc_plugin/test/generated_message_test.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart=2.11
-
 import 'package:protobuf/protobuf.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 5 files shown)


protoc_plugin/test/generated_message_test.dart
```diff
@@ -807,12 +720,6 @@ void main() {
     message.m3M = p3.M_M();
   });
 
-  test('testToplevel', () {
-    var message = t.M();
-    message.t = T();
-    t.SApi(null);
```
kevmoo, line 813: The call to `SApi` is not valid with null. Not sure what else to do other than delete this. 🤷

kevmoo:
Feel free to squash/merge and delete the branch if you're happy

mraleph:
@osa1 I think this might be related to what you looked at today.

osa1:
@kevmoo we've been discussing these tests in Dart Ecosystem chat. I don't think we want to remove these tests yet as internally we have non-null-safe Dart that uses this library.

kevmoo:
@osa1 – but we can't run Dart 3.0.0- tests against this package w/ them there.

We could pull those out into another test file and only run them on stable or something 🤷

osa1:
Right, I think we will need to move these to another file/package, and configure mono_repo to only test that package/file with SDK 2.x.

Apparently `dart:test` has a separate packages for legacy tests, for example https://github.com/dart-lang/test/blob/master/legacy_tests/spawn_hybrid_with_optout/pubspec.yaml. We could do the same. If we can specify running a specific file with a specific SDK in mono_repo.yaml we could also just move the file outside of `test/` (maybe to `legacy_test/`), that would be simpler.

kevmoo:
@osa1 – sure? Just trying to get CI to green so we can keep the signals of knowing when the SDK breaks things.

osa1:
It seems like on CI `dart analyze` analyzes the new file. I don't understand why that happens, locally when I run `dart analyze` with Dart 3.0.x it doesn't analyze the new file.

We need to somehow make `dart analyze` skip the file with dev SDK.

protoc_plugin/test/generated_message_test.dart
osa1: Test moved to legacy test file.

osa1 (approved):
LGTM now, but someone should review my mono_repo changes :-)

devoncarew (approved):
mono_repo changes and repo reconfiguration lgtm

PR #795, opened 2023-01-31
Fix a test

osa1:
This test is currently using an invalid map key type (int instead of string) for indexing and comparing two `null` values. It was valid in older versions where maps were represented as lists, but became invalid when we switched to a map type.

Also updated the length check to rule out the case where lists have the same length, but both are wrong.

Syncs cl/504248137 with some changes.


sigurdm (approved):
LGTM

PR #796, opened 2023-01-31
contribute a publishing automation action

devoncarew:
This PR contributes a publishing automation action. You can find more info here: https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose, but briefly:

- on PRs, general publishing information is validated (pubspec version, changelog version, pub publish dry run, ...)
- after a change (intended for publishing) has landed on the default branch, tagging that commit with a well formed git tag (`package_name-v1.2.3`) will trigger a github workflow which will invoke `dart pub publish --force` w/ the correct package

@osa1 - after landing this there is some minor config that needs to happen in the pub.dev admin pages for package:protobuf and package:protoc_plugin. And fair warning, anyone with push access to this repo will be able to trigger a publish (it's possible to lock it down more than that - using github environments and some additional signoff on publish - but that's not enabled by default).

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ master ]
+  push:
+    tags: [ 'v[0-9]+.[0-9]+.[0-9]+*' ]
+
+jobs:
+  publish:
+    if: ${{ github.repository_owner == 'google' }}
+    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,7 +6,7 @@ on:
   pull_request:
     branches: [ master ]
   push:
-    tags: [ 'v[0-9]+.[0-9]+.[0-9]+*' ]
+    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
 
 jobs:
   publish:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -3,11 +3,16 @@
 name: Publish
 
 on:
+  # Trigger on pull requests that target the default branch.
   pull_request:
     branches: [ master ]
+  # Also, trigger when tags are pushed to the repo.
   push:
     tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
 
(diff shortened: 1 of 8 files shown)


devoncarew:
Interesting; the 2.12 bot doesn't recognize the `pubspec_overrides.yaml` file, so its not getting the version of protobuf from the repo (but the last published version). https://github.com/google/protobuf.dart/actions/runs/4056670368/jobs/6981459199

I'll look up when support for that file was added. We can either not use the file (use dependency overrides in the regular pubspec file) or rev. our sdk min a bit higher.

devoncarew:
It looks like support for pubspec_overrides may have been added in 2.17.0. Thoughts on rev'ing the minimums to that version, at least for the protoc_plugin package? That's the current stable sdk (2.19) - 2.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ master ]
```
osa1, line 7: Doesn't this line cause publishing when a pull request is made to master branch? Without a tag pushed to the repo?

osa1 (commented):
Thanks!

> And fair warning, anyone with push access to this repo will be able to trigger a publish (it's possible to lock it down more than that - using github environments and some additional signoff on publish - but that's not enabled by default)

Hm, should we maybe announce somewhere (Dart Ecosystem channel?) that we use this GitHub action now and pushing a tag with `protobuf-v...` will automatically make a release? It's an unlikely event though.

.github/workflows/publish.yaml
devoncarew: Yeah, it's not straightforward to know what this will do just by looking at it; this series of lines will:
- execute the job below if a PR is opened which targets the master branch, or
- if a tag is pushed to the repo

If you dig into the script this references (https://github.com/dart-lang/ecosystem/blob/main/.github/workflows/publish.yaml), you can see a bit more detail. There, it runs our publishing tool in validation mode when its triggered from a PR, and in publish mode when its triggered from a push event.

devoncarew:
> Hm, should we maybe announce somewhere (Dart Ecosystem channel?) that we use this GitHub action now and pushing a tag with `protobuf-v...` will automatically make a release? It's an unlikely event though.

It is unlikely (and the tool will perform validation on the tag, so chances of something happening accidentally here are low). But certainly over communicating the workflow would be helpful - docs, chat, ... .

sigurdm (approved):
LGTM

PR #807, opened 2023-03-08
Sort message constructors first.

xclud:
Message constructors must be the first in the class.

According to:

https://dart-lang.github.io/linter/lints/sort_constructors_first.html

https://dart-lang.github.io/linter/lints/sort_unnamed_constructors_first.html

This leaves the generated code with less lint warnings.

### DIFF
--- a/protoc_plugin/test/goldens/grpc_service.pb
+++ b/protoc_plugin/test/goldens/grpc_service.pb
@@ -14,14 +14,14 @@ import 'dart:core' as $core;
 import 'package:protobuf/protobuf.dart' as $pb;
 
 class Empty extends $pb.GeneratedMessage {
+  factory Empty() => create();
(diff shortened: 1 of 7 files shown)


sigurdm:
Lints can be contradicting each other, so it is not possible to generate code that obeys all lints.

Instead it would probably be better to generate the new:
```
// ignore_for_file: type=lint
```

See:

https://github.com/dart-lang/site-www/issues/3576

xclud:
I understand, but this PR is an enhancement and better generated code.

kevmoo (approved):

kevmoo:
@sigurdm : it's not bad to try to align here, though. If they are simple fixes.

sigurdm:
Right - not against this fix at all. Feel free to merge!

PR #809, opened 2023-03-16
Format the generated code.

xclud:
This is equivalent to running `dart format .` in the directory after the code is generated.

Usually I get warnings from the generated code (Which can be turned of by the linter, i know). Also, If the code is uploaded as a package to pub.dev, It will lose 10 points for formatting.

> #### Pass static analysis
> code has no errors, warnings, lints, or formatting issues.

### DIFF
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -223,6 +223,13 @@ class FileGenerator extends ProtobufContainer {
   List<CodeGeneratorResponse_File> generateFiles(OutputConfiguration config) {
     if (!_linked) throw StateError('not linked');
 
+    String formatDartCode(String code) {
+      final formatter = DartFormatter();
+
+      final formattedCode = formatter.format(code);
+      return formattedCode;
+    }
+
     CodeGeneratorResponse_File makeFile(String extension, String content) {
       var protoUrl = Uri.file(descriptor.name);
       var dartUrl = config.outputPathFor(protoUrl, extension);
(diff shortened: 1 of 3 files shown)


protoc_plugin/pubspec.yaml
```diff
@@ -7,15 +7,16 @@ environment:
...
 dependencies:
+  dart_style: ^2.2.3
   fixnum: ^1.0.0
   path: ^1.8.0
   protobuf: ^2.0.0
 
 dev_dependencies:
   collection: ^1.15.0
-  lints: ^1.0.0
+  lints: ^2.0.0
```
kevmoo, line 17: revert these changes!

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -223,6 +223,13 @@ class FileGenerator extends ProtobufContainer {
   List<CodeGeneratorResponse_File> generateFiles(OutputConfiguration config) {
     if (!_linked) throw StateError('not linked');
 
+    String formatDartCode(String code) {
+      final formatter = DartFormatter();
```
kevmoo, line 227: maybe make the formater as instance var? So it's not created every call?

osa1 (commented):
We had discussed this PR before with @sigurdm. The problem here is that this formats based on the formatter used by the plugin, instead of the project's formatter, so `dart format` may format differently than `protoc_plugin`, and since you will be using `dart format` to format rest of your project anyway, you will be reformatting the generated files.

So I think this isn't solving any problems and no need to add a dependency to the project and make build times slower.

We also had a relevant PR in the past: https://github.com/google/protobuf.dart/pull/702. When the gprc support was added to this library (in the commit linked in the PR) formatting was added to the open source version, but not to the internal version. It wasn't documented in the commit (or the CL) why this was done differently but perhaps there are other reasons why we don't want to format, at least internally.

osa1:
Closing, for the reasons described above.

PR #824, opened 2023-04-22
Include the proto package name as a constant

### LINKED ISSUE #823: Include the package name as a constant
Hello,
Would it be possible to include the package name as a Dart constant? 

For a project I have a versioning system in the package name, so I parse a version number from the package and send it to a server. Currently I need to update a constant manually every time the protocol versioning changes, where I could simply use what gets autogenerated from the proto file.

davidmartos96:
Fixes #823 

This PR adds a constant with the package name at the top of the main .pb.dart file if the proto has a non empty package field.

```dart
const $core.String kPackageName = 'myPackage';
```

I opted to optionally include the constant based on the proto having a package defined or not. Do you think it should be included regardless as an empty String/null?

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.5.1
+# Created with package:mono_repo v6.5.4
 name: Dart CI
 on:
   push:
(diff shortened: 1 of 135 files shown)


davidmartos96:
I don't have the immediate need for this at the current time, but when I did, I didn't see any way to access the package name. How would that be using BuilderInfo?

osa1:
> How would that be using BuilderInfo?

I thought `BuilderInfo` exposes it, but apparently it doesn't, at least not directly.

You can get the qualified message name with `MyMessage.info_.qualifiedMessageName` and drop the parts after the last dot.

Something like (not tested):

```dart
final qualifiedName = MyMessage.info_.qualifiedMessageName;
final lastDotIndex = qualifiedName.lastIndexOf(".")
final packageName = qualifiedName.substring(0, lastDotIndex);
```

It's hacky but it should work.

Closing as this is no longer needed.

PR #816, opened 2023-04-13
Apply internally enabled lints

osa1:
This enables lints `prefer_final_in_for_each` and `prefer_final_locals` and fixes lints errors.

These lints were previously enabled in internal version in the following CLs:

- cl/523340795: updated protoc_plugin source
- cl/523340868: updated protoc_plugin tests
- cl/523340813: updated protobuf source
- cl/523952148: updated protobuf tests

Majority of the changes are generated by `dart fix --apply`. Manual changes were done in two places:

- `dart fix` does not fix lints in `<type> <var> = <expr>;` syntax, it looks like it only replaces `var` with `final` and does not add `final` before a type. A few lint errors were fixed manually for this.

- protoc_plugin/lib/src/client_generator.dart updated manually to fix the lints in generated code. The difference in generated code can be seen in changes in `protoc_plugin/test/goldens`.

PR #790, opened 2023-01-05
Support nullable types

johynpapin:
Hello everyone! :wave: 

The purpose of this PR is to support nullable types (#523). This allows to use generated messages in this way:

```dart
final user = User();
user.firstName = null;
print(user.firstName?.isNotEmpty());
```

The main advantage of this solution over hazzers is to take advantage of null safety.

Currently, this PR generates nullable types if the field is a message, or if the field is `optional` in the proto3 sense. Hazzers are still generated. To generate nullable types, you must use the `nullable` option, which is set to `false` by default.

I'm sure this PR is not mergeable in its current state, but I'm opening it to be helped and to ask questions. Don't hesitate to tell me if I'm going in the wrong direction!

- Why is `ArgumentError.checkNotNull(value, 'value')` used in `$_setFloat` for example, in `generated_message.dart`?
- Is it a good idea to modify `generated_message.dart` and `field_set.dart`?
- Is it a good idea to add an option to the generator?
- For tests, is it better to copy the current tests, modifying them to check for nullable types, or is it better to add new tests just for that?
- Is this feature desired in this package?

Thank you very much in advance!

PR #828, opened 2023-05-01
Update to latest lints and Dart SDK

kevmoo:


kevmoo:
@mraleph ?

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -396,7 +397,7 @@ class CodedBufferWriter {
     }
   }
 
-  void _writeBytesNoTag(dynamic value) {
+  void _writeBytesNoTag(Uint8List value) {
```
kevmoo, line 400: eliminated a dynamic invocation!

mraleph:
I will let @osa1 review once he has cycles.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -338,10 +338,10 @@ class CodedBufferWriter {
         break;
       case PbFieldType._BYTES_BIT:
         _writeBytesNoTag(
-            value is TypedData ? value : Uint8List.fromList(value));
+            value is Uint8List ? value : Uint8List.fromList(value));
```
osa1, line 341: I'm amazed that the linter can do this change, but I can't see the implications and whether this is safe (i.e. won't break user code). If I had to guess how the linter was able to do this change I would say it run type flow analysis and figured that the only `TypedData` that flows into this code is `Uint8List`, but I don't really know.

After these changes we could actually update `writeRawBytes` parameter too, but it's a public function (which is probably the reason why the linter didn't update it) and we have internal users so that would be a breaking change.

osa1 (commented):
Most of the changes LGTM, but I'm not sure if the changes in `coded_buffer_writer` are safe (added an inline comment).

Secondly, we should land #816 first or opt-out from internal lints in the internal version (which I think makes sense given the package is third-party) and revert the CLs listed in #816.

I have no cycles to keep these two versions in sync and this kind of style changes (either in the internal version or open source) only make it difficult for actual users to make changes. For example #825 is for an internal user and we have to do those changes manually in the internal version to be able to test and actually use it. #790 is another example of a useful change.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
kevmoo: This was a manual change on my part. Trying to get rid of the dynamic invoke.

kevmoo:
@osa1 – now ready for your review!

protoc_plugin/test/message_test.dart
```diff
@@ -173,26 +180,25 @@ repeatedString: qux
     message.repeatedMessage.add(testRequiredUninitialized);
     final List<int> buffer = message.writeToBuffer();
 
-    try {
-      (TestRequiredForeign.fromBuffer(buffer)).check();
-      fail('Should have thrown an exception.');
-    } on StateError catch (e) {
-      // NOTE: error message differs from Java in that
-      // fields are referenced using Dart fieldnames r.t.
-      // proto field names.
```
osa1, line 182: It looks like `dart fix` removed this comment.

protoc_plugin/test/message_test.dart
```diff
@@ -122,30 +126,29 @@ repeatedString: qux
...
-      final message = TestRequiredForeign();
-      message.optionalMessage = testRequiredUninitialized;
-      message.repeatedMessage.add(testRequiredUninitialized);
-      message.repeatedMessage.add(testRequiredUninitialized);
-      message.check();
-      fail('Should have thrown an exception.');
-    } on StateError catch (e) {
-      // NOTE: error message differs from Java in that
-      // fields are referenced using Dart fieldnames r.t.
-      // proto field names.
```
osa1, line 135: Same as below.

kevmoo:
@osa1 – PTAL

osa1 (approved):

PR #778, opened 2022-11-28
Emit imports in sorted order

devoncarew:
Part of https://github.com/google/protobuf.dart/issues/771 - ensure that the imports in the generated code are in lexical order.

- introduce an `ImportWriter` class
- use that when generating code
- have `ImportWriter` use regular dart import conventions (dart: first, package: second, path imports third, and lexically ordered within the groups)

Done as separate commits for easier review.

PR #833, opened 2023-05-26
Fix generated ignore_for_file directives

osa1:
- Bring unnecessary_import back. (was removed with #778)

- Generate unused_import in grpc files as well.

- Don't generate return_of_invalid_type: this lint seems to be removed.

(The first two issues were found in cl/535141580)

### DIFF
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -710,6 +710,7 @@ const _fileIgnores = {
   'library_prefixes',
   'non_constant_identifier_names',
   'prefer_final_fields',
+  'unnecessary_import',
   'unnecessary_this',
   'unused_import',
 };
--- a/protoc_plugin/lib/src/file_generator.dart
+++ b/protoc_plugin/lib/src/file_generator.dart
@@ -303,7 +303,7 @@ class FileGenerator extends ProtobufContainer {
   /// Writes the header and imports for the .pb.dart file.
   void writeMainHeader(IndentingWriter out,
       [OutputConfiguration config = const DefaultOutputConfiguration()]) {
-    _writeHeading(out, extraIgnores: {'unnecessary_import'});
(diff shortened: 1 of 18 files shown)


sigurdm (approved):
LGTM.

Should we move to suppressing all lints?:

```// ignore_for_file: type=lint```

 https://dart.dev/guides/language/analysis-options#suppressing-rules-for-a-file

osa1:
> Should we move to suppressing all lints?

That makes sense to me, I don't think linting generated code is useful. @devoncarew any thoughts on this?

devoncarew:
> > Should we move to suppressing all lints?
> 
> That makes sense to me, I don't think linting generated code is useful. @devoncarew any thoughts on this?

Sounds good to me. I think have generating code generally follow the dart style guide is a good practice, but don't have concerns w/ using `ignore_for_file: type=lint` - lint diagnostics in generated code can be challenging for client code.

osa1:
Re: `type=lint`, apparently it does not include some of the lints like `deprecated_member_use_from_same_package` with Dart 3.0.0. I wonder if this is a bug?

I also just realized that `unnecessary_import` and `unused_import` are two separate flags and we need both. The former when the imported stuff are also imported from other modules and I think it's triggered when an imported module is empty as well. Second is when an imported thing is not used.

osa1:
I reported the issue with `type=lint` in https://github.com/dart-lang/linter/issues/4402.

devoncarew:
> Re: type=lint, apparently it does not include some of the lints like deprecated_member_use_from_same_package with Dart 3.0.0. I wonder if this is a bug?

It likely just disables lints and not hints. Lints are things you can opt-into via an analysis options file; hints are shipped w/ the sdk as a part of normal analysis (along with warnings and errors).

You'll probably need `// ignore_for_file: type=lint` along with a list of a few specific hints that this package may trigger .

PR #836, opened 2023-06-06
Implement message set wire format

osa1:
Message set wire format is old and deprecated format used instead of extensions
in ancient protos.

This adds a new `GeneratedMessage` subclass `$_MessageSet` to override binary
serialization and deserialization methods for message sets. Message set
messages now extend `$_MessageSet` instead of `GeneratedMessage`.

The new class is prefixed as `$_` to make the addition backwards compatible.


protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -41,6 +41,8 @@ class BuilderInfo {
 
   List<FieldInfo>? _sortedByTag;
 
+  bool messageSetWireFormat;
```
osa1, line 44: I thought I would use this, but with a new class it's not necessary.

protobuf/lib/src/protobuf/message_set.dart
```diff
@@ -74,12 +85,17 @@ abstract class MessageSet extends GeneratedMessage {
...
     final ext = extensionRegistry.getExtension('MessageSet', typeId);
     if (ext == null) {
-      _fieldSet._ensureUnknownFields().addField(typeId, UnknownFieldSetField()..addLengthDelimited(message));
+      _fieldSet._ensureUnknownFields().addField(
+          typeId, UnknownFieldSetField()..addLengthDelimited(message));
     } else {
-      setExtension(ext, (ext.makeDefault!() as GeneratedMessage).deepCopy()..mergeFromBuffer(message));
+      setExtension(
+          ext,
+          (ext.makeDefault!() as GeneratedMessage).deepCopy()
```
sigurdm, line 97: Doesn't `makeDefault` already create a fresh message? Are we sure we have to do deepcopy here?

sigurdm: Ah - I think you want `subBuilder()` it should give you a fresh instance

protobuf/lib/src/protobuf/extension_registry.dart
```diff
@@ -109,14 +109,22 @@ T _reparseMessage<T extends GeneratedMessage>(
...
+    } else {
+      extensionRegistry._extensions[message.info_.qualifiedMessageName]
+          ?.forEach((tagNumber, extension) {
+        final unknownField = messageUnknownFields._fields[tagNumber];
+        if (unknownField != null) {
+          unknownField.writeTo(tagNumber, codedBufferWriter);
+          ensureUnknownFields()._fields.remove(tagNumber);
+        }
+      });
+    }
```
osa1, line 127: The special case for message sets above is inconsistent with the current handling of unknown fields. Currently if I have unknown fields with tags 1 and 2, and reparse with an extension for 1, the resulting message won't have 2 as an unknown field. The code for message sets OTOH serializes all extensions so 2 will be in unknown fields of the reparsed message.

protobuf/lib/src/protobuf/message_set.dart
```diff
@@ -0,0 +1,129 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../../protobuf.dart';
+
+const _messageSetItemsTag = 1;
+const _messageSetItemTypeIdTag = 2;
+const _messageSetItemMessageTag = 3;
+
+abstract class $_MessageSet extends GeneratedMessage {
```
osa1, line 11: We should add `/// nodoc` here otherwise it will appear in documentation.

protobuf/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
   [#763])
 * Fix handling negative JSON values when parsing uint32 fields. ([#839])
 * Avoid serializing unknown fields twice in `reparseMessage`. ([#840])
+* Implement message set wire format. ([#836])
```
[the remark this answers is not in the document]
sigurdm: Perhaps link to docs for this (or just write that this is mainly for supporting internal uses), as it is a very cryptic to outsiders.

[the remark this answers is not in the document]
osa1: I don't think we have public documentation for this. My understanding is it's an ancient proto1 compatibility format and not supposed to be used. I think it should be OK to skip the changelog entry.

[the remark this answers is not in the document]
sigurdm: Perhaps link to https://protobuf.dev/programming-guides/proto2/#options

protobuf/lib/src/protobuf/message_set.dart
sigurdm: Yes please

protoc_plugin/test/message_set_test.dart
```diff
@@ -0,0 +1,63 @@
...
+    0x1a,
+    0x06,
+    0x08,
+    0x7b,
+    0x12,
+    0x02,
+    0x68,
+    0x69,
+    0x0c
+  ];
```
sigurdm, line 25: Nit: You can format it nicer like so:
```suggestion
  final encoded = [
    0xda, 0x07, 0x0e, 0x0b, 0x10, //
    0xc8, 0xa6, 0x6b, 0x1a, 0x06,
    0x08, 0x7b, 0x12, 0x02, 0x68,
    0x69, 0x0c
  ];
```

protobuf/lib/src/protobuf/message_set.dart
sigurdm: Also please add a (short) DartDoc explaining what this is.

protobuf/lib/src/protobuf/message_set.dart
```diff
@@ -0,0 +1,129 @@
...
+        } else if (tagNumber == _messageSetItemMessageTag) {
+          message = input.readBytes();
+          if (typeId != null) {
+            _parseExtension(typeId, message, extensionRegistry);
+            typeId = null;
+            message = null;
+            continue outer;
+          }
+        } else {
+          throw UnsupportedError('Invalid message set item (tag = $tagNumber)');
```
sigurdm, line 94: I wonder if it is better to ignore the extra field here.

I wonder what other implementations do.

Probably OK to throw. We don't have anywhere to put unknown fields.

sigurdm (approved):
LGTM with a few nits.

protobuf/CHANGELOG.md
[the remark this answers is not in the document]
osa1: I removed the changelog entry. The documentation you linked clearly says it's for Google only.

protobuf/lib/src/protobuf/message_set.dart
osa1: Yeah we could ignore the field as the encoding is basically a repeated message encoding, and when decoding messages extra fields are always OK.

OTOH if there's an extra field here it probably means some kind of non-standard behavior or a bug in the program generating the message.

No strong opinions either way..

osa1:
One concern here is binary sizes and runtime performance when the new class is not used.

For binary sizes I checked a large internal dart2js app that currently does not use message sets. The binary size increased 0.007% so that should be OK. I don't know yet why the size increased though. Because the difference is so small I suspect it may just be some non-determinism in compilation.

For runtime performance I checked the benchmarks in the repo. The binaries before and after this PR are identical, so it seems like at least for small apps that don't use the feature there will be no difference in generated code and runtime performance.

PR #840, opened 2023-06-10
Avoid serializing unknown fields twice in reparseMessage

osa1:


sigurdm (approved):
LGTM

Do we have tests of unknown fields in reparseMessage?

sigurdm:
Good catch!

osa1:
> Do we have tests of unknown fields in reparseMessage?

The tests are here: https://github.com/google/protobuf.dart/blob/master/protoc_plugin/test/extension_test.dart I don't see any tests that check unknown fields after a `reparseMessage`.

PR #853, opened 2023-06-22
Fix unknown enum handling

### LINKED ISSUE #849: Repeated enum does not honor ignoreUnknownFields
Consider following snippet of code:

```
import 'dart:convert';

import 'response.pb.dart';

void main() {
  final response = Response()
    ..repeatedEnumList.addAll([EnumType.ENUM_TYPE_C, EnumType.ENUM_TYPE_B])
    ..singleEnumValue = EnumType.ENUM_TYPE_A;

  final json = response.toProto3Json() as Map<String, dynamic>;
  print(json);
  final serverAndClientInSync = Response.create()..mergeFromProto3Json(json);
 
  // WORKS
  final json2 = json;
  json2['singleEnumValue'] = 'ENUM_TYPE_CLIENT_UNAWARE';
  print(json2);
  final worksForUpgradedSingleField = Response.create()
    ..mergeFromProto3Json(json2, ignoreUnknownFields: true);

  // FAILS
  final json3 = json;
  json3['repeatedEnumList'][0] = 'ENUM_TYPE_CLIENT_UNAWARE';
  print(json3);
  final failsForUpgradedRepeatedEnumField = Response.create()
    ..mergeFromProto3Json(json3, ignoreUnknownFields: true);
}
```

#### Output
```
{repeatedEnumList: [ENUM_TYPE_C, ENUM_TYPE_B], singleEnumValue: ENUM_TYPE_A}
{repeatedEnumList: [ENUM_TYPE_C, ENUM_TYPE_B], singleEnumValue: ENUM_TYPE_CLIENT_UNAWARE}
{repeatedEnumList: [ENUM_TYPE_CLIENT_UNAWARE, ENUM_TYPE_B], singleEnumValue: ENUM_TYPE_CLIENT_UNAWARE}
Unhandled exception:
type 'Null' is not a subtype of type 'EnumType' of 'element'
#0      PbList.add (package:protobuf/src/protobuf/pb_list.dart)
#1      _mergeFromProto3Json.recursionHelper.<anonymous closure> (package:protobuf/src/protobuf/proto3_json.dart:385:24)
#2      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:625:13)
#3      _mergeFromProto3Json.recursionHelper (package:protobuf/src/protobuf/proto3_json.dart:338:14)
#4      _mergeFromProto3Json (package:protobuf/src/protobuf/proto3_json.dart:416:3)
#5      GeneratedMessage.mergeFromProto3Json (package:protobuf/src/protobuf/generated_message.dart:267:7)
#6      main (package:enums/enums.dart:26:7)
#7      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:296:19)
#8      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:189:12)
```

Full Code Snippet: https://github.com/Anushi1998/DartProtobufBug/

In case of single enum field, unknown value is ignored but for repeated enums it tried to add null. 
If we can avoid adding null [here](https://github.com/google/protobuf.dart/blob/master/protobuf/lib/src/protobuf/proto3_json.dart#L385) it will support backward compatability for repeated enums too. Happy to send a patch if this approach works.

osa1:
This fixes handling of unknown enum values in repeated and map fields and in
top-level fields.

Current behavior:

- Unknown enums in map and repeated fields throw an exception.
- Unknown enums in top-level fields reset the field value to the default one.

With this PR we just "ignore" the values:

- Unknown enums in map and repeated fields are skipped.
- Unknown enums in top-level fields do not reset the field's value to the
  default.

Fixes #849.

---

cl/796328163


sigurdm:
...Of course it is not trivial to use actual dart enums, because we want to store the integer value behind the "unknown" for re-serialization. I can only see how this could work by having a getter that translates the enum to a dart enum, but that is syntactically breaking.

We could also try to think in sealed classes - not sure how well that would work.

osa1:
It looks like resetting the top-level fields when the value is an unknown enum was deliberate as we have tests about it, but it doesn't make sense to me as resetting a field means we're not really *ignoring* the value (i.e. it has an effect on the message). As far as I can see proto3 spec doesn't mention this case, so I'll check other implementations and see what they do about this.

sigurdm:
Now I realize this PR is in terms of proto3 json decoding. Here we (usually) only have the name of the enum value, so we cannot translate it to the index...

Resetting vs skipping: 
* this only makes a difference when merging messages, right?
* both behaviors seem questionable under certain circumstances, but so does stuffing the enum **names** into the encoding - it seems to break some fundamental properties that are upheld by the binary protocol. Not much we can do about that.
* I guess following java is fine. I cannot remember why I implemented clearing the field.

osa1:
> this only makes a difference when merging messages, right?

If you mean as in `mergeFromMessage`, then no, this happens when merging a JSON representing a message. E.g.

```
myMessage.mergeFromProto3Json(..., ignoreUnknownFields: true)
```

`myMessage` can be an empty message (in which case skip vs. reset does not matter, though I should check if the current behavior also resets the hazzer) but it may also be a non-empty message with the enum field already set. In that case skipping keeps the current value (i.e. actually ignores the unknown value), resetting resets the value to the default (does not really ignore the unknown value).

> both behaviors seem questionable under certain circumstances, but so does stuffing the enum names into the encoding - it seems to break some fundamental properties that are upheld by the binary protocol. Not much we can do about that.

I think ignoring makes sense. Why do you say it's questionable? The flag says "ignore" so it should ignore.

I agree, proto3 JSON format is a bit strange..

sigurdm:
> myMessage can be an empty message (in which case skip vs. reset does not matter, ...

That's what I meant :) .

> I think ignoring makes sense. Why do you say it's questionable? The flag says "ignore" so it should ignore.

Well - one thing is what the flag says, another one is what is ergonomic for actual use. If I have one proto represented by
```
{
  "enumField": "A"
}
```
('A' is a known enum value).

And I merge it with another proto
```
{
  "enumField": "B"
}
```

If these were messages in the binary protocol, the unknown value would be preserved, and we would be able to see that from the merged proto.

I  find it somehow surprising that now `m.enumField` still returns `A`. It seems that the fact that the merged field carries a value, although unknown should be reflected in the merge. I cannot think of a use-case where that is the semantics I want.

But again, if that is what other implementations do, I guess we should just follow along.

osa1:
> I find it somehow surprising that now m.enumField still returns A

proto3 JSON (at least as we implement it in this library, AFAICS this is unspecified) does not support unknown fields, so the behavior you describe is not possible (you can't preserve the unknown value). You have these two options instead: (1) fail (the default) (2) ignore the unknown value (opt-in, with the flag). The default is (1). Binary format has (3) which is store the unknown field in unknown field set.

sigurdm:
Yeah - I argue that we get closer to option 3 by resetting.

But that is all splitting hairs. I don't hope anyone is actually using the proto3 json format for anything serious 🙈

sigurdm (approved):

PR #855, opened 2023-06-27
Generate constructor arguments again, add a flag to disable

### LINKED ISSUE #850: Add back named constructor arguments (or include as an option)
All the recent changes and compatibility with Dart 3 are phenomenal! So first of all thanks for that!

However, removing constructor arguments is an _incredibly_ harmful change for code quality. These constructors provide much needed structure and encapsulation. Given Dart's formatting, the cascading `Foo()..bar = 4` approach is unwieldily and leads to significantly less readable code. [The example](https://github.com/google/protobuf.dart/issues/155#issuecomment-504052280) given in the original issue illustrates this well.

In our case, this is a barrier to upgrading our proto libraries. We'd have to undergo a painful migration just to have worse code quality in the end. In the meantime, we'll likely either stay on the old version or fork the repo and revert [this change](https://github.com/google/protobuf.dart/pull/703/files) (which will not be fun either).

For us, the slightly increased binary size is a negligible side effect of having an easy way to construct messages. However, at the very least, I'd request making it configurable via an option (defaulting to `false` if you consider it prudent).

Thanks again for this awesome library! I think more people should be using it, and this would make it more developer friendly 🙂

osa1:
Fixes #850.

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,15 @@
+## 21.0.3
+
+* New protoc_plugin flag `generate_constructor_arguments` added to bring back
+  the old message factory methods with arguments. These arguments were removed
+  in 21.0.0 ([#703]) as they caused bloat in generated binaries in some cases.
+
+  Example usage:
+
+  ```
+  protoc --dart_out='generate_constructor_arguments,<other options>:.' ...
+  ```
+
 ## 21.0.2
 
 * Fix missing protobuf import in generated grpc files. ([#844])
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Generate code comments for annotated protobuf inputs. ([#161])
 * Generate message constructor arguments by default again. New flag
-  `disable_constructor_args` added to disable generating the arguments.
+  `disable_constructor_args` disables generating the arguments.
 
   Constructor arguments were removed in 21.0.0 as they increase dart2js binary
   sizes even when the arguments are not used.
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -2,14 +2,14 @@
 
 * Generate code comments for annotated protobuf inputs. ([#161])
 
-* New protoc_plugin flag `generate_constructor_arguments` added to bring back
-  the old message factory methods with arguments. These arguments were removed
-  in 21.0.0 ([#703]) as they caused bloat in generated binaries in some cases.
+* New protoc_plugin flag `constructor_args` added to bring back the old message
+  factory methods with arguments. These arguments were removed in 21.0.0
+  ([#703]) as they caused bloat in generated binaries in some cases.
 
   Example usage:
 
   ```
-  protoc --dart_out='generate_constructor_arguments,<other options>:.' ...
+  protoc --dart_out='constructor_args,<other options>:.' ...
   ```
 
 [#161]: https://github.com/google/protobuf.dart/issues/161
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,18 +1,23 @@
 ## 21.1.0-dev
 
 * Generate code comments for annotated protobuf inputs. ([#161])
+* Generate message constructor arguments by default again. New flag
+  `disable_constructor_args` added to disable generating the arguments.
 
-* New protoc_plugin flag `constructor_args` added to bring back the old message
-  factory methods with arguments. These arguments were removed in 21.0.0
-  ([#703]) as they caused bloat in generated binaries in some cases.
+  Constructor arguments were removed in 21.0.0 as they increase dart2js binary
+  sizes even when the arguments are not used.
 
-  Example usage:
+  Example usage to disable constructor arguments:
 
   ```
-  protoc --dart_out='constructor_args,<other options>:.' ...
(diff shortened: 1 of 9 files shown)


sigurdm (approved):
LGTM, but perhaps they should be turned on by default for non google3 use?

osa1:
> turned on by default for non google3 use

How would you implement this? By patching the library internally, or with an env variable, or ...?

sigurdm:
> How would you implement this? By patching the library internally, or with an env variable, or ...?

Maybe by passing the option to not create the named parameters from the blaze-rule that compiles the protos?

devoncarew:
Jumping in to plus one this PR - I did encounter the usability difference in creating proto objects in a project I'm working on (dartpad uses protos to communicate between the frontend and backend).

osa1:
@sigurdm PTAL.

protoc_plugin/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 * Generate code comments for annotated protobuf inputs. ([#161])
 
+* New protoc_plugin flag `generate_constructor_arguments` added to bring back
+  the old message factory methods with arguments. These arguments were removed
+  in 21.0.0 ([#703]) as they caused bloat in generated binaries in some cases.
+
+  Example usage:
+
+  ```
+  protoc --dart_out='generate_constructor_arguments,<other options>:.' ...
```
osa1, line 12: The flag is a bit long, I wonder if something shorter like `constructor_args` would be better?

sigurdm: Yeah, shorter is probably better here.

sigurdm:
Still LGTM.

I would still prefer for the constructor arguments to be generated by default. I think that is the most user friendly and what most people should use. Google3 (and others) are free to turn them off if they really think it saves them anything significant.

sigurdm (approved):

shilangyu:
As a non-google user of this library I agree that having the constructors generated by default is preferable. As I understand, the only reason one would not want to have constructors is to reduce the final binary size. I cannot speak for the whole community, but I believe this optimization is not something many people worry about. The PR which removed constructors mentions the increase in binary size but with no concrete numbers, perhaps the file size increase is indeed significant? I also fear that people will simply not know about the constructor option and new users will just use the cascade operator construction.

osa1:
> Google3 (and others) are free to turn them off if they really think it saves them anything significant.

@sigurdm I don't think this makes sense for internal users. Internal users don't explicitly update their dependencies, so they won't check the changelog and see if they need migration.

It's also not feasible to reach out to every single user, as there are just too many.

So projects with binary size benchmarks will potentially be notified, and then they have to figure out what changed and how to revert the changes. Projects without binary size benchmarks will silently get larger.

Also, no one "needs" this feature (it's not enabling anything new), and no internal user so far requested this feature even though it's been in the open source version for 2+ years (since Nov 2020).

So there seems to be no benefit for the internal users, but their projects may get slightly larger.

That said, I actually investigated a little bit why some projects get larger with these arguments. TFA is quite good at dropping unused arguments, and I've confirmed that VM and dart2wasm projects are not affected if they don't use the arguments. The problem is with dart2js projects, as dart2js doesn't use TFA and it doesn't seem to be dropping the arguments as aggressively.

Something else I've considered is to make this the default in open source, and just modify one line internally:

```diff
-      _generateFactory(out);
+      out.println('factory $classname() => create();');
```

But the problem is tests and golden files need to be different too, it's not as simple as just one line. This problem also exists if I somehow implement different defaults in internal and OS versions.

protoc_plugin/lib/src/message_generator.dart
```diff
@@ -430,6 +431,42 @@ class MessageGenerator extends ProtobufContainer {
...
+        if (field.isRepeated && !field.isMapField) {
+          out.println(
+              '  ${field.baseType.getRepeatedDartTypeIterable(fileGen)}? ${field.memberNames!.fieldName},');
+        } else {
+          out.println(
+              '  ${field.getDartType()}? ${field.memberNames!.fieldName},');
+        }
+      }
+      out.println('}) {');
+      out.println('  final result = create();');
```
devoncarew, line 448: Note that this local variable (`result`) can conflict with named params corresponding to a proto field.

For dartpad's protos code, we're seeing:

```
A value of type 'CompileResponse' can't be assigned to a variable of type 'String'.
Try changing the type of the variable, or casting the right-hand type to 'String'.
```

Perhaps name this `_result`? `$result`?

devoncarew: Our proto definition:

```
message CompileResponse {
  string result = 1;
  string sourceMap = 2;

  // Make this response compatible with BadRequest
  ErrorMessage error = 99;
}
```

devoncarew: https://github.com/google/protobuf.dart/pull/869

mikemoore13:
is it already included in a release ? how to check that ?

Rexios80:
This was included in version 21.1.0

https://pub.dev/packages/protoc_plugin/changelog#2110

mikemoore13:
i am running 
libprotoc 26.0-rc1 and it do not generate object with name constructors

Levi-Lesches:
What version of `protoc_plugin` are you using? Try re-installing it: 
```
dart pub global deactivate protoc_plugin
dart pub global activate protoc_plugin
```

mikemoore13:
Indeed the global protoc version I used was old. 
Something surprising is that all objects get generated with nullable fields.

PR #856, opened 2023-06-29
Ignore unknown tags in message set items

osa1:
Reference: https://github.com/protocolbuffers/protobuf/blob/a95eb31ac45eed38a586a11cfb9c93faf765bafa/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java#L850-L854


sigurdm (approved):
LGTM

PR #859, opened 2023-07-03
Add more message set tests, fix a bug

osa1:
When parsing an item we should consume the "end group" tag before breaking. Remove the `break` statements after parsing the extension.

### DIFF
--- a/protobuf/lib/src/protobuf/message_set.dart
+++ b/protobuf/lib/src/protobuf/message_set.dart
@@ -60,7 +60,7 @@ abstract class $_MessageSet extends GeneratedMessage {
       final tagNumber = getTagFieldNumber(tag);
 
       if (tag == 0) {
-        break;
+        break; // End of input.
       }
 
       if (tagNumber != _messageSetItemsTag) {
@@ -86,7 +86,7 @@ abstract class $_MessageSet extends GeneratedMessage {
         final tagNumber = getTagFieldNumber(tag);
 
         if (tag == 0) {
-          break;
+          break; // End of input.
         }
 
         if (tagNumber == _messageSetItemTypeIdTag) {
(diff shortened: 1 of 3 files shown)


protoc_plugin/test/message_set_test.dart
```diff
@@ -11,17 +11,27 @@ import '../out/protos/message_set.pb.dart';
 
 void main() {
   final encoded = [
```
osa1, line 13: I will document what this encodes.

protobuf/lib/src/protobuf/message_set.dart
```diff
@@ -96,20 +95,18 @@ abstract class $_MessageSet extends GeneratedMessage {
...
         } else if (tagNumber == _messageSetItemMessageTag) {
           message = input.readBytes();
           if (typeId != null) {
             _parseExtension(typeId, message, extensionRegistry);
             typeId = null;
             message = null;
-            continue outer;
           }
         } else {
           // Skip unknown tags.
```
mraleph, line 107: Please add a comment that this will consume end-group marker.

mraleph: You can actually add an assertion after this loop that the last consumed tag was the 1+EGROUP

osa1: I added the comment. We discussed the assertion offline, there are two ways that a message set message can end: with `EGROUP` (if it started with a `SGROUP`), or with end-of-input (when the message is at the top-level). So the assertion doesn't make sense (Java implementation [has this assertion][1] which I think is a bug).

I added a test to make sure we handle top-level message sets.

[1]: https://github.com/protocolbuffers/protobuf/blob/a95eb31ac45eed38a586a11cfb9c93faf765bafa/java/core/src/main/java/com/google/protobuf/GeneratedMessageLite.java#L856

mraleph (approved):
still LGTM

osa1:
I'll add a test for length delimited encoding of message sets. C++ library only uses groups but we should be able to generate length delimited encoding using other messages that look like the message set format.

Issue #161, opened 2018-10-05
Comments from .proto files should move over to dart


ened:
Given this enum defined in a proto file:

``` proto
// Defines the response status for a IAP operation.
enum IAPResponseStatus {
    // All went well, see the other fields for response details.
    ok = 0;

    // Fatal error during the API action.
    error = 1;

    // (iOS only) Library user needs to fetch the list of products from the store first.
    emptyProductList = 2;

    // In-App purchases not available (device or policy constraints etc).
    disabled = 3;
}
```

Then this plugin currently generates:

``` dart
class IAPResponseStatus extends $pb.ProtobufEnum {
  static const IAPResponseStatus ok = const IAPResponseStatus._(0, 'ok');
  static const IAPResponseStatus error = const IAPResponseStatus._(1, 'error');
  static const IAPResponseStatus emptyProductList = const IAPResponseStatus._(2, 'emptyProductList');
  static const IAPResponseStatus disabled = const IAPResponseStatus._(3, 'disabled');
}
```

However, it should generate:

``` dart
/// Defines the response status for a IAP operation.
class IAPResponseStatus extends $pb.ProtobufEnum {
  /// All went well, see the other fields for response details.
  static const IAPResponseStatus ok = const IAPResponseStatus._(0, 'ok');
  /// Fatal error during the API action.
  static const IAPResponseStatus error = const IAPResponseStatus._(1, 'error');
  /// (iOS only) Library user needs to fetch the list of products from the store first.
  static const IAPResponseStatus emptyProductList = const IAPResponseStatus._(2, 'emptyProductList');
  /// In-App purchases not available (device or policy constraints etc).
  static const IAPResponseStatus disabled = const IAPResponseStatus._(3, 'disabled');
}
```

v0l:
With gRPC  this would be very helpful to copy all the comments from rpc methods into the dart code!

armiol:
Is this still considered for implementation? We are struggling with the comments not being transferred to the Dart output as well.

sigurdm:
It would be awesome to support I agree. Currently we don't have a lot of resources, so don't expect it to happen soon.

MattyBoy4444:
Any news on this?

bartekpacia:
This would be a very useful feature.

[...]

PR #860, opened 2023-07-10
Compiler: generate doc comments

kevmoo:
Fixes https://github.com/google/protobuf.dart/issues/161

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 21.1.0-dev
 
-* Experimental support for generating comments. Use the `generate_comments` option. ([#161])
+* Generate code comments for annotated protobuf inputs. ([#161])
 
 [#161]: https://github.com/google/protobuf.dart/issues/161
 
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 21.1.0-wip
+
+* Experimental support for generating comments. Use the `generate_comments` option. ([#161])
+
+[#161]: https://github.com/google/protobuf.dart/issues/161
+
 ## 21.0.2
 
 * Fix missing protobuf import in generated grpc files. ([#844])
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 21.1.0-wip
+## 21.1.0-dev
 
 * Experimental support for generating comments. Use the `generate_comments` option. ([#161])
 
--- a/protoc_plugin/lib/src/shared.dart
+++ b/protoc_plugin/lib/src/shared.dart
@@ -38,6 +38,8 @@ String? toDartComment(String value) {
 
   var lines = LineSplitter.split(value).toList();
 
+  // Find any leading spaces in the first line.
+  // If all of the lines have the same leading spaces, remove them all.
   final leadingSpaces = _leadingSpaces.firstMatch(lines.first);
   if (leadingSpaces != null) {
     final prefix = leadingSpaces.group(0)!;
@@ -46,6 +48,7 @@ String? toDartComment(String value) {
     }
   }
 
+  // Remove empty, trailing lines
   while (lines.last.isEmpty) {
     lines.removeLast();
   }
--- a/protoc_plugin/lib/src/shared.dart
+++ b/protoc_plugin/lib/src/shared.dart
@@ -24,9 +24,7 @@ extension FileDescriptorProtoExt on FileGenerator {
 
     if (bits.length == 1) {
       final match = bits.single;
-      if (match.hasLeadingComments()) {
-        return toDartComment(match.leadingComments);
-      }
+      return toDartComment(match.leadingComments);
     }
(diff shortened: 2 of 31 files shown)


protoc_plugin/lib/src/shared.dart
```diff
@@ -2,8 +2,61 @@
...
+
+    final bits = descriptor.sourceCodeInfo.location
+        .where((element) => element.path.toString() == path.toString())
+        .toList();
+
+    if (bits.length == 1) {
+      final match = bits.single;
+      if (match.hasLeadingComments()) {
+        return toDartComment(match.leadingComments);
+      }
```
osa1, line 29:
```suggestion
      return toDartComment(match.leadingComments);
```

When the field `leadingComments` isn't present the field value will be the default value of the field which is empty string, and `toDartComment` handles empty strings. So no need to check for presence.

protoc_plugin/lib/src/shared.dart
```diff
@@ -2,8 +2,61 @@
...
+    }
+
+    if (bits.length > 1) {
+      throw 'WTH? ${descriptor.sourceCodeInfo.toProto3Json()}';
```
osa1, line 33: I wonder if we should just print a warning (to stderr, stdout is used to print the generated code) and continue. `SourceCodeInfo` stuff does not affect runtime behavior of the protos so maybe a malformed `SourceCodeInfo` shouldn't block the operation.

We should also rephrase this with something helpful. Maybe something like "proto definition at $span has multiple source code infos".

protoc_plugin/lib/src/shared.dart
```diff
@@ -2,8 +2,61 @@
...
+      lines.removeLast();
+    }
+
+    return lines.map((e) => '/// $e'.trimRight()).join('\n');
+  }
+
+  return null;
+}
+
+final _leadingSpaces = RegExp('^ +');
```
osa1, line 62: What about other whitespace characters? https://protobuf.dev/reference/protobuf/textformat-spec/#whitespace

I really dislike regexes for simple things like this, but maybe this is the easiest in Dart..

protoc_plugin/lib/src/shared.dart
```diff
@@ -2,8 +2,61 @@
...
+    }
+
+    while (lines.last.isEmpty) {
+      lines.removeLast();
+    }
+
+    return lines.map((e) => '/// $e'.trimRight()).join('\n');
+  }
+
+  return null;
```
osa1, line 59: If you move this check at the beginning with

```dart
if (value.isEmpty) return null;
```

Then the function will need less indentation and be easier to read. Same as in `commentBlock` where you check for the early return first.

protoc_plugin/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: protoc_plugin
-version: 21.0.2
+version: 21.1.0-wip
```
osa1, line 2:
```suggestion
version: 21.1.0-dev
```

The convention we use in protoc_plugin and protobuf is `-dev`. protobuf lib version currently is `3.0.1-dev`.

osa1 (approved):
Looks great, thanks @kevmoo. Added a few inline comments and suggestions.

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 21.1.0-wip
```
osa1, line 1:
```suggestion
## 21.1.0-dev
```

(Same as the other suggestion)

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+
+* Experimental support for generating comments. Use the `generate_comments` option. ([#161])
```
osa1, line 3: Why do you say it's experimental? If there isn't anything unclear (maybe regarding how the `SourceCodeInfo` etc. are used) or unsupported I think we should just advertise this as "support".

kevmoo: I'm just not very committal. We can call it "ready"

In fact, happy to discuss not putting it behind a flag.

osa1: I checked a few of the protoc compilers used internally. Java and Go generate comments. C++ doesn't. AFAIK this library mostly copies Java (the lite version) so I say we generate comments always and remove the flag.

protoc_plugin/lib/src/shared.dart
kevmoo: yup. fixed

protoc_plugin/lib/src/shared.dart
osa1: Yeah tabs should be common. I don't even know what vertical tab and form feed are and I suspect they're not used in practice. Carriage return should be handled by the line splitter.

kevmoo: I think it's okay to handle these case-by-case. In all of the bits I looked through they ALWAYS seem to have one leading space – that's it!

MattyBoy4444:
This doesn't grab comments at the end of the line.

EX: 

message Animal {
  string name = 1; //This is the name of the animal
}

kevmoo:
File another bug on that

PR #863, opened 2023-07-29
Avoid holding onto the buffer when parsing unknown length-delimited fields

osa1:
Currently unknown field set parser stores length-delimited fields as views of the input buffer. This has a few issues:

- It's inconsistent with the known-field parsing where we copy `bytes` fields.

- A single unknown length-delimited field can cause keeping a large input buffer alive.

- Because the caller is free to modify the input buffer, this implementation does not allow freezing an unknown field set.

  (This can also be fixed by copying the length-delimited fields when freezing.)

- Even when the parsed message is not frozen, this aliasing can cause bugs as it's not documented. Even if we document it, it would probably be a footgun.

  This can cause segfaults or worse when the input buffer is passed from e.g. C++ or C as a `Uint8List`, and the caller frees the buffer after parsing while the message is still in use.

This PR makes `readBytes` copy the bytes before returning to avoid aliasing. A new member `readBytesAsView` added with the previous behavior. Unknown length-delimited field parsing fixed with the new `readBytes`.

Sync CL: cl/552077275

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 3.0.1-dev
 
+* Avoid holding onto the input buffer when parsing unknown length-delimited
+  fields. ([#863])
+
+[#863]: https://github.com/google/protobuf.dart/pull/863
+
 ## 3.0.0
 
 * Require Dart `2.19`.
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,8 +1,11 @@
-## 4.0.0-dev
+## 3.1.0-dev
 
-* Avoid holding onto the input buffer when parsing unknown length-delimited
-  fields. `CodedBufferReader` `readBytes` is renamed as `readBytesAsView`.
-  ([#863])
+* `CodedBufferReader` `readBytes` now copies the returned bytes to avoid
+  accidental sharing of the input buffer with the returned `Uint8List`. New
+  member `readBytesAsView` added with the old behavior. ([#863])
+
+* Avoid sharing the input buffer in unknown length-delimited fields using the
+  new `readBytes`. ([#863])
 
 [#863]: https://github.com/google/protobuf.dart/pull/863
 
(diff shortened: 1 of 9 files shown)


protobuf/lib/src/protobuf/coded_buffer_reader.dart
```diff
@@ -133,7 +133,13 @@ class CodedBufferReader {
...
 
   bool readBool() => _readRawVarint32(true) != 0;
-  List<int> readBytes() {
+
+  /// Read a length-delimited field as bytes. Note that the returned
+  /// [Uint8List] will be a view of the [CodedBufferReader]'s buffer. When
+  /// storing the returned value directly (instead of e.g. parsing it as a
+  /// UTF-8 string and copying) make sure to copy it to avoid holding on to the
+  /// whole message.
+  Uint8List readBytes() {
```
mkustermann, line 142: Consider renaming this to something that indicates to the caller it'S only a view, e.g. `readBytesAsView()`

protobuf/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## 3.0.1-dev
 
+* Avoid holding onto the input buffer when parsing unknown length-delimited
```
mkustermann, line 3: Would it make sense to publish it as `3.0.1` or `3.0.0+1`?

mkustermann (approved):
Thanks, @osa1

protobuf/lib/src/protobuf/coded_buffer_reader.dart
osa1: I think it makes sense, but it also makes this a breaking change as `CodedBufferReader` is a public type. We should still do it though, I'll update.

protobuf/CHANGELOG.md
osa1: It makes sense to publish this, yes. With the renaming we'll need a major version bump.

osa1: It makes sense to publish this, yes. With the renaming we'll need a major version bump.

PR #869, opened 2023-08-15
Rename a local variable to avoid proto field name conflicts

devoncarew:
- rename a local variable to avoid proto field name conflicts

This fixes an issue we're seeing with the latest version of `package:protoc_plugin` and the dartpad proto definitions.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,6 +6,7 @@ on:
   # Trigger on pull requests that target the default branch.
   pull_request:
     branches: [ master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   # Also, trigger when tags are pushed to the repo.
   push:
     tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 21.1.1
+
+* Rename a local variable used with message constructors to avoid potential
+  conflicts with protobuf field names.
+
 ## 21.1.0
 
 * Generate code comments for annotated protobuf inputs. ([#161])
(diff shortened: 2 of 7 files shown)


devoncarew:
It looks like the publishing warnings will happen for all PRs here; we might want a way to disable publishing warnings via a config file.

```
Package validation found the following potential issues:
  * `dart analyze` found the following issue(s):
    Analyzing lib, test, bin, pubspec.yaml...
    
      error - test/any_test.dart:8:8 - Target of URI hasn't been generated: '../out/protos/google/protobuf/any.pb.dart'. Try running the generator that will generate the file referenced by the URI. - uri_has_not_been_generated
      error - test/any_test.dart:9:8 - Target of URI hasn't been generated: '../out/protos/service.pb.dart'. Try running the generator that will generate the file referenced by the URI. - uri_has_not_been_generated
      error - test/any_te[...]he override annotation. - override_on_non_overriding_member
       info - test/extension_test.dart:24:7 - Missing type annotation on a public API. Try adding a type annotation. - type_annotate_public_apis
       info - test/map_field_test.dart:279:12 - Closure should be a tearoff. Try using a tearoff rather than a closure. - unnecessary_lambdas
       info - test/map_field_test.dart:353:5 - Unnecessary statement. Try completing the statement or breaking it up. - unnecessary_statements
    
    2287 issues found.
  * 1 checked-in file is ignored by a `.gitignore`.
    Previous versions of Pub would include those in the published package.
    
    Consider adjusting your `.gitignore` files to not ignore those files, and if you do not wish to
    publish these files use `.pubignore`. See also dart.dev/go/pubignore
    
    Files that are checked in while gitignored:
    
    test/goldens/serviceGenerator.pb.json  
    
  Package validation found the following hint:
  * Non-dev dependencies are overridden in pubspec_overrides.yaml.
    
    This indicates you are not testing your package against the same versions of its
    dependencies that users will have when they use it.
    
    This might be necessary for packages with cyclic dependencies.
    
    Please be extra careful when publishing.
  
  Package has 2 warnings and 1 hint.
```

.github/workflows/publish.yaml
```diff
@@ -6,6 +6,7 @@ on:
   # Trigger on pull requests that target the default branch.
   pull_request:
     branches: [ master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
devoncarew, line 9: This will re-run the publishing validation when a PR's labels change.

devoncarew:
Note that I'll be OOO for the rest of the week; please feel free to edit or form this PR as necessary.

osa1 (approved):
Thanks!

PR #880, opened 2023-10-10
Remove event support

### LINKED ISSUE #738: Remove event mixin stuff
`PbEventMixin,` `PbFieldChange`, and `EventBuffer` are currently unused. No need to maintain unused code, we should just remove them.

They're just a few lines of code and implemented as a separate library (instead of as a part of the protobuf library), so users can implement that stuff themselves if needed.

osa1:
Event support is currently unused internally and we don't want to introduce new
uses of it, to keep the API surface small and the library easier to update, and
to make it easier to migrate to another library.

`createRepeatedField` and `createMapField` existed to support events (as
`PbList` and `PbMap` don't support this), so with the event mixin removed we
remove these members as well.

Removing these members give us other opportunities: we now have full control
over the field value types. This allows, for example, refactoring `PbMap` and
`PbList` types for marking them as frozen without visiting the elements, which
makes it possible to implement decoders that create frozen objects without
having to make another pass after decoding to mark every object as frozen.

Closes #738.

cl/571893384


protoc_plugin/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
-## 21.1.2-dev
+## 22.0.0-dev
 
-* Fix a bug in comment parsing ([#871], [#879]).
+* Fix a bug in comment parsing. ([#871], [#879])
```
osa1, line 3: Note: @sigurdm I'll release this bug fix in a minor release.

protobuf/lib/meta.dart
```diff
@@ -21,9 +21,6 @@ const GeneratedMessage_reservedNames = <String>[
   'clone',
   'copyWith',
   'createEmptyInstance',
-  'createMapField',
```
sigurdm, line 24: Perhaps we should leave these? Otherwise the names will no longer be disambiguated breaking existing code?
Not sure what is better here...

sigurdm (approved):

protobuf/lib/meta.dart
osa1: Do you suggest we keep them indefinitely, or remove later?

If you want to remove later, this is already a breaking change so we could do it now.

sigurdm: Yeah - it is either remove 

* now or 
* never.

Not sure what is better. We could check if we have any usages inside google3. These names are probably quite rare - so maybe it is better to remove now.

osa1: I tested the CL globally internally it doesn't break anything.

sigurdm: Ok, in that case I vote for just removing them

PR #882, opened 2023-10-12
More GeneratedMessage docs, hide meta library

osa1:


protobuf/lib/src/protobuf/generated_message.dart
```diff
@@ -162,15 +167,18 @@ abstract class GeneratedMessage {
     }
   }
 
+  /// Serialize the message.
```
sigurdm, line 170: Should we mention the serialization format (i.e. binary protobuf encoding)?

sigurdm (approved):

protobuf/lib/src/protobuf/generated_message.dart
osa1: @sigurdm PTAL.

sigurdm (approved):

PR #885, opened 2023-10-23
Optimize `CodedBufferWriter._copyInto` to memcpy

osa1:
dart2wasm currently can't optimize loops into memcpy, however `setRange` methods have type tests to generate `array.copy` (Wasm's `memcpy`).

Replacing the loops in `CodedBufferWriter._copyInto` with `setRange` improves an internal benchmark extracted from a real use case significantly in all targets:

| AOT                          | 127,587 us | 95,634 us  | -31,953 us,  -25.0% |
| JIT                          | 106,880 us | 92,800 us  | -14,080 us,  -13.1% |
| dart2js -O4                  | 285,587 us | 262,222 us | -23,365 us,  -8.1%  |
| dart2wasm --omit-type-checks | 337,000 us | 236,100 us | -100,900 us, -29.9% |

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -421,22 +421,19 @@ class CodedBufferWriter {
 
   /// Copy bytes from the given typed data array into the output buffer.
   ///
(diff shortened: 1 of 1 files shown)


mkustermann (approved):
Nice \o/

PR #886, opened 2023-10-24
Avoid adding empty splices in `CodedBufferWriter`

osa1:
By avoiding allocating empty `Uint8List`s this saves a few percent in dart2js -O4 in the benchmark reported in #885.

| AOT                          | 122,917 us | 122,741 us |                     |
| JIT                          |  93,376 us |  94,880 us |                     |
| dart2js -O4                  | 271,111 us | 258,250 us | -12,861 us, -4.7%   |
| dart2wasm --omit-type-checks | 196,454 us | 195,300 us |                     |

Number of splices in the benchmark before this change: 21,123
After: 20,857

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -337,11 +337,22 @@ class CodedBufferWriter {
         _writeVarint32(value ? 1 : 0);
         break;
       case PbFieldType._BYTES_BIT:
-        _writeBytesNoTag(
-            value is Uint8List ? value : Uint8List.fromList(value));
+        final List<int> bytes = value;
+        if (bytes is Uint8List) {
+          _writeBytesNoTag(bytes);
+        } else if (bytes.isEmpty) {
+          writeInt32NoTag(0);
+        } else {
+          _writeBytesNoTag(Uint8List.fromList(value));
+        }
         break;
       case PbFieldType._STRING_BIT:
-        _writeBytesNoTag(_utf8.encoder.convert(value));
+        final String string = value;
+        if (string.isEmpty) {
+          writeInt32NoTag(0);
+        } else {
+          _writeBytesNoTag(_utf8.encoder.convert(string));
+        }
         break;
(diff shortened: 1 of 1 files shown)


protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -337,11 +337,22 @@ class CodedBufferWriter {
         _writeVarint32(value ? 1 : 0);
         break;
       case PbFieldType._BYTES_BIT:
-        _writeBytesNoTag(
-            value is Uint8List ? value : Uint8List.fromList(value));
+        final List<int> bytes = value;
+        if (bytes is Uint8List) {
+          _writeBytesNoTag(bytes);
```
mkustermann, line 342: If `bytes.length == 0` we still add a splice, etc. Consider adding a `bytes.length == 0` case here inside the `if (is Uint8List)` (as that `.length` load will be inlined if we know it's a `Uint8List`)

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -337,11 +337,22 @@ class CodedBufferWriter {
...
+        } else if (bytes.isEmpty) {
+          writeInt32NoTag(0);
```
mkustermann, line 344: Consider introducing a `_writeEmptyBytes()` helper and use it here and other places where we basically intend to write an empty byte array.

```
@pramga('vm:prefer-inline')
@...
void _writeEmptyBytes() {
  writeInt32NoTag(0);
}
```

mkustermann (approved):

osa1:
I will get back to this after cl/576427817.

protobuf/lib/src/protobuf/coded_buffer_writer.dart
osa1: I added the helper but I didn't add the pragmas. We have a bunch of other one-liners in this file, I'd like to experiment with inlining them manually separately.

mkustermann (approved):

PR #887, opened 2023-10-24
Use `setRange` when copying output chunks to the final buffer in `CodedBufferWriter`

osa1:
Similar to #885, this optimizes some more buffer copying to `memcpy`.

Results from the same benchmark in #885:

| AOT                          | 114,713 us | 109,838 us | - 4,875 us, -4.2%   |
| JIT                          |  91,960 us |  92,887 us | +   927 us, +1.0%   |
| dart2js -O4                  | 259,125 us | 257,000 us | - 2,125 us, -0.8%   |
| dart2wasm --omit-type-checks | 196,909 us | 182,333 us | -14,576 us, -7.4%   |

AOT and JIT tested on x64.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -152,21 +152,11 @@ class CodedBufferWriter {
             final leftInChunk = bytesInChunk - chunkPos;
             final bytesToCopyFromChunk =
                 leftInChunk > bytesToCopy ? bytesToCopy : leftInChunk;
-            final endPos = chunkPos + bytesToCopyFromChunk;
-
-            if (bytesToCopyFromChunk <= 20) {
-              while (chunkPos < endPos) {
-                buffer[outPos++] = chunk[chunkPos++];
-              }
-              bytesToCopy -= bytesToCopyFromChunk;
-            } else {
-              final chunkSlice = Uint8List.sublistView(chunk, chunkPos, endPos);
-              buffer.setRange(
-                  outPos, outPos + bytesToCopyFromChunk, chunkSlice);
-              chunkPos += bytesToCopyFromChunk;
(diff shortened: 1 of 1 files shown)


osa1:
@rakudrama any suggestions on how to improve the dart2js performance here?

protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -141,22 +141,32 @@ class CodedBufferWriter {
...
-            // Copy at most bytesToCopy bytes from the current chunk.
+            // Copy at most `bytesToCopy` bytes from the current chunk.
             final leftInChunk = bytesInChunk - chunkPos;
             final bytesToCopyFromChunk =
                 leftInChunk > bytesToCopy ? bytesToCopy : leftInChunk;
             final endPos = chunkPos + bytesToCopyFromChunk;
-            while (chunkPos < endPos) {
-              buffer[outPos++] = chunk[chunkPos++];
+
+            if (bytesToCopyFromChunk <= 20) {
```
mkustermann, line 157: The "optimization" in the `if` branch is probably due to the `else` branch doing the view object allocation, is it?

Our `Uint8List.setRange()` API actually takes an optional `skipCount` parameter that says how many bytes to skip in the `from`. So why not do this:
```dart
buffer.setRange(outPos, outPos + bytesToCopyFromChunk, chunk, chunkPos);
chunkPos += bytesToCopyFromChunk;
outPos += bytesToCopyFromChunk;
bytesToCopy -= bytesToCopyFromChunk;
```
instead of the `if { manual loop } else { allocate view + setRange }`?

osa1: Good idea, thanks. PTAL. I've also updated the benchmark results.

mkustermann (approved):

PR #888, opened 2023-10-25
Avoid allocating temporary views in when decoding strings in `CodedBufferReader`

osa1:
Benchmark results from the same internal benchmark reported in previous PRs/commits:

|                              | Before     | After      | Diff                |
|------------------------------|------------|------------|---------------------|
| AOT                          |  31,025 us |  27,597 us | - 3,428 us, -11.0%  |
| JIT                          |  34,829 us |  26,846 us | - 7,983 us, -22.9%  |
| dart2js -O4                  | 300,571 us | 248,300 us | -52,271 us, -17.3%  |
| dart2wasm --omit-type-checks | 130,812 us | 107,850 us | -22,962 us, -17.5%  |

PR #890, opened 2023-10-25
Keep a `ByteData` around in `CodedBufferReader` to avoid repeated `ByteData` allocs

osa1:
In Dart, the only way to convert a list of bytes to an `int` or `double` is by using a `ByteData`.

Currently `CodedBufferReader` allocates a new `ByteData` for every fixed-size `int` or `float` decoding.

With this change we now allocate a `ByteData` once initialization and reuse it.

Benchmark results, using the same benchmark reported in #888:

| AOT                          |  26,763 us |  25,305 us | - 1,458 us, - 5.4%  |
| JIT                          |  24,573 us |  25,521 us | +   948 us, + 3.8%  |
| dart2js -O4                  | 212,000 us | 199,300 us | -12,700 us, - 5.9%  |
| dart2wasm --omit-type-checks |  89,750 us |  78,958 us | -10,792 us, -12.0%  |

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_reader.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_reader.dart
@@ -13,6 +13,11 @@ class CodedBufferReader {
   static const int DEFAULT_SIZE_LIMIT = 64 << 20;
 
   final Uint8List _buffer;
+
+  /// [ByteData] of [_buffer], created once to be able to decode fixed-size
+  /// integers and floats without having to allocate a [ByteData] every time.
+  final ByteData _byteData;
+
   int _bufferPos = 0;
   int _currentLimit = -1;
   int _lastTag = 0;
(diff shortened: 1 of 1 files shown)


protobuf/lib/src/protobuf/coded_buffer_reader.dart
```diff
@@ -13,17 +13,30 @@ class CodedBufferReader {
...
+          int sizeLimit = DEFAULT_SIZE_LIMIT}) =>
+      CodedBufferReader._(
+        buffer is Uint8List ? buffer : Uint8List.fromList(buffer),
+        recursionLimit,
+        sizeLimit,
+      );
+
+  CodedBufferReader._(Uint8List buffer, int recursionLimit, int sizeLimit)
+      : _buffer = buffer,
+        _byteData = ByteData.sublistView(buffer),
```
mkustermann, line 39: With this we're paying for a `ByteData` allocation even if we don't need it when decoding a proto. There may be many protos that don't have float/double/fixed-ints in them.

Consider doing that lazily:
```
late final ByteData _byteData = ByteData.sublistView(_buffer);
```

(then you may also undo the changes to the constructor)

mkustermann (approved):
LGTM with comments

protobuf/lib/src/protobuf/coded_buffer_reader.dart
osa1: Right, but it's one allocation per top-level `fromBuffer` call, nested messages are decoded using the same `CodedBufferReader`.

I'm not sure how expensive the allocation is, but it will be optimizing small messages at the cost of large ones with float etc. fields.

Do you think it's the right trade-off?

mkustermann: I was more thinking about dart2js. We know that JS typed data allocations are expensive. Maybe typed data view (like `ByteData`) allocations are not? In that case, feel free to leave like it is.

osa1: I implemented your suggestion. It didn't make significant different in my benchmarks.

PR #891, opened 2023-10-25
Make `CodedBufferWriter.writeRawByte` argument type more accurate

osa1:
`writeRawBytes` adds bytes to the output and it will only be fast when the argument is `Uint8List`.

Since all the users already pass `Uint8List`, update the argument type.

This allows simplifyping `_copyInto`. Because it has only one call site, we simplify it and inline it in the call site.

cl/576427817

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer_writer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer_writer.dart
@@ -22,8 +22,8 @@ part of '../../protobuf.dart';
 class CodedBufferWriter {
   /// Array of splices representing the data written into the writer.
   /// Each element might be one of:
-  ///   * a TypedData object - represents a sequence of bytes that need to be
-  ///                          emitted into the result as-is;
+  ///   * a [Uint8List] object - represents a sequence of bytes that need to be
+  ///                            emitted into the result as-is;
   ///   * a positive integer - a number of bytes to copy from [_outputChunks]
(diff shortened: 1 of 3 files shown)


protobuf/lib/src/protobuf/coded_buffer_writer.dart
```diff
@@ -220,9 +223,9 @@ class CodedBufferWriter {
     _lastSplicePos = pos;
   }
 
-  /// Add TypedData splice - these bytes would be directly copied into the
-  /// output buffer by [writeTo].
-  void writeRawBytes(TypedData value) {
+  /// Add a [Uint8List] splice, without copying. These bytes will be directly
+  /// copied into the output buffer by [writeTo].
+  void writeRawBytes(Uint8List value) {
```
mkustermann, line 228: Currently all callers already have a `Uint8List` so this is fine.

But if there were some callers that may need to explicitly create views due to this API, we couldn've made this `void writeRawBytes(Uint8List value, int offset, int length)` (analogous to the `.setRange()` API)

mkustermann (approved):

protobuf/lib/src/protobuf/coded_buffer_writer.dart
osa1: Good idea. As you say, currently users always pass a whole `Uint8List` but we can add optional positionals if needed.

PR #902, opened 2023-11-20
Make wrapped lists in PbList monomorphic

osa1:
In `PbList`, the list field becomes monomorphic growable list (from the list base class). This makes `add` calls monomorphic and inlinable, and avoids double mutability checks (once in `PbList.add`, again in `_wrappedList.add`).

Also simplifies immutable `PbMap` allocations.

`PbMap._isReadonly` is renamed to `_isReadOnly` for consistency with `PbList._isReadOnly`.

### DIFF
--- a/protobuf/lib/src/protobuf/pb_list.dart
+++ b/protobuf/lib/src/protobuf/pb_list.dart
@@ -11,7 +11,14 @@ typedef CheckFunc<E> = void Function(E? x);
 
 /// A [ListBase] implementation used for protobuf `repeated` fields.
 class PbList<E> extends ListBase<E> {
+  /// The actual list storing the elements.
+  ///
+  /// We want only one [List] implementation class to be stored here to make
+  /// sure the list operations are monomorphic and can be inlined. To make this
+  /// explicit we use the `<E>[]` syntax when initializing this field (instead
+  /// of factory methods like `List.from`).
   final List<E> _wrappedList;
+
   final CheckFunc<E> _check;
 
   bool _isReadOnly = false;
@@ -23,12 +30,14 @@ class PbList<E> extends ListBase<E> {
         _check = check;
 
   PbList.unmodifiable()
-      : _wrappedList = const [],
+      : _wrappedList = <E>[],
         _check = _checkNotNull,
         _isReadOnly = true;
 
-  PbList.from(List from)
-      : _wrappedList = List<E>.from(from),
+  PbList.from(List<E> from)
+      // Use `<E>[]` to allocate the list, see [_wrappedList] documentation.
+      // ignore: prefer_spread_collections
+      : _wrappedList = <E>[]..addAll(from),
         _check = _checkNotNull;
 
   @override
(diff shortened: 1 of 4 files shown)


protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -23,12 +30,14 @@ class PbList<E> extends ListBase<E> {
...
-      : _wrappedList = const [],
+      : _wrappedList = <E>[],
         _check = _checkNotNull,
         _isReadOnly = true;
 
-  PbList.from(List from)
-      : _wrappedList = List<E>.from(from),
+  PbList.from(List<E> from)
+      // Use `<E>[]` to allocate the list, see [_wrappedList] documentation.
+      // ignore: prefer_spread_collections
```
mkustermann, line 39: It may be that the spread operator gets more optimization than doing follow-up method calls.

Why not make this `_wrappedList = <E>[...from]`?

(similar below

mkustermann (approved):

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -23,12 +30,14 @@ class PbList<E> extends ListBase<E> {
         _check = check;
 
   PbList.unmodifiable()
-      : _wrappedList = const [],
+      : _wrappedList = <E>[],
```
mkustermann, line 33: Consider
```
static _emptyList = <Never>[];
```
and use it here, to avoid repeated allocation.

mkustermann (approved):

osa1:
@mkustermann I mostly rewrote this PR, ptal.

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -11,7 +11,21 @@ typedef CheckFunc<E> = void Function(E? x);
...
+  /// Note: We want only one [List] implementation class to be stored here to
+  /// make sure the list operations are monomorphic and can be inlined. In
+  /// constructors make sure initializers for this field all return the same
+  /// implementation class. (e.g. `_GrowableList` on the VM)
   final List<E> _wrappedList;
+
+  /// A growable list, to be used in `unmodifiable` constructor to avoid
+  /// allocating a list every time.
+  ///
+  /// We can't use `const <E>[]` as it makes the `_wrappedList` field
```
mkustermann, line 25: We cannot have `const <E>[]`. Constant expressions in dart can only refer to instantiated types, not free type parameters.

-> Should be `const <Never>[]`

protobuf/lib/src/protobuf/pb_map.dart
```diff
@@ -21,24 +21,34 @@ class PbMap<K, V> extends MapBase<K, V> {
...
+  /// Note: We want only one [Map] implementation class to be stored here to
+  /// make sure the map operations are monomorphic and can be inlined. In
+  /// constructors make sure initializers for this field all return the same
+  /// implementation class.
   final Map<K, V> _wrappedMap;
 
-  bool _isReadonly = false;
+  /// The map instance to be used in `PbMap.unmodifiable`.
+  ///
+  /// We can't use `const <K, V>{}` as it makes the `_wrappedMap` field
```
mkustermann, line 34: ditto

mkustermann (approved):

protobuf/lib/src/protobuf/pb_list.dart
osa1: updated the comment with `const []`.

PR #903, opened 2023-11-22
Update getList and getMap return types

osa1:
This improves iteration performance by making `moveNext` and `current` direct calls in kernel. #902 does not have the same effect on `moveNext` and `current` calls, this change is needed even with #902.

This change was not possible before #880 as users could override the list and map types to types that are not `PbList`s or `PbMap`s.

Note: changes in the message field types (the plugin changes in this PR) are not necessary for the kernel improvements, but it doesn't hurt to generate more precise types everywhere.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -389,7 +389,7 @@ class _FieldSet {
   /// The implementation of a generated getter for map fields.
   PbMap<K, V> _$getMap<K, V>(GeneratedMessage parentMessage, int index) {
     final value = _values[index];
-    if (value != null) return value as PbMap<K, V>;
+    if (value != null) return value;
 
     final fi = _nonExtensionInfoByIndex(index) as MapFieldInfo<K, V>;
     assert(fi.isMapField);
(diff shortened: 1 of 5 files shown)


protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -387,9 +387,9 @@ class _FieldSet {
   }
 
   /// The implementation of a generated getter for map fields.
-  Map<K, V> _$getMap<K, V>(GeneratedMessage parentMessage, int index) {
+  PbMap<K, V> _$getMap<K, V>(GeneratedMessage parentMessage, int index) {
     final value = _values[index];
-    if (value != null) return value as Map<K, V>;
+    if (value != null) return value as PbMap<K, V>;
```
mkustermann, line 392: nit: For consistency: Either manually downcast with `as` (as here) or rely on auto-inserted downcasts (like in `_$getList<T>` above). The reason we support implicit casts is because the `_values` is `List<dynamic` (if it was `List<Object?>` then one would be forced to downcast explicitly all uses)

mkustermann (approved):
Nice!

protobuf/lib/src/protobuf/field_set.dart
osa1: Good catch. Updated this with an implicit cast.

vvanm:
Hi,

When will this change hit the pub.dev protobuf package?

Latest version on pub.dev is from Published [6 months ago](https://pub.dev/packages/protobuf/versions) ago

Vincent

PR #906, opened 2023-12-05
Small PbList improvements

osa1:
- Reduce binary size increase caused by cl/586928869 by adding `dart2js:never-inline` pragmas to `add`, `addAll`, `clear`.

- Override `isEmpty`, `isNotEmpty` avoid virtual calls to `length`.

- Override `get iterator` to use the known iterator type of `_wrappedList`, which is always a `_GrowableList`.

cl/587627017

### DIFF
--- a/protobuf/lib/src/protobuf/pb_list.dart
+++ b/protobuf/lib/src/protobuf/pb_list.dart
@@ -45,6 +45,7 @@ class PbList<E> extends ListBase<E> {
         _check = _checkNotNull;
 
   @override
+  @pragma('dart2js:never-inline')
   void add(E element) {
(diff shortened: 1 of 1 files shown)


mkustermann:
> Reduce binary size increase caused by cl/586928869 by adding dart2js:never-inline pragmas to add, addAll, clear.

That CL in itself is just making types more precise. If this leads to a size increase, it's because the dart2js inlining heuristics decided it's worthwhile inlining.  @rakudrama are the heuristics in dart2js doing the right thing here?

mkustermann (approved):

osa1:
@mkustermann see b/314275739 for discussion on this.

PR #908, opened 2023-12-29
Handle deprecated options in grpc services and methods, enum types and values, messages

osa1:
Similar to #909, a new kind of golden test is added here where the test input
is not a hand-crafted message descriptor but an actual .proto file.


sigurdm (approved):
LGTM

Issue #900, opened 2023-11-13
Generated gRPC client method stubs get no comments or deprecation annotations


calmh:
It would be nice if comments on rpc:s in the proto files were passed on as comments on generated method stubs. It would also be nice if `option deprecated = true` resulted in a `@Deprecated` annotation. Currently none of these things happen; for example,

```
// Deprecated: use GetLatestReportV2 (1.18.3)
rpc GetLatestReport(GetLatestReportRequest) returns (AerodromeReport) {
    option deprecated = true;
}
```

generates as

```
  $grpc.ResponseFuture<$8.AerodromeReport> getLatestReport($12.GetLatestReportRequest request, {$grpc.CallOptions? options}) {
    return $createUnaryCall(_$getLatestReport, request, options: options);
  }
```

which the client has no idea is deprecated, and gets no guidance from a comment.

osa1:
#908 implements generating deprecated annotations, #909 implements generating comments.

PR #909, opened 2023-12-29
Generate docs of enums and rpc clients, some refactoring

osa1:
This implements generating doc comments for enum types, enum values, rpc types,
and rpc methods.

Existing code for generating doc comments is refactored: the comment generator
handles trailing whitespace so we don't have to handle it in the call sites.
Comment generator is documented with the invariants: it never returns an empty
string (returns `null` instead) and never returns lines with trailing
whitespace.

To be able to test comment generating, a new kind of golden file test is added.
The difference from the other golden file tests is that this file is compiled
from .proto source instead of from a hand-crafted descriptor. This is because
attaching comments to proto field and types by hand is difficult and error
prone.

Fixes the comment part of #900.

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -5,9 +5,13 @@
   fields is now `PbMap` (instead of `Map`). ([#903])
 
   This change requires protobuf-4.0.0.
+* Generate doc comments for enum types and values, rpc services and methods.
+  ([#900], [#909])
 
 [#738]: https://github.com/google/protobuf.dart/issues/738
 [#903]: https://github.com/google/protobuf.dart/pull/903
+[#900]: https://github.com/google/protobuf.dart/issues/900
+[#909]: https://github.com/google/protobuf.dart/pull/909
 
 ## 21.1.2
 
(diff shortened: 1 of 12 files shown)


sigurdm (approved):

osa1:
@sigurdm for field comments, any thoughts on which accessors should have the comments copied?

Currently for a field `x` we generate these public members:

- `get x`
- `set x`
- `hasX`
- `clearX`

The documentation is currently copied only to `get x`, but deprecation annotations are added to all of these (in master branch, not with #908).

If duplicating potentially large comment blocks 4 times is not a problem I think we should add docs to all.

sigurdm:
> @sigurdm for field comments, any thoughts on which accessors should have the comments copied?

Really good question. I don't think the setter should have it: https://dart.dev/effective-dart/documentation#dont-write-documentation-for-both-the-getter-and-setter-of-a-property

Perhaps the hasX should just have a referring comment along the lines of:

> Checks for presence of [X].
>
> True if [X] has been set.

And clearX:

> Unsets [X].

osa1:
Makes sense, and thanks for the reference.

In that case I think we don't have to change anything in this PR.

I'm not sure if "clearX" and "hasX" docs are necessary, but if we want them we can add them separately.

PR #911, opened 2024-02-01
Optimize repeated field decoding

ignatz:
Avoid allocating tear-off closures when decoding repeated but non-packed fields.

Also refactors `BuilderInfo._decodeEnum`.

# Benchmarks

`repeated_int64`, `repeated_string`, and `repeated_enum` are the new benchmarks added with this PR.

`fromBuffer` is an internal benchmark that decodes a large buffer.

`fromBufferUnmodifiableInput` is the same as `fromBuffer`, but the input `Uint8List` is made unmodifiable.

## AOT

|                             | Before    | After     | Diff               |
|-----------------------------|-----------|-----------|--------------------|
| repeated_int64              | 84,549 us | 61,820 us | -22,729 us, -25.8% |
| repeated_string             | 67,183 us | 67,984 us | +801 us, +1.19%    |
| repeated_enum               | 52,335 us | 44,772 us | -7,563 us, -14.4%  |
| fromBuffer                  | 25,661 us | 25,299 us | -362 us, -1.4%     |
| fromBufferUnmodifiableInput | 26,135 us | 25,689 us | -446 us, -1.7%     |

## JS

|                             | Before     | After      | Diff               |
|-----------------------------|------------|------------|--------------------|
| repeated_int64              | 112,050 us | 109,850 us | -2,200 us, -1.9%   |
| repeated_string             | 123,050 us | 126,650 us | +3,600 us, +2.9%   |
| repeated_enum               |  43,920 us |  39,400 us | -4,520 us, -10.2%  |
| fromBuffer                  | 211,888 us | 168,250 us | -43,638 us, -20.5% |
| fromBufferUnmodifiableInput | 227,444 us | 185,454 us | -41,990 us, -18.4% |

### DIFF
--- a/benchmarks/bin/repeated_field.dart
+++ b/benchmarks/bin/repeated_field.dart
@@ -1,0 +1,45 @@
+import 'dart:typed_data';
+
+import 'package:fixnum/fixnum.dart';
+import 'package:protobuf_benchmarks/benchmark_base.dart';
+import 'package:protobuf_benchmarks/generated/f12.pb.dart' as f12;
+import 'package:protobuf_benchmarks/generated/google_message2.pb.dart';
+
+class RepeatedBenchmark extends BenchmarkBase {
+  final Uint8List _buffer;
+
+  RepeatedBenchmark(super.name, GoogleMessage2 message)
+      : _buffer = message.writeToBuffer();
+
+  @override
+  void run() => GoogleMessage2.fromBuffer(_buffer);
+}
+
+class RepeatedEnumBenchmark extends BenchmarkBase {
+  final Uint8List _buffer;
+
+  RepeatedEnumBenchmark(super.name, f12.A58 message)
+      : _buffer = message.writeToBuffer();
+
+  @override
+  void run() => f12.A58.fromBuffer(_buffer);
+}
+
+void main() {
+  const kSize = 500000;
+
+  RepeatedBenchmark(
+    'repeated_int64',
+    GoogleMessage2(field130: List<Int64>.generate(kSize, Int64.new)),
+  ).report();
+
+  RepeatedBenchmark(
+    'repeated_string',
+    GoogleMessage2(field128: List<String>.generate(kSize, (i) => i.toString())),
+  ).report();
+
+  RepeatedEnumBenchmark(
+    'repeated_enum',
+    f12.A58(a306: List<f12.A322>.generate(kSize, (_) => f12.A322.A324)),
+  ).report();
+}
--- a/protobuf/lib/src/protobuf/coded_buffer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer.dart
@@ -33,19 +33,8 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
     CodedBufferReader input, ExtensionRegistry registry) {
   ArgumentError.checkNotNull(registry);
   fs._ensureWritable();
-
-  // Micro-optimization: cache the storage lookup for repeated fields.
-  var prevTag = -1;
-  List? cachedList;
-
   while (true) {
     final tag = input.readTag();
-    // If the current field's tag is different from previous, invalidate cache.
-    if (tag != prevTag) {
-      cachedList = null;
-      prevTag = tag;
-    }
-
     if (tag == 0) return;
     final wireType = tag & 0x7;
     final tagNumber = tag >> 3;
@@ -139,7 +128,7 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
         }
         break;
       case PbFieldType._REPEATED_BOOL:
-        final list = cachedList ??= fs._ensureRepeatedField(meta, fi);
+        final list = fs._ensureRepeatedField(meta, fi);
         if (wireType == WIRETYPE_LENGTH_DELIMITED) {
           _readPacked(input, () => list.add(input.readBool()));
         } else {
@@ -147,15 +136,15 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
         }
         break;
       case PbFieldType._REPEATED_BYTES:
-        final list = cachedList ??= fs._ensureRepeatedField(meta, fi);
+        final list = fs._ensureRepeatedField(meta, fi);
         list.add(input.readBytes());
         break;
       case PbFieldType._REPEATED_STRING:
-        final list = cachedList ??= fs._ensureRepeatedField(meta, fi);
+        final list = fs._ensureRepeatedField(meta, fi);
         list.add(input.readString());
         break;
       case PbFieldType._REPEATED_FLOAT:
-        final list = cachedList ??= fs._ensureRepeatedField(meta, fi);
+        final list = fs._ensureRepeatedField(meta, fi);
         if (wireType == WIRETYPE_LENGTH_DELIMITED) {
           _readPacked(input, () => list.add(input.readFloat()));
         } else {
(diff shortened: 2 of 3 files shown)


osa1:
Thanks, this is great. I've also confirmed in some internal benchmarks (not micro benchmarks) that this makes quite a big difference. I will try to review and merge this this week.

protobuf/lib/src/protobuf/coded_buffer.dart
```diff
@@ -199,52 +280,45 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
...
     // Not packed.
-    readToList(list);
+    final rawValue = input.readEnum();
+    final value = meta._decodeEnum(tagNumber, registry, rawValue);
+    if (value == null) {
+      final unknown = fs._ensureUnknownFields();
+      unknown.mergeVarintField(tagNumber, Int64(rawValue));
+    } else {
+      list.add(value);
+    }
```
osa1, line 322: Why do you handle enums differently than all the others? E.g. why not do this:

```dart
        final list = cachedList ??= fs._ensureRepeatedField(meta, fi);
        if (wireType == WIRETYPE_LENGTH_DELIMITED) {
          _readPacked(input, () => readEnum(input));
        } else {
          list.add(readEnum(input));
        }
        break;
```

or handle rest of the types similarly, with functions like `_readPackableToListInt64`, `_readPackableToListSing32` etc.?

Currently enums are handled differently (more efficiently) than all the others, but I don't know why this should be the case.

protobuf/lib/src/protobuf/coded_buffer.dart
```diff
@@ -33,8 +33,19 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
...
+  var prevTag = -1;
+  List? cachedList;
+
   while (true) {
     final tag = input.readTag();
+    // If the current field's tag is different from previous, invalidate cache.
+    if (tag != prevTag) {
+      cachedList = null;
+      prevTag = tag;
+    }
```
osa1, line 47: In my benchmarking, caching the list actually makes it slower when compiled to JS with `dart2js -O4`.

It seems to make a small improvement in AOT so can still keep it, but I'm not sure if it's worth it.

See my top-level comment for details.

osa1 (commented):
Thanks again, this is great.

While reviewing this I noticed that the wins here are not coming from the cached list, but from avoiding the tear-off closure creating when reading repeated fields, for every element.

Indeed, `fs._ensureRepeatedField(...)` is just a list access in the common case:

```dart
  PbList<T> _ensureRepeatedField<T>(BuilderInfo meta, FieldInfo<T> fi) {
    assert(!_isReadOnly);
    assert(fi.isRepeated);
    if (fi.index == null) {                             // <----- branch not taken in the common case
      return _ensureExtensions()._ensureRepeatedField(fi as Extension<T>);
    }
    final value = _getFieldOrNull(fi);                  // <----- returns non-null after the first element of a list
    if (value != null) return value;
    ...
  }

  dynamic _getFieldOrNull(FieldInfo fi) {
    if (fi.index != null) return _values[fi.index!];    // <----- common case
    return _extensions?._getFieldOrNull(fi as Extension<dynamic>);
  }
```

So it makes sense that it doesn't make too much difference. (actually it makes my benchmark slower when compiled to JS, but that may be noise, see below for numbers)

The wins here are coming from avoiding the closures in e.g.

```dart
      case PbFieldType._REPEATED_BOOL:
        _readPackable(meta, fs, input, wireType, fi, input.readBool);
```

Here for an unpacked list, we create one tear-off closure (`input.readBool`) for every element. That's expensive compared to the cost of reading one element. The version in this PR avoids it when the field is not packed:

```dart
      case PbFieldType._REPEATED_BOOL:
        final list = cachedList ?? fs._ensureRepeatedField(meta, fi);
        if (wireType == WIRETYPE_LENGTH_DELIMITED) {
          _readPacked(input, () => list.add(input.readBool()));
        } else {
          list.add(input.readBool());           // <----- no closure allocation when not packed
        }
```

(Allocating a closure when reading packed is fine, as closure allocation is probably a small part of the whole operation, though we could even avoid that if we wanted, similar to enum code in this PR, which doesn't allocate any closures)

We should update the commit message to reflect this. I can do this if you prefer.

# Benchmark results

From an internal benchmark.

## Master branch

### AOT

fromBuffer(RunTime): 26,048 us.
fromBufferUnmodifiableInput(RunTime): 26,596 us.

### JS

fromBuffer(RunTime): 214,111 us.
fromBufferUnmodifiableInput(RunTime): 230,777 us.

## With cached list

### AOT

fromBuffer(RunTime): 24,877 us.
fromBufferUnmodifiableInput(RunTime): 24,864us.

### JS

fromBuffer(RunTime): 193,333 us.
fromBufferUnmodifiableInput(RunTime): 192,500 us.

## Same patch, but without the cached list

### AOT

fromBuffer(RunTime): 25,388 us.
fromBufferUnmodifiableInput(RunTime): 25,073 us.

### JS

fromBuffer(RunTime): 171,272 us.
fromBufferUnmodifiableInput(RunTime): 189,000 us.

# Notes

Cached list is not that important (actually makes dart2js output slower), the important part is avoiding creating a tear-off closure for every element when reading non-packed repeated fields.

Similar differences can be seen with the benchmark in this PR.

benchmarks/bin/repeated_field.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  @override
+  void run() => f12.A58.fromBuffer(_buffer);
+}
+
+void main() {
+  const kSize = 500000;
+
+  RepeatedBenchmark(
+    'repeated_in64',
```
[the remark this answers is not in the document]
ignatz: I swear I can spell ,at least 5% of the time :)

protobuf/lib/src/protobuf/coded_buffer.dart
ignatz: I've lost a lot of context but looking at the code I'd say it's probably because enums and ints are handled differently in the skew case. Specifically look at the "unknown fields" path:

```dart
final rawValue = input.readEnum();
final value = meta._decodeEnum(tagNumber, registry, rawValue);
if (value == null) {
  final unknown = fs._ensureUnknownFields();
  unknown.mergeVarintField(tagNumber, Int64(rawValue));
} else {
  list.add(value);
}
```

If you look at the non repeated flavors, enums are also handled differently:

```dart
case PbFieldType._OPTIONAL_DOUBLE:
  fs._setFieldUnchecked(meta, fi, input.readDouble());
  break;
case PbFieldType._OPTIONAL_ENUM:
  final rawValue = input.readEnum();
  final value = meta._decodeEnum(tagNumber, registry, rawValue);
  if (value == null) {
    final unknown = fs._ensureUnknownFields();
    unknown.mergeVarintField(tagNumber, Int64(rawValue));
  } else {
    fs._setFieldUnchecked(meta, fi, value);
  }
  break;
````

And lastly, your proposal simply breaks the benchmark

ignatz:
> We should update the commit message to reflect this. I can do this if you prefer.

Your findings match my findings. I never intended to claim that the caching has a big impact. My commit message specifically calls out "Boxing overhead", which is my, maybe inadequate lingo, to say I removed a closure.

I'm not sure you wanted to say that it's the allocation itself that is solely responsible for the overhead because I'm not sure that's true.

I left the cache from my experiments because it seemed fairly simple and not entirely pointless. It's not clear to me how it would slow things down unless this is some memory layout weirdness. We can remove it

Lastly, you should always feel free to edit my commit message

osa1:
I'm testing this internally in cl/609281182.

osa1 (approved):

PR #913, opened 2024-02-23
Allow sizeLimit to be passed to mergeFromBuffer so it can be passed t…

vvanm:
…o CodedBufferReader to override the 64MB size limit

We have cases in our flutter application where we need to parse files larger than 64MB.
Since there is a hard limit in CodedBuffferReader of 64MB we are running into issues there.

Since sizeLimit cannot be passed in from anywhere, this PR hopes to fix that so the sizeLimit is customizable from outside protobuf.


osa1:
> Since sizeLimit cannot be passed in from anywhere

Instead of

```dart
final msg = MyMessage.fromBuffer(data, registry);
```

you can do:

```dart
final reader = CodedBufferReader(data, sizeLimit: data.length);
final msg = MyMessage.create()..mergeFromCodedBufferReader(reader, registry);
```

vvanm:
Hi,

I didn't notice that, thanks for the tip.

We using protoc like this:
protoc --proto_path=packages/metadata/proto --dart_out packages/metadata/lib packages/metadata/proto/root.proto packages/metadata/proto/metadata.proto packages/metadata/proto/protobuf-net/bcl.proto

Which generates code like 

class Asset extends $pb.GeneratedMessage {
////

factory Asset.fromBuffer($core.List<$core.int> i, [$pb.ExtensionRegistry r = $pb.ExtensionRegistry.EMPTY]) => create()..mergeFromBuffer(i, r);

////
}

Any tips on how to modify the generation so the reader is created and a limit can be set? Are there parameters on protoc command that can do this?

Thank you,
Vincent

osa1:
You can't modify the generated code, you need to use `mergeFromCodedBufferReader` with a `CodedBufferReader` with a custom limit, as shown above.

PR #918, opened 2024-03-04
Support unknown json data and add internal set/clear field methods for generated accessors.

### LINKED ISSUE #49: Add support for unknown fields in JSON serialization
When reading and writing protobufs to JSON, we currently only serialize extensions if they are registered. Unknown fields are ignored.

This is a problem for the JSON-encoded constants for metadata in .pb.dart files. Any extensions in the original protobuf descriptor get stripped. It would be good to encode them somehow, instead.

biggs0125:
Add support for unknown fields to JSON (de-)serialized proto messages. This allows data to be roundtripped from a server via this format without dropping fields when the client doesn't contain metadata for some fields.

Store the data as decoded but unparsed json and append it back to the message when re-serializing a message.

Ignore unknown json data for equality/hashing to preserve existing behavior for users. Taking this data into account would change the result of existing `==` and `hashCode` calls.

By separating out `$_setField` and `$_clearField` this allows one to differentiate reflective uses of `setField` and `clearField` (i.e. from user code, usually with the result of `getTagNumber`) vs. internal uses from the generated accessors.

Fixes #49.

---

cl/611522109
cl/612283849


osa1 (approved):
Thanks!

PR #931, opened 2024-05-11
Generate actuall `enums`

### LINKED ISSUE #862: Consider generating actual `enum`s
A Protobuf enum gets generated to a class that extends `ProtobufEnum`, to ensure they provide `name` and `value`. But now with enhanced enums, they can be actual `enums` and have better switch support, namely "switch expressions", which doesn't allow exhaustiveness with the current implementation.

Currently, the `ProtobufEnum` class has two final fields that need to be passed to a constructor. Enhanced enums can't extend classes, but they can implement them. So enums can be generated as: 
```dart
enum MyEnum implements ProtobufEnum {
  a("a", 0), 
  b("b", 1),
  c("c", 2);

  @override final String name;
  @override final int value;
  const MyEnum(this.name, this.value);
}
```
Which will then enable proper exhaustiveness checking in switch expressions, like the following: 
```dart
String getName(MyEnum e) => switch (e) {
  MyEnum.a => "a",
  // Error: The type 'MyEnum' is not exhaustively matched by the switch cases since it doesn't match 'MyEnum.b'.
}
```

fischerscode:
This PR changes the enum generation to generate actual enums.

I think sooner or later this is a necessary step to take full advantage of dart`s features.

#862 Also motivates this change.

I tried to keep everything as compatible as possible. I think my implementation should be fully compatible. (Except of sup-typing generated enums. But I think nobody has done this.)

I've grouped my changes into four commits to simplify reviewing this PR.

Close #862

### DIFF
--- a/protoc_plugin/Makefile
+++ b/protoc_plugin/Makefile
@@ -111,7 +111,7 @@ update-pregenerated: $(PLUGIN_PATH) $(PREGENERATED_SRCS)
 protos: $(PLUGIN_PATH) $(TEST_PROTO_LIBS)
 
 run-tests: protos
-	dart test
+	dart test && dart -Dprotobuf.omit_enum_names=true test
 
 update-goldens: protos
 	rm -rf test/goldens
--- a/protoc_plugin/lib/src/enum_generator.dart
+++ b/protoc_plugin/lib/src/enum_generator.dart
@@ -110,10 +110,10 @@ class EnumGenerator extends ProtobufContainer {
       out.println('@$coreImportPrefix.Deprecated(\'This enum is deprecated\')');
     }
     out.addAnnotatedBlock(
-        'class $classname extends $protobufImportPrefix.ProtobufEnum {',
+        'enum $classname implements $protobufImportPrefix.ProtobufEnum {',
         '}\n', [
       NamedLocation(
-          name: classname, fieldPathSegment: fieldPath, start: 'class '.length)
+          name: classname, fieldPathSegment: fieldPath, start: 'enum '.length)
     ], () {
       // -----------------------------------------------------------------
       // Define enum types.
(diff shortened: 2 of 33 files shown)


protoc_plugin/lib/src/enum_generator.dart
```diff
@@ -164,27 +162,29 @@ class EnumGenerator extends ProtobufContainer {
...
+      out.println('const $classname(this.value, this.name);');
+      out.println();
       out.println(
-          'const $classname._($coreImportPrefix.int v, $coreImportPrefix.String n) '
-          ': super(v, n);');
+          "/// Returns this enum's [name] or the [value] if names are not"
+          ' represented.');
+      out.println('@$coreImportPrefix.override');
+      out.println("$coreImportPrefix.String toString() => name == '' ? "
+          'value.toString() : name;');
```
fischerscode, line 187: I had to generate a "custom" `toString` method for each enum, since the default enum `toString` takes priority over the one from `ProtobufEnum`.

This method is designed to be 100% compatible, but it might be better to return `Enum.name` (`super.toString()`) instead of `name`.

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -644,7 +644,6 @@ class FileGenerator extends ProtobufContainer {
 //  Generated code. Do not modify.
 //  source: ${descriptor.name}
 //
-// @dart = 2.12
```
fischerscode, line 647: I've removed all of them, since they would mess up the enum files and I don't think they serve any purpose.

protoc_plugin/lib/src/generated/dart_options.pb.dart
```diff
@@ -171,7 +172,7 @@ class Imports extends $pb.GeneratedMessage {
   ///  so the generated code may contain errors. Therefore, running dartanalyzer
   ///  on the generated file is a good idea.
   @$pb.TagNumber(1)
-  $core.List<DartMixin> get mixins => $_getList(0);
+  $pb.PbList<DartMixin> get mixins => $_getList(0);
```
fischerscode, line 175: I'm not sure why this change. I assume the files have not been regenerated in the past.

protoc_plugin/Makefile
```diff
@@ -111,7 +111,7 @@ update-pregenerated: $(PLUGIN_PATH) $(PREGENERATED_SRCS)
 protos: $(PLUGIN_PATH) $(TEST_PROTO_LIBS)
 
 run-tests: protos
-	dart test
+	dart test && dart -Dprotobuf.omit_enum_names=true test
```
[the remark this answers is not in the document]
fischerscode: Turns out this has to be `dart -Dprotobuf.omit_enum_names=true test --use-data-isolate-strategy` (changed it in 2b145bee53d0187a1d9beca7535f7e4f2c26e0cc)
See https://github.com/dart-lang/test/issues/1794#issuecomment-1322721423

Unfortunately this is much slower. It might be best to change the tests that depend on this.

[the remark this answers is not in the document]
fischerscode: Update: I've introduced the test tag `tests_omit_enum_names`. (d6600159445e6f47cef92150c93b74520347693f)

PR #935, opened 2024-06-14
Change the method of IntList equals from toString to ListEquality.equals. In my case, codegen 15x faster

qpalwo:
I have a 6000-line proto file, and it takes one minute to generate Dart code using the Dart proto plugin. After changing to the ListEquality.equals method, it only takes 4.6 seconds. If I completely remove the commentBlock call and compile it into an AOT (Ahead-of-Time) artifact, it only takes 200ms.

> ❯ time protoc --dart_out=./dart/lib1 test.proto
> protoc --dart_out=./dart/lib1 test.proto  62.60s user 0.75s system 100% cpu 1:03.31 total
> ❯ time protoc --dart_out=./dart/lib1 --plugin=protobuf.dart/protoc_plugin/bin/protoc-gen-dart test.proto
> protoc --dart_out=./dart/lib1  test.proto  4.61s user 0.32s system 125% cpu 3.918 total

### DIFF
--- a/protoc_plugin/lib/src/shared.dart
+++ b/protoc_plugin/lib/src/shared.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 import 'dart:io';
+import 'package:collection/collection.dart';
 
 import '../protoc.dart';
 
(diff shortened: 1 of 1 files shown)


osa1:
Thanks @qpalwo.

In the internal verison of this library we improve this even more by scanning the lists backwards, as it's more likely for two paths to share a prefix than a suffix.

But this version is still a big improvement compared to the current implementation.

Are you able to sign the CLA so that we can merge this?

PR #953, opened 2025-02-05
Replace removed `dart:typed_data` type in test code

osa1:
`Unmodifiable...View` types were deprecated in favor of `asModifiableView`s members introduced in SDK 3.3.0.

These types are now removed, so the protobuf package tests can't be built with the most recent SDK.

Bump the SDK dependency to 3.3.0 (the version introduced `asModifiableView`) and replace the removed type with `asModifiableView` in the test.

### DIFF
--- a/protobuf/test/coded_buffer_reader_test.dart
+++ b/protobuf/test/coded_buffer_reader_test.dart
@@ -84,11 +84,7 @@ void main() {
     });
 
     test('unmodifiable-uint8-list-view', () {
-      // TODO: Use `Uint8List.asUnmodifiableView` instead of
-      // `UnmodifiableUint8ListView` when it's available in the oldest
-      // supported SDK version.
-      // ignore: deprecated_member_use
-      testWithList(UnmodifiableUint8ListView(Uint8List.fromList(inputBuffer)));
+      testWithList(Uint8List.fromList(inputBuffer).asUnmodifiableView());
     });
 
     test('uint8-list-view', () {
--- a/protobuf/pubspec.yaml
+++ b/protobuf/pubspec.yaml
@@ -2,11 +2,11 @@ name: protobuf
 version: 4.0.0-dev
 description: >-
   Runtime library for protocol buffers support. Use with package:protoc_plugin
-  to generate dart code for your '.proto' files.
+  to generate Dart code for your '.proto' files.
 repository: https://github.com/google/protobuf.dart/tree/master/protobuf
 
 environment:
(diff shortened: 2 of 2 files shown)


.github/workflows/dart.yml
```diff
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.5.7
+# Created with package:mono_repo v6.6.2
```
devoncarew, line 1: We can also move away from using this mono_repo generated workflow file - in favor of some simple hand-written ones - if that's preferable.

api_benchmark/mono_pkg.yaml
```diff
@@ -6,5 +6,4 @@ stages:
       - format
       - command: ./../tool/setup.sh
       - command: ./compile_protos.sh
-      - analyze: --fatal-infos
```
devoncarew, line 9: +1 to removing this package

protobuf/test/coded_buffer_reader_test.dart
```diff
@@ -84,11 +84,7 @@ void main() {
     });
 
     test('unmodifiable-uint8-list-view', () {
-      // TODO: Use `Uint8List.asUnmodifiableView` instead of
-      // `UnmodifiableUint8ListView` when it's available in the oldest
-      // supported SDK version.
-      // ignore: deprecated_member_use
-      testWithList(UnmodifiableUint8ListView(Uint8List.fromList(inputBuffer)));
+      testWithList(Uint8List.fromList(inputBuffer).asUnmodifiableView());
```
devoncarew, line 87: 👍

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -644,7 +644,7 @@ class FileGenerator extends ProtobufContainer {
 //  Generated code. Do not modify.
 //  source: ${descriptor.name}
 //
-// @dart = 2.12
+// @dart = 3.3
```
devoncarew, line 647: Fantastic!

devoncarew (approved):
lgtm!

.github/workflows/dart.yml
osa1: I think it's fine, I don't have any problems with it.

devoncarew: https://github.com/google/protobuf.dart/pull/957 moves to using simple, hand-written workflow files per-package instead of a generated mono_repo one.

PR #957, opened 2025-02-08
refactor repo CI and analysis options files

devoncarew:
- refactor the repo's CI setup - prefer small, per-package workflow files to using mono_repo
- refactor the analysis options files - move to using per-package files
- update the config for the publishing workflow (allow workflow feedback from forks)


osa1 (commented):
What is the problem with mono_repo? I liked that the mono_repo configs are short (at least for what we do in this library) and I don't have to review and test generated YAMLs as they're generated and I can trust that mono_repo is well tested.

> update the config for the publishing workflow (allow workflow feedback from forks)

What is "workflow feedback from forks"?

Regarding lints: I think it would be good to be in sync with the g3 version rather than Flutter (or any other) lints as at this point this library is more actively developed in g3 than here.

devoncarew:
> What is the problem with mono_repo? I liked that the mono_repo configs are short (at least for what we do in this library) and I don't have to review and test generated YAMLs as they're generated and I can trust that mono_repo is well tested.

I'm not a fan of the abstraction over github workflow files; it may have been a useful abstraction over travis configs but I don't think adds real value over using straight github workflow files. But if you're happy w/ the setup for this repo I won't push for changing it.

> > update the config for the publishing workflow (allow workflow feedback from forks)
> 
> What is "workflow feedback from forks"?

The publish pre-check workflow can only provide feedback for PRs from this repo - it can't write comments to PRs coming from forks of the repo. This updated config allows us to see the results of the publish workflow from PRs from forks as well.

> Regarding lints: I think it would be good to be in sync with the g3 version rather than Flutter (or any other) lints as at this point this library is more actively developed in g3 than here.

We no longer have a lint set that represents what's used internally (pedantic was that set a while ago). dart_flutter_team_lints may be the closest, with some additional knobs tightened.

devoncarew:
(closing in favor of more specific PRs: https://github.com/google/protobuf.dart/pull/960, https://github.com/google/protobuf.dart/pull/961, https://github.com/google/protobuf.dart/pull/962)

PR #959, opened 2025-02-10
Improve packed field decoding

osa1:
- Inline `_readPacked` manually and `_withLimit` with a pragma to eliminate
  closure allocation and calls in packed decoding loops.

  - `_readPacked` is manually inlined as VM's inliner doesn't properly fold
    constants after the inlinings, see
    https://github.com/dart-lang/sdk/issues/60068.

- Introduce `PbList._addUnchecked` to add to the list without checking the
  value for validity and list for mutability.

- When decoding a packed field, check the list mutability once, instead of in
  every element.

- When decoding a packed scalar field, don't check for value validity.

  For scalar fields we need to make sure the field value is not null, which is
  already guaranteed in the call sites as e.g. `input.readDouble` doesn't
  return nullable.

- Sprinkle a bunch of `prefer-inline`s to make sure VM will inline one liners.

Looking at the optimized flow graph, thish would be quite fast compared to the
current version. However I can't benchmark this yet as none of the benchmarks
use packed fields.

I will update the benchmarks first, then revisit this PR.

### DIFF
--- a/protobuf/lib/src/protobuf/coded_buffer.dart
+++ b/protobuf/lib/src/protobuf/coded_buffer.dart
@@ -130,9 +130,11 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
       case PbFieldType._REPEATED_BOOL:
         final list = fs._ensureRepeatedField(meta, fi);
         if (wireType == WIRETYPE_LENGTH_DELIMITED) {
-          @pragma('vm:prefer-inline')
-          void readElem() => list.add(input.readBool());
-          _readPacked(input, readElem);
+          input._withLimit(input.readInt32(), () {
+            while (!input.isAtEnd()) {
+              list.add(input.readBool());
+            }
+          });
         } else {
           list.add(input.readBool());
         }
@@ -148,9 +150,11 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
       case PbFieldType._REPEATED_FLOAT:
         final list = fs._ensureRepeatedField(meta, fi);
         if (wireType == WIRETYPE_LENGTH_DELIMITED) {
-          @pragma('vm:prefer-inline')
-          void readElem() => list.add(input.readFloat());
-          _readPacked(input, readElem);
+          input._withLimit(input.readInt32(), () {
+            while (!input.isAtEnd()) {
+              list.add(input.readFloat());
+            }
+          });
         } else {
           list.add(input.readFloat());
         }
@@ -158,9 +162,11 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
       case PbFieldType._REPEATED_DOUBLE:
         final list = fs._ensureRepeatedField(meta, fi);
         if (wireType == WIRETYPE_LENGTH_DELIMITED) {
-          @pragma('vm:prefer-inline')
-          void readElem() => list.add(input.readDouble());
-          _readPacked(input, readElem);
+          input._withLimit(input.readInt32(), () {
+            while (!input.isAtEnd()) {
+              list.add(input.readDouble());
+            }
+          });
         } else {
           list.add(input.readDouble());
         }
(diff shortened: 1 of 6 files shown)


protobuf/lib/src/protobuf/coded_buffer.dart
```diff
@@ -129,33 +129,65 @@ void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
...
         } else {
-          list.add(input.readBool());
+          list._checkModifiable('add');
+          list._addUnchecked(input.readBool());
         }
         break;
       case PbFieldType._REPEATED_BYTES:
         final list = fs._ensureRepeatedField(meta, fi);
-        list.add(input.readBytes());
+        list._checkModifiable('add');
```
mkustermann, line 151: It seems to be hand-inlined code, omitting the `_check` call. It seems not very nice we have to copy&paste this many times.

Why not pass in a `bool` whether to check or not, and force inline the function. E.g.

```dart
_addInternal(input.readBytes(), omitCheck: true);

@pragma('vm/dart2js/dart2wasm:prefer-inline')
_addInternal(E value, {bool omitCheck = false}) {
  _checkModifiable(value, 'add');
  if (!omitCheck) _check(value);
  _wrappedList.add(value);
}
```

mkustermann:
It seems in dart2wasm we have
```
// before
protobuf_PackedInt64Decoding(RunTimeRaw): 77833.33333333334 us.

// after
protobuf_PackedInt64Decoding(RunTimeRaw): 90650.0 us.
```
i.e. a regression, is this expected?

benchmarks/bin/binary_decode_packed.dart
```diff
@@ -0,0 +1,171 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:math';
+import 'dart:typed_data';
+
+import 'package:fixnum/fixnum.dart';
+import 'package:protobuf_benchmarks/benchmark_base.dart';
+import 'package:protobuf_benchmarks/generated/packed_fields.pb.dart';
+
+PackedFields? sink;
```
mkustermann, line 12: It seems this global `sink` is never used. That will make TFA remove it, which in turn makes the results of calls like `message.writeToBuffer()` unused, which may make optimizer remove optimize things away.

Could you add a `if (int.parse('1') == 0) print(sink)` to `main` to avoid this?

mkustermann (approved):
lgtm with comments

osa1:
> It seems in dart2wasm we have
> 
> ```
> // before
> protobuf_PackedInt64Decoding(RunTimeRaw): 77833.33333333334 us.
> 
> // after
> protobuf_PackedInt64Decoding(RunTimeRaw): 90650.0 us.
> ```
> 
> i.e. a regression, is this expected?

Maybe my CPU became busy with other things while running it, when I run the benchmarks again I consistently get better numbers with this PR. I updated the PR description with more recent benchmark numbers from a run just now.

benchmarks/bin/binary_decode_packed.dart
osa1: Good catch -- I wanted to do this but then forgot..

With `print(sink)` VM benchmarks are the same as before, but in Wasm both before and after the benchmarks get much slower. The new version is still faster. I updated the PR description with new benchmarks.

protobuf/lib/src/protobuf/coded_buffer.dart
```diff
@@ -29,8 +29,8 @@ void _writeToCodedBufferWriter(_FieldSet fs, CodedBufferWriter out) {
   }
 }
 
-void _mergeFromCodedBufferReader(BuilderInfo meta, _FieldSet fs,
-    CodedBufferReader input, ExtensionRegistry registry) {
+void _mergeFromCodedBufferReader(final BuilderInfo meta, final _FieldSet fs,
```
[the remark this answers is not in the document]
osa1: It's been a while since I wrote this code, but IIRC this was done based on discussions with you and/or Alex.

IIRC VM was more likely to inline with `final` arguments.

However I now benchmarked before and after these `final`s and it didn't make any difference. So I removed them now.

protobuf/lib/src/protobuf/coded_buffer.dart
osa1: This function wouldn't work below where we check for modifiable once before adding the elements, and then never check it while adding the elements.

We could have this version:

```dart
_addInternal(E value, {bool omitElementCheck = false, bool omitModifiableCheck = false}) {
  if (!omitModifiableCheck) _checkModifiable(value, 'add');
  if (!omitElementCheck) _check(value);
  _wrappedList.add(value);
}
```

But we would still need a `_checkModifiable` call in packed fields as we want to run it once.

Overall not sure if this version is better than the current..

osa1:
I tested this internally in cl/755309114. Merging ...

PR #961, opened 2025-02-10
update the publish workflow; add a post_summaries one

devoncarew:
- update the publish workflow to write its comment to a workflow file
- add a post_summaries workflow - pull from the file and comment on the current PR

This config will allow PRs from forks to still get feedback in the PR from the publish workflow.

### DIFF
--- a/.github/workflows/post_summaries.yaml
+++ b/.github/workflows/post_summaries.yaml
@@ -1,0 +1,17 @@
+name: Comment on the pull request
+
+on:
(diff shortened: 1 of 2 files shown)


kevmoo (approved):
interesting...is this because it's not our org?

devoncarew:
This is the new best practice for setting up the publish (or health) workflows. cc @mosuem

mosuem:
> interesting...is this because it's not our org?

Unfortunately, this is a general Github thing.

PR #967, opened 2025-03-09
Update to latest lints

kevmoo:


kevmoo:
@osa1 ? @mraleph ?

osa1 (commented):
These lints change every quarter or so and cause massive headache every time, as internal and open source versions use different set of lints.

Given that the library has been mainly developed internally in the past year or so (in the last year internal version got protobuf editions support, new, more efficient JSON encoder and decoder for web builds) and it got almost no patches in the open source in the same time period, it might make sense to just sync the internal version.

protoc_plugin/lib/src/generated/descriptor.pb.dart
```diff
@@ -2226,7 +2226,7 @@ class MessageOptions extends $pb.GeneratedMessage {
   ///  maps field.
   ///
   ///  For maps fields:
-  ///      map<KeyType, ValueType> map_field = 1;
+  ///      `map<KeyType, ValueType> map_field = 1;`
```
osa1, line 2229: Is this intentional? This edits a generated file.

There are also other code blocks in this file that are not quoted like this one.

osa1 (commented):
I partially reviewed this.

Changes in these directores LGTM:

- protoc_plugin
- protobuf/test
- api_benchmark
- benchmarks

I added two inline comments about the rest (protobuf/lib).

protoc_plugin/lib/src/generated/descriptor.pb.dart
kevmoo: It is. Perhaps the lint needs to be tweaked. It's to avoid treating `<` like HTML. We generally like triple-tick blocks for these, perhaps.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -162,17 +162,17 @@ class FieldInfo<T> {
...
   /// That is, it doesn't contain any required fields that aren't initialized.
-  bool _hasRequiredValues(value) {
+  bool _hasRequiredValues(Object? value) {
     if (value == null) return !isRequired; // missing is okay if optional
     if (!_isGroupOrMessage(type)) return true; // primitive and present
 
     if (!isRepeated) {
       // A required message: recurse.
-      final GeneratedMessage message = value;
+      final message = value as GeneratedMessage;
```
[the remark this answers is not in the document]
kevmoo: done!

kevmoo:
@osa1 – @devoncarew is working so the lints change about once per year. This package is just very behind.

protoc_plugin/lib/src/generated/descriptor.pb.dart
kevmoo: I ignored this lint instead in generated code.

kevmoo:
Oh! If we have a version internally, then we should sync that! I see folks want more features here, too!

devoncarew:
> Oh! If we have a version internally, then we should sync that! I see folks want more features here, too!

Yeah, if the source-of-truth has mostly moved internally, we should reduce the diffs that PRs here may cause to that sourcebase. Perhaps that's moving the lints down to core? recommended? Whatever will provide some signal for PRs here w/o causing diffs from the internal code.

PR #976, opened 2025-05-01
Fix factory argument types for map fields

### LINKED ISSUE #975: Generated Dart code expects `PbMap` instead of `Map` which is hard to create
Previously it was:
```dart
class Foo extends $pb.GeneratedMessage {
  factory Foo({
    $core.Map<$core.String, Bar>? bars,
  }) {
    final $result = create();
    if (bars!= null) {
      $result.bars.addAll(bars);
    }
    return $result;
  }
  ...
}
```
now it is:
```dart
class Foo extends $pb.GeneratedMessage {
  factory Foo({
    $pb.PbMap<$core.String, Bar>? bars,
  }) {
    final $result = create();
    if (bars!= null) {
      $result.bars.addAll(bars);
    }
    return $result;
  }
  ...
}
```

Creating `PbMap` manually is very problematic, as it expects 2 ints (one for key and one for value). Personally I don't know how to obtain them. Situation is much simpler with `PbList` which just exposes `.from(List<E> from)` constructor

osa1:
Fixes #975.

---

cl/755740416

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 22.0.2-wip
+
+* Fix factory argument types for protobuf `Map` fields. ([#975])
+
+[#975]: https://github.com/google/protobuf.dart/issues/975
+
 ## 22.0.1
 
 * Bump `protobuf` constraint to `^4.0.0`
--- a/protoc_plugin/lib/src/message_generator.dart
+++ b/protoc_plugin/lib/src/message_generator.dart
@@ -446,6 +446,12 @@ class MessageGenerator extends ProtobufContainer {
         if (field.isRepeated && !field.isMapField) {
           out.println(
               '  ${field.baseType.getRepeatedDartTypeIterable(fileGen)}? ${field.memberNames!.fieldName},');
+        } else if (field.isMapField) {
+          final keyType = field.getDartMapKeyType();
+          final valueType = field.getDartMapValueType();
+          out.println(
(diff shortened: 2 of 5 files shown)


osa1:
I think we could just publish this. Any thoughts @devoncarew? I can bump the version number to non-wip.

devoncarew (approved):
lgtm

> I think we could just publish this. Any thoughts @devoncarew? I can bump the version number to non-wip.

For many commonly used packages, we like to roll into the sdk and the then google3 before publishing in order to shake out any issues. This package is lightly used in the sdk (so testing there would be useful but less critical), and afaik doesn't roll into google3 anymore.

I think landing this and rolling into the sdk would make sense; we could roll the commits for both this and #959 together.

If that looks good, we can rev the version here and publish (we don't need to validate a version only change in other repos before publishing).

osa1:
I tested this internally in cl/755740416, merging.

devoncarew:
@osa1 - I think we should plan to publish this soonest. I ran into https://github.com/google/protobuf.dart/issues/975 myself; it's challenging to create message objects w/o this fix.

I think we could publish this independent of the enum optimizations you're working on (but I don't know how ready those are to land).

osa1:
@devoncarew #981 is almost ready but internal testing reveals a failure. I'm not sure yet whether it's related to the patch or a failure that already exists, but it's in a Dart test so I need to investigate it further tomorrow. Making a release today without the enum patch SGTM. (but we should fix the plugin version number as you noted in https://github.com/google/protobuf.dart/pull/986#discussion_r2087195765)

devoncarew:
Makes sense! version rev. PR out: https://github.com/google/protobuf.dart/pull/989

PR #980, opened 2025-05-07
Improve enum decoding

osa1:
Currently when generating the `ProtobufEnum` value for an enum value on the
wire, we call the function `valueOfFunc` of `FieldInfo` of the enum field,
which directly indexes the enum class's `_byValue` map.

With this PR we avoid the closure call by making `_byValue` public and calling
it directly.

Wasm benchmarks:

- Before: `protobuf_PackedEnumDecoding(RunTimeRaw): 48200.0 us`
- After: `protobuf_PackedEnumDecoding(RunTimeRaw): 42120.0 us`
- Diff: -12.6%

AOT benchmarks:

- Before: `protobuf_PackedEnumDecoding(RunTimeRaw): 49180.0 us`
- After: `protobuf_PackedEnumDecoding(RunTimeRaw): 45726.82 us` 
- Diff: -7%

See also alternative PR: #981.

### DIFF
--- a/protobuf/lib/src/protobuf/builder_info.dart
+++ b/protobuf/lib/src/protobuf/builder_info.dart
@@ -62,7 +62,7 @@ class BuilderInfo {
       dynamic defaultOrMaker,
       CreateBuilderFunc? subBuilder,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       {String? protoName}) {
     final index = byIndex.length;
@@ -72,7 +72,7 @@ class BuilderInfo {
             defaultOrMaker: defaultOrMaker,
             subBuilder: subBuilder,
             valueOf: valueOf,
-            valueOfMap: valueOfMap,
+            enumValueMap: enumValueMap,
             enumValues: enumValues,
             protoName: protoName);
     _addField(fieldInfo);
@@ -100,7 +100,7 @@ class BuilderInfo {
       CheckFunc<T> check,
       CreateBuilderFunc? subBuilder,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       {ProtobufEnum? defaultEnumValue,
       String? protoName}) {
@@ -108,7 +108,7 @@ class BuilderInfo {
     _addField(FieldInfo<T>.repeated(
         name, tagNumber, index, fieldType, check, subBuilder,
         valueOf: valueOf,
-        valueOfMap: valueOfMap,
+        enumValueMap: enumValueMap,
         enumValues: enumValues,
         defaultEnumValue: defaultEnumValue,
         protoName: protoName));
@@ -131,11 +131,11 @@ class BuilderInfo {
       {dynamic defaultOrMaker,
       CreateBuilderFunc? subBuilder,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       String? protoName}) {
     add<T>(tagNumber, name, fieldType, defaultOrMaker, subBuilder, valueOf,
-        valueOfMap, enumValues,
+        enumValueMap, enumValues,
         protoName: protoName);
   }
 
@@ -176,11 +176,11 @@ class BuilderInfo {
   void e<T>(int tagNumber, String name, int fieldType,
       {dynamic defaultOrMaker,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       String? protoName}) {
     add<T>(tagNumber, name, fieldType, defaultOrMaker, null, valueOf,
-        valueOfMap, enumValues,
+        enumValueMap, enumValues,
         protoName: protoName);
   }
 
@@ -196,13 +196,13 @@ class BuilderInfo {
   void pc<T>(int tagNumber, String name, int fieldType,
       {CreateBuilderFunc? subBuilder,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       ProtobufEnum? defaultEnumValue,
       String? protoName}) {
     assert(_isGroupOrMessage(fieldType) || _isEnum(fieldType));
     addRepeated<T>(tagNumber, name, fieldType, _checkNotNull, subBuilder,
-        valueOf, valueOfMap, enumValues,
+        valueOf, enumValueMap, enumValues,
         defaultEnumValue: defaultEnumValue, protoName: protoName);
   }
 
@@ -248,7 +248,7 @@ class BuilderInfo {
       required int valueFieldType,
       CreateBuilderFunc? valueCreator,
       ValueOfFunc? valueOf,
-      Map<int, ProtobufEnum>? valueOfMap,
+      Map<int, ProtobufEnum>? enumValueMap,
       List<ProtobufEnum>? enumValues,
       ProtobufEnum? defaultEnumValue,
       PackageName packageName = const PackageName(''),
@@ -265,7 +265,7 @@ class BuilderInfo {
               valueDefaultOrMaker,
               valueCreator,
               valueOf,
-              valueOfMap,
+              enumValueMap,
               enumValues);
 
     addMapField<K, V>(tagNumber, name, keyFieldType, valueFieldType,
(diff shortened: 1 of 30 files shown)


protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -169,11 +169,11 @@ class BuilderInfo {
   // Enum.
   void e<T>(int tagNumber, String name, int fieldType,
       {dynamic defaultOrMaker,
-      ValueOfFunc? valueOf,
```
mkustermann, line 172: Aren't these changes removing those parameters from public APIs? They seem to be breaking changes. So old generated code will no longer work with newest `package:protobuf`. If `package:protobuf` has to upgrade to new major version then it causes trouble for ecosystem:

If an app depends transitively on A and B and A uses old and B uses new protobuf version, then we get constraint conflict.

osa1: Yes, it's a breaking change. You need to update `protoc_plugin` and `protobuf` together.

We've done it in the past, for example with the most recent `protobuf` and `protoc_plugin`s.

> If an app depends transitively on A and B and A uses old and B uses new protobuf version, then we get constraint conflict.

Yes, but you can't do any major bumps and breaking changes if you want to avoid this. We deal with it the same way we've dealt with it in the past.

protoc_plugin/lib/src/generated/descriptor.pb.dart
```diff
@@ -765,12 +767,12 @@ class FieldDescriptorProto extends $pb.GeneratedMessage {
     ..e<FieldDescriptorProto_Label>(
         4, _omitFieldNames ? '' : 'label', $pb.PbFieldType.OE,
         defaultOrMaker: FieldDescriptorProto_Label.LABEL_OPTIONAL,
-        valueOf: FieldDescriptorProto_Label.valueOf,
+        enumValueMap: FieldDescriptorProto_Label.byValue,
         enumValues: FieldDescriptorProto_Label.values)
```
mkustermann, line 771: So it seems we are already passing `<XXX>.values` into the `BuilderInfo`.

With this change we also pass in `<XXX>.byValue` into the `BuilderInfo`.

But `<XXX>.byValue` (the new thing we pass into `BuilderInfo` is defined as
```

  static final $core.Map<$core.int, XXX> byValue = $pb.ProtobufEnum.initByValue(values);
```

So actually  the `BuilderInfo` can calculate `XXX.byValue` itself from the `XXX.values` it gets passed.
So it seems we are passing redundant information here, no?

protobuf/lib/src/protobuf/builder_info.dart
```diff
@@ -323,13 +318,12 @@ class BuilderInfo {
 
   ProtobufEnum? _decodeEnum(
       int tagNumber, ExtensionRegistry? registry, int rawValue) {
-    final f = valueOfFunc(tagNumber);
-    if (f != null) {
-      return f(rawValue);
+    final valueMap = fieldInfo[tagNumber]?.enumValueMap;
+    if (valueMap != null) {
+      return valueMap[rawValue];
```
mkustermann, line 323: The benchmarks show this PR is slower than the other. This isn't surprising as we are using a map index here and a list in the other PR.

Now what prevents this PR from using a list instead of a map?

Right now this PR passes in the map in the `BuilderInfo` methods, but it may as well pass a list, why not?

But as mentioned in the other comment we may not need to pass any redundant information in at all.

protobuf/lib/src/protobuf/field_info.dart
```diff
@@ -101,7 +101,7 @@ class FieldInfo<T> {
   FieldInfo(this.name, this.tagNumber, this.index, this.type,
       {dynamic defaultOrMaker,
       this.subBuilder,
-      this.valueOf,
+      this.enumValueMap,
```
mkustermann, line 104: Instead of getting this new parameter, why can't we take `enumValues` we already get and calculate this based on that? And if we can do that, we can decide whether we calculate a list or a map depending on density of the enum values, etc.

protobuf/lib/src/protobuf/builder_info.dart
osa1: > So from your experience it doesn't cause too much churn on the ecosystem?

I don't have too much experience update downstream with major version bumps, but I expect this particular change to not be a big problem.

Maybe someone who works on updating dependencies can comment how much churn this kind of thing causes.

Internally, no one notice this change other than maybe their benchmarks getting faster.

protobuf/lib/src/protobuf/field_info.dart
osa1: > Instead of getting this new parameter, why can't we take enumValues we already get and calculate this based on that? And if we can do that, we can decide whether we calculate a list or a map depending on density of the enum values, etc.

We have one function to decode enum `int` values: https://github.com/google/protobuf.dart/blob/006d3aa9cbf27a8f8925e2eb96bef7754792124a/protobuf/lib/src/protobuf/builder_info.dart#L324-L334

For it to index a list sometimes and map other times, we would need a type test, or a closure, as in `master` branch and the alternative PR.

Re: generating the list or map from the `enumValues` argument: the map and list should be per-type, but runtime allocated. So it needs to be in a `final static` in the generated the enum classes. Example: https://github.com/google/protobuf.dart/blob/006d3aa9cbf27a8f8925e2eb96bef7754792124a/protoc_plugin/lib/src/generated/plugin.pbenum.dart#L30-L31

And then this static field needs to be passed to `FieldInfo` so that `BuilderInfo` can find it from the `FieldInfo`.

Does this make sense?

protobuf/lib/src/protobuf/builder_info.dart
osa1: Answered in the other comment.

protoc_plugin/lib/src/generated/descriptor.pb.dart
osa1: This is because all `BuilderInfo`s need to share the `byValue` map or list for the same enum type.

mkustermann: Ah, I see. So we want to avoid making N `byValue` maps/lists for the same enum.

I think the smell here is this:
* Each `Foo extends GeneratedMessage` has metadata about it in the `<Foo>.info_` getter.
* Each `FooEnum extends ProtobufEnum` does not have metadata about it in a similar `<FooEnum>.info_` getter

Instead it has these various different static variables as well as `valueOf` closure, ... that we pass around individually - instead of as one e.g. `ProtobufEnumDescriptor` object.

protobuf/lib/src/protobuf/field_info.dart
mkustermann: Makes sense.

But instead of passing the representation (e.g. `Map<int, XX>`, `List<XX>` in here) or a `valueOf` closure. One could pass an instance of a `ProtobufEnumDescriptor` in here that has a `.decode()` method.

This could be a direct call, it could have a `List<T?> values` and a `bool` (whether it's indexed by value or binary search). It would then lookup and return. So there wouldn't be indirect calls (neither closure calls nor dispatch table calls). We would also pass the integer argument unboxed (in closure call we pass it boxed).

Something like
```dart
// package:protobuf
class ProtobufEnum {}
class ProtobufEnumDescriptor {
  final bool _binarySearch;
  final List<ProtobufEnum?> _byValue;
  ProtobufEnumDescriptor(this._binarySearch, this._byValue);  

  ProtobufEnum? lookup(int value) { ... }
}

// foo.pbenum.dart
class FooEnum extends ProtobufEnum {
  ...
  static final info_ = ProtobufEnumDescriptor(values, useBinarySearch: <if sparse>);
}

// foo.pb.dart

class Foo extends GeneratedMessage {
  final info_ = BuilderInfo()
    ..enumField(FooEnum.info_);
}
```

Then the decoder would see
```dart
  ProtobufEnum? _decodeEnum(int tagNumber, ExtensionRegistry? registry, int rawValue) {
    final f = enumDescriptorOf(tagNumber);
    if (f != null) {
      return f.lookup(rawValue); // direct call, passing `rawValue` unboxed, does array lookup or binary search
    }
    ...;
  }
```

Would that work?
If so, wouldn't that be the most efficient way to do it?

osa1: I think this would work and I agree that it should improve things.

Compared to the current PR, it's similar to passing a list (instead of a map) + a bool to whether the list should be binary searched or indexed.

A `BuilderInfo.enumField` wouldn't work. The enum infos need to be attached to field infos for repeated and map fields as well, so we need to pass the info to `BuilderInfo.m` (adds map field info) and `BuilderInfo.addRepeated`.

An `class ProtobufEnumDescriptor` would also mean a level of indirection when accessing the list and the boolean flag for whether to binary search. So perhaps passing an extra argument to the current methods for the `bool` flag would perform the best.

I'll first update the other PR with binary search because that's easy to do. Then update benchmarks to have some sparse enums/enums with large encodings. We can merge the benchmarks separately. Then revisit this idea.

osa1:
Closing this one as alternatives #981 and #985 are faster.

protobuf/lib/src/protobuf/builder_info.dart
devoncarew: Releasing a major version (w/ planning, batching breaking changes, planning deprecations, ...) should be just a normal part of the process. The diamond constraint issue becomes a problem when one of the packages in the cycle has become unmaintained (rev'ing their deps doesn't happen or happens w/ lots of latency). You're more likely to run into that when more things depend on the package being released - when it's further down the overall ecosystem stack.

For something like protobuf I'd just do regular planning for a new major version but would still rev. when necessary. Even for breaking changes, you want to minimize the work for customers - they will eventually need to rev from 3 => 4, or 4 => 5.

There are a few packages that are very core to the ecosystem, and you have to be cautious about rev'ing, or choose to not rev at all (like, the direct deps of flutter - https://github.com/flutter/flutter/blob/master/packages/flutter/pubspec.yaml#L8).

PR #981, opened 2025-05-08
Improve enum decoding -- alternative

osa1:
When decoding an enum value, we call a callback in the enum field's `FieldInfo`. The callback then indexes a map mapping enum numbers to Dart values.

When these conditions hold:

- The known enum numbers are all positive. (so that we can use a list and index it with the number)

- The known enum numbers are more than 70% of the large known enum number. (so that the list won't have a lot of `null` entries, wasting heap space)

We now generate a list instead of a map to map enum numbers to Dart values.

Similar to the map, the list is runtime allocated. No new code generated per message or enum type.

AOT benchmarks:

- Before: `protobuf_PackedEnumDecoding(RunTimeRaw): 47585.14 us.`
- After: `protobuf_PackedEnumDecoding(RunTimeRaw): 38974.566666666666 us.`
- Diff: -18%

Wasm benchmarks:

- Before: `protobuf_PackedEnumDecoding(RunTimeRaw): 52225.0 us.`
- After: `protobuf_PackedEnumDecoding(RunTimeRaw): 34283.33333333333 us.`
- Diff: -34%

**Alternatives considered:**

- #980 uses a map always, but eliminates the `valueOf` closure.
- #985 uses a list always, and does binary search in the list when the list is "shallow".
- #987 is the same as #985, but instead of calling the `valueOf` closure it stores an extra field in `FieldInfo`s for whether to binary search or directly index.

These are all slower than the current PR.

---

cl/757724889

### DIFF
--- a/protobuf/lib/src/protobuf/protobuf_enum.dart
+++ b/protobuf/lib/src/protobuf/protobuf_enum.dart
@@ -41,13 +41,13 @@ class ProtobufEnum {
 
   /// This function is for generated code.
   ///
-  /// Creates a Map for all of the [ProtobufEnum]s in [byIndex], mapping each
+  /// Creates a Map for all of the [ProtobufEnum]s in [enumValues], mapping each
   /// [ProtobufEnum]'s [value] to the [ProtobufEnum].
   ///
   /// @nodoc
-  static Map<int, T> initByValue<T extends ProtobufEnum>(List<T> byIndex) {
+  static Map<int, T> initByValue<T extends ProtobufEnum>(List<T> enumValues) {
     final byValue = <int, T>{};
-    for (final v in byIndex) {
+    for (final v in enumValues) {
       byValue[v.value] = v;
     }
     return byValue;
@@ -56,10 +56,19 @@ class ProtobufEnum {
   /// This function is for generated code.
   ///
   /// @nodoc
-  static List<T?> $_initByValueList<T extends ProtobufEnum>(List<T> byIndex) {
-    if (byIndex.isEmpty) return [];
-    final byValue = List<T?>.filled(byIndex.last.value + 1, null);
-    for (final enumValue in byIndex) {
+  static List<T?> $_initByValueList<T extends ProtobufEnum>(List<T> enumValues) {
+    if (enumValues.isEmpty) return [];
+
+    var maxValue = enumValues[0].value;
+    for (var i = 0; i < enumValues.length; i += 1) {
+      final value = enumValues[i].value;
+      if (value > maxValue) {
+        maxValue = value;
+      }
+    }
+
+    final byValue = List<T?>.filled(maxValue + 1, null);
+    for (final enumValue in enumValues) {
       byValue[enumValue.value] = enumValue;
     }
     return byValue;
--- a/protobuf/lib/src/protobuf/protobuf_enum.dart
+++ b/protobuf/lib/src/protobuf/protobuf_enum.dart
@@ -37,9 +37,16 @@ class ProtobufEnum {
   /// Creates a new constant [ProtobufEnum] using [value] and [name].
   const ProtobufEnum(this.value, this.name);
 
-  /// Creates a Map for all of the [ProtobufEnum]s in [byIndex], mapping each
-  /// [ProtobufEnum]'s [value] to the [ProtobufEnum].
-  static Map<int, T> initByValue<T extends ProtobufEnum>(List<T> byIndex) {
+  static List<T?> initByValueList<T extends ProtobufEnum>(List<T> byIndex) {
+    if (byIndex.isEmpty) return [];
+    final byValue = List<T?>.filled(byIndex.last.value + 1, null);
+    for (final enumValue in byIndex) {
+      byValue[enumValue.value] = enumValue;
+    }
+    return byValue;
+  }
+
+  static Map<int, T> initByValueMap<T extends ProtobufEnum>(List<T> byIndex) {
     final byValue = <int, T>{};
     for (final v in byIndex) {
       byValue[v.value] = v;
--- a/protobuf/lib/src/protobuf/protobuf_enum.dart
+++ b/protobuf/lib/src/protobuf/protobuf_enum.dart
@@ -47,8 +47,8 @@ class ProtobufEnum {
   /// @nodoc
   static Map<int, T> initByValue<T extends ProtobufEnum>(List<T> enumValues) {
     final byValue = <int, T>{};
-    for (final v in enumValues) {
-      byValue[v.value] = v;
+    for (final enumValue in enumValues) {
+      byValue[enumValue.value] = enumValue;
     }
     return byValue;
   }
--- a/protobuf/lib/src/protobuf/protobuf_enum.dart
+++ b/protobuf/lib/src/protobuf/protobuf_enum.dart
@@ -57,18 +57,9 @@ class ProtobufEnum {
   ///
   /// @nodoc
   static List<T?> $_initByValueList<T extends ProtobufEnum>(
-      List<T> enumValues) {
-    if (enumValues.isEmpty) return [];
-
-    var maxValue = enumValues[0].value;
-    for (var i = 0; i < enumValues.length; i += 1) {
-      final value = enumValues[i].value;
-      if (value > maxValue) {
-        maxValue = value;
-      }
-    }
-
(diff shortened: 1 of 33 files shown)


protobuf/lib/src/protobuf/protobuf_enum.dart
```diff
@@ -37,9 +37,16 @@ class ProtobufEnum {
...
+    if (byIndex.isEmpty) return [];
+    final byValue = List<T?>.filled(byIndex.last.value + 1, null);
+    for (final enumValue in byIndex) {
+      byValue[enumValue.value] = enumValue;
+    }
+    return byValue;
+  }
+
+  static Map<int, T> initByValueMap<T extends ProtobufEnum>(List<T> byIndex) {
     final byValue = <int, T>{};
```
mkustermann, line 50: This could also use `List<T>` where we do binary search in. It may be much more compact in memory, quite possibly also faster.

The binary searching code could also live here as a static
```
static T? binarySearchFor<T extendfs ProtobufEnum>(List<T> byValue) {
  ...
}
```

protobuf/lib/src/protobuf/protobuf_enum.dart
```diff
@@ -37,9 +37,16 @@ class ProtobufEnum {
   /// Creates a new constant [ProtobufEnum] using [value] and [name].
   const ProtobufEnum(this.value, this.name);
 
-  /// Creates a Map for all of the [ProtobufEnum]s in [byIndex], mapping each
-  /// [ProtobufEnum]'s [value] to the [ProtobufEnum].
```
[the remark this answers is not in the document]
mkustermann: This is kind of a public API: Pub packages check in the generated code of running protoc plugin. The generated code uses this API. So if one updates `package:protobuf` version with change in this API, the checked-in generated protobuf files won't work anymore.

protobuf/lib/src/protobuf/protobuf_enum.dart
```diff
@@ -37,9 +37,16 @@ class ProtobufEnum {
...
-  static Map<int, T> initByValue<T extends ProtobufEnum>(List<T> byIndex) {
```
mkustermann, line 42: To avoid making this a breaking change, you can just keep this function under this name (even if it's unused by generated code from new generator, package may still use it)

mkustermann (commented):
I'd be ok with landing this, but see my comments on the other PR. If they make sense, maybe that could be an even better solution?

protoc_plugin/lib/src/enum_generator.dart
```diff
@@ -175,11 +177,32 @@ class EnumGenerator extends ProtobufContainer {
...
 
-      out.println(
-          'static final $coreImportPrefix.Map<$coreImportPrefix.int, $classname> _byValue ='
-          ' $protobufImportPrefix.ProtobufEnum.initByValue(values);');
-      out.println('static $classname? valueOf($coreImportPrefix.int value) =>'
-          ' _byValue[value];');
+      var useList = _canonicalValues.isEmpty;
+      if (_canonicalValues.isNotEmpty) {
+        if (_canonicalValues.every((val) => !val.number.isNegative)) {
+          if (_canonicalValues.length / (_canonicalValues.last.number + 1) >=
```
osa1, line 183: TODO: Can the largest known enum value be `-1`, or do we always need to define `0`? If we don't need to define `0` always then this can divide by zero when the largest enum value is `-1`.

protobuf/lib/src/protobuf/protobuf_enum.dart
osa1: This is done in #985 and #987, but the results are not that much better.

protoc_plugin/lib/src/enum_generator.dart
osa1: It looks like the first enum value needs to be `0`. Resolving.

osa1:
@mkustermann I tried:

- Using a map always, but eliminating the closure: #980.
- Keeping the closure, but using a list instead of map when the enum is "dense": this PR.
- Keeping the closure, but using a list always, doing binary search or indexing based on whether the enum is "dense": #985.
- Eliminating the closure, doing binary search or indexing based on whether the enum is "dense": #987.

Among these, this PR makes most sense to merge, as it's backwards compatible, and it performs the best.

If you could take another look I'd like to merge this.

In the meantime I'll test it internally.

protoc_plugin/lib/src/enum_generator.dart
```diff
@@ -175,11 +174,27 @@ class EnumGenerator extends ProtobufContainer {
...
 
-      out.println(
-          'static final $coreImportPrefix.Map<$coreImportPrefix.int, $classname> _byValue ='
-          ' $protobufImportPrefix.ProtobufEnum.initByValue(values);');
-      out.println('static $classname? valueOf($coreImportPrefix.int value) =>'
-          ' _byValue[value];');
+      final useList = _canonicalValues.isEmpty ||
+          (_canonicalValues.every((val) => !val.number.isNegative) &&
+              _canonicalValues.length / (_canonicalValues.last.number + 1) >=
+                  0.7);
```
osa1, line 180: Bug here: `_canonicalValues` is in the order the enum values appear in the source code, so the last enum here won't necessarily have the largest enum number.

I'll try to change order of this in g3 to see if it breaks anything. I don't think having `values` in source code order is useful. We should either sort it based on names (which will allow proto3 JSON decoder to binary search when decoding enums) or sort it based on values (so that the last element will give us the largest value both in runtime when initializing the lists, and in compile time).

mkustermann (approved):

protoc_plugin/lib/src/enum_generator.dart
```diff
@@ -175,11 +174,37 @@ class EnumGenerator extends ProtobufContainer {
...
+
+      if (useList) {
+        out.println(
+            'static final $coreImportPrefix.List<$classname?> _byValue ='
+            ' $protobufImportPrefix.ProtobufEnum.\$_initByValueList(values);');
+
+        out.println('static $classname? valueOf($coreImportPrefix.int value) =>'
+            '  value < 0 || value >= _byValue.length ? null : _byValue[value];');
+      } else {
+        out.println(
```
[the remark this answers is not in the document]
osa1: Yes, see the enums called "sparse" in the tests.

protobuf/lib/src/protobuf/protobuf_enum.dart
```diff
@@ -37,16 +39,42 @@ class ProtobufEnum {
...
     }
     return byValue;
   }
 
+  /// This function is for generated code.
+  ///
+  /// @nodoc
+  static List<T?> $_initByValueList<T extends ProtobufEnum>(
+      List<T> enumValues) {
+    if (enumValues.isEmpty) return [];
```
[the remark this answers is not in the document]
osa1: I forgot that `filled` returns fixed-length by default. Fixed, thanks!

protobuf/lib/src/protobuf/protobuf_enum.dart
```diff
@@ -37,16 +39,42 @@ class ProtobufEnum {
...
+
+    var maxValue = enumValues[0].value;
```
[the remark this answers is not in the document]
osa1: Done, thanks!

osa1:
Internal testing reveals a failure which may be related. I'll investigate further before merging this.

devoncarew:
@osa1 - I'm seeing some issue w/ this change locally when generating the firestore API.

For `ProtobufEnum` subclasses, the generated code is:

```dart
  ...
  static final $core.List<TargetChange_TargetChangeType?> _byValue = $pb.ProtobufEnum.$_initByValueList(values, 4);
  ...
```

But the `ProtobufEnum.$_initByValueList` reference doesn't exist:

```
The method '$_initByValueList' isn't defined for the type 'ProtobufEnum'.
```

It looks like `ProtobufEnum.$_initByValueList()` was added to package:protobuf. I think we'll want to rev the package version (to a new minor version) + changelog, and plan to publish concurrent w/ the next protoc_plugin publish.

osa1:
@devoncarew right, sorry, I should've documented this in the changelog probably.

This changes both the library and plugin. Library changes are backwards compatible, if you don't change your generated classes you can update the library.

But if you update the plugin and re-generate your protos, then you need the new version of the library.

So I think this should be a minor version bump in the library and major one in the plugin.

PR #982, opened 2025-05-08
Update protoc_plugin pre-generated protos

osa1:


osa1:
These changes add noise to PRs that update generated code, like #980 and #981.

devoncarew (approved):
lgtm

If it's not super critical to review changes to these files, you can configure it so that changes to these files are not expanded by default in a github PR. A reviewer can always expand and inspect if interested. You just indicate that some glob of files is machine generated, ala:

https://github.com/dart-lang/build/blob/master/.gitattributes

osa1:
> lgtm
>
> If it's not super critical to review changes to these files, you can configure it so that changes to these files are not expanded by default in a github PR. A reviewer can always expand and inspect if interested. You just indicate that some glob of files is machine generated, ala:
>
> https://github.com/dart-lang/build/blob/master/.gitattributes

Ah, interesting. I'm not sure if we want to skip reviewing these files always. The main problem is if we don't re-generate these files as you change `protoc_plugin`, when we have to re-generate them in a later PR we end up with irrelevant changes. If we make sure to always re-generate them that shouldn't be an issue. I think this is probably a safer option as we don't miss any important changes as we work on `protoc_plugin`.

So perhaps a CI check in PRs that re-generates the files and checks whether they change would be a better option.

devoncarew:
> So perhaps a CI check in PRs that re-generates the files and checks whether they change would be a better option.

👍

PR #986, opened 2025-05-12
Add fromDart() and toDart() methods to convert between $core.Duration and proto Duration

AgamAgarwal:


protoc_plugin/test/duration_test.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:fixnum/fixnum.dart';
+import 'package:protoc_plugin/test/protos/google/protobuf/duration.pb.dart'
```
osa1, line 6: Imports are different in the g3 and OS versions. This should be something like: `import '../out/protos/google/protobuf/duration.pb.dart';`

osa1 (approved):
@AgamAgarwal Thanks!

If you could fix the import in the test we can merge this.

osa1:
@AgamAgarwal it looks like there's a formatting issue as well..

protoc_plugin/CHANGELOG.md
```diff
@@ -3,9 +3,12 @@
 * Fix factory argument types for protobuf `Map` fields. ([#975])
 * Fix import order changes when files are passed in different order to `protoc`.
 ([#952])
+* Add fromDart() and toDart() methods to convert between core Duration and proto
```
devoncarew, line 6: As a drive-by-comment, since this is adding new API, we should bump the minor version of the plugin (`22.0.2-wip` => `22.1.0-wip`). This will let people that know they need this API specify that in their dep (`^22.1.0`).

PR #985, opened 2025-05-09
Improve enum decoding -- alternative with binary search instead of map lookup

osa1:
This PR is the same as #981, except instead of map lookup we do binary search.

This is slightly faster than #981 on Wasm, and slightly slower than #981 on native.

Wasm results:

```
// PR 981
protobuf_PackedEnumDecoding(RunTimeRaw): 36483.33333333333 us.
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 56450.0 us.

// This PR
protobuf_PackedEnumDecoding(RunTimeRaw): 38016.66666666667 us.
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 53750.0 us.
```

AOT results:

```
// PR 981
protobuf_PackedEnumDecoding(RunTimeRaw): 42096.54 us.
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 55845.825 us.

// This PR
protobuf_PackedEnumDecoding(RunTimeRaw): 44518.14 us.
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 60258.2 us.
```

Note: in the benchmarks we have a small number (13) known values in the enum. I wonder if the benchmark results change in favor of one or the other with larger number of known values.

PR #987, opened 2025-05-12
Improved enum decoding 4

osa1:
This is like #985, but the closure for decoding enum values is replaced with list indexing or binary search, without calling a closure.

This is significantly slower than #985. AOT benchmarks:

```
//  #985
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 55845.825 us.

// This PR
protobuf_PackedSparseEnumDecoding(RunTimeRaw): 76054.36666666667 us.
```

Wasm benchmarks are also similarly slow.


osa1:
Closing in favor or #981.

PR #991, opened 2025-05-14
read from and generate grpc service metadata

devoncarew:
- read `default_host` and `oauth_scopes` options from gRPC service definitions and write that information to the generated gRPC clients

In generated service clients this looks like:

```dart
  /// The hostname for this service.
  static const $core.String defaultHost = 'www.example.com';

  /// OAuth scopes needed for the client.
  static const $core.List<$core.String> oauthScopes = [
    'https://www.googleapis.com/auth/cloud-platform',
    'https://www.googleapis.com/auth/datastore',
  ];
```

W/o this, people will have to look up the correct info for the service through some other means.


protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 22.2.0-wip
+
+* Read `default_host` and `oauth_scopes` options from gRPC service definitions
+  and write that information to the generated gRPC clients.
```
osa1, line 4: Do we have any documentation showing generated classes and how they should be used?

If we do, it would be good to update that with the new members.

If not, maybe we should show the new members here?

osa1 (approved):

protoc_plugin/CHANGELOG.md
devoncarew: Here's what they look like in proto files: https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L53

and what they look like in generated code:

```dart
class FirestoreClient extends $grpc.Client {
  /// The hostname for this service.
  static const $core.String defaultHost = 'firestore.googleapis.com';

  /// OAuth scopes needed for the client.
  static const $core.List<$core.String> oauthScopes = [
    'https://www.googleapis.com/auth/cloud-platform',
    'https://www.googleapis.com/auth/datastore',
  ];

  ...
```

We'd need some additional protos (`google/api/annotations.proto`, ...) to be able to use these annotations in this repo. I think having an example here would be useful, but we'll need to re-work a bit how we bring the google/* protos in, perhaps what our goldens look like, ... That's all useful, but likely work for future PRs. I'll add some docs to the readme or to the changelog for now.

devoncarew:
@ osa1, fyi, I rev'd to a new stable version for protoc_plugin in prep for publishing.

Separately, https://github.com/dart-lang/labs/tree/main/pkgs/appengine has many generated google protos and grpc services (appengine, datastore, logging, iam, ...). With a little work it could serve as more effective canary for new releases than the sdk repo.

PR #992, opened 2025-05-14
remove the api_benchmark/ package

devoncarew:
- remove the api_benchmark/ package

This PR removes `api_benchmark/`. I believe we get all our our perf validation from `benchmarks/`; cc @osa1 and @mraleph for whether that's true, and, any thoughts on removing or retaining `api_benchmark`.


osa1 (approved):
:+1: these benchmarks should be gone, or rewritten.

The benchmarked functions are tiny and don't do anything that you want to benchmark and improve. They're getters in messages, which just index the field set.

The way they benchmark is extremely complicated for reasons that I never understood. Why is it running a web server and load the web page to benchmark `fieldSet._$getS(...)`?

I haven't reviewed every single line in this directory, but it looks like an incorrect way of benchmarking things that don't matter.

devoncarew:
Perhaps these benchmarks made sense at one point or where an interesting project to develop. But its a net benefit to have less to maintain overall.

PR #996, opened 2025-05-16
Move plugin dependency_overrides from separate file to pubspec.yaml

osa1:
It's more common to have the overrides in the main pubspec.


devoncarew (approved):
lgtm

We can also switch to using the newer pub workspace support. There wouldn't be any downsides for this repo and we could remove the overrides completely.

osa1:
> We can also switch to using the newer pub workspace support. There wouldn't be any downsides for this repo and we could remove the overrides completely.

Makes sense, done in #1001.

PR #999, opened 2025-05-16
shorten the deprecation messages for clone() and copyWith()

devoncarew:
- shorten the deprecation messages for clone() and copyWith(); reference a tracking issue for the deprecations

Done as two commits for easier review.


osa1:
What's the goal with this? The issue and old depreciation message have the same information, but now it's more difficult to see as you have to click on a link and load a web page.

devoncarew:
The generated code is currently pretty hard to read (and the code serves as API - both for regular protos, and for generated grpc client libraries - so readability matters). I think switching from three lines for the deprecation annotation + description to one line helps here.

osa1 (approved):

PR #1001, opened 2025-05-19
Switch to pub workspace

osa1:


devoncarew (approved):
🎉 

We should also mention in the changelogs for protobuf and protoc_plugin that the new min. sdk is 3.6.

osa1:
> We should also mention in the changelogs for protobuf and protoc_plugin that the new min. sdk is 3.6.

Good point, do you know if this should be a major or minor version bump?

Technically it could break protobuf users and require them to bump Dart minor version, but bumping Dart minor version won't break any code. So I'm not sure if we consider this PR as breaking?

devoncarew:
> Technically it could break protobuf users and require them to bump Dart minor version, but bumping Dart minor version won't break any code. So I'm not sure if we consider this PR as breaking?

I think we generally don't consider it breaking, but more something to note in the changelog. They'll still be able to solve in the previous publish, if they're doing something like running on a 3.5 SDK. You can probably just append to the current changelog entry (which is ready to publish).

WDYT about publishing after this and the deprecation annotation PR land? I'll have more changes this week, some of which will require minor version bumps (new API).

osa1:
Ah, I thought we already published the major versions as we don't have the `-wip` suffix. I updated the changelogs.

> WDYT about publishing after this and the deprecation annotation PR land? I'll have more changes this week, some of which will require minor version bumps (new API).

SGTM.

protoc_plugin/CHANGELOG.md
```diff
@@ -5,36 +5,38 @@
...
+    static const $core.String defaultHost = 'firestore.googleapis.com';
+  
+    /// OAuth scopes needed for the client.
+    static const $core.List<$core.String> oauthScopes = [
+      'https://www.googleapis.com/auth/cloud-platform',
+      'https://www.googleapis.com/auth/datastore',
+    ];
+  
+    ...
+  ```
```
osa1, line 35: This block is indented so that it'll be a part of the bulleted list.

PR #1003, opened 2025-05-20
Update the generated code to improve readability and to better follow common Dart patterns.

devoncarew:
update the generated code to improve readability and to better follow common Dart patterns:

- use `final` on some fields
- introduce whitespace around some groups of code
- avoid very short (one letter) param names
- use `@override` annotations where relevant
- add type annotations to some fields

Done as a few separate commits for easier review.


osa1 (approved):
I wonder if the short names were deliberate. We use cryptic names in other places too, for example `BuilderInfo.aOM` instead of `BuilderInfo.addOptionalMessage`. @sigurdm may know more about the name choices here.

sigurdm:
I vaguely remember something about the short names being to make the non-minified dart2js output shorter. Not sure how meaningful that is
@rakudrama might be the right one to ask.

sigurdm:
Seems like these short names has been there from the very first commit: https://github.com/google/protobuf.dart/commit/d6b192352a4816810b7f5688bd30cc16f05cd57a#diff-43bc0c5c4223d3a4fb2ee3c3e974ec23ef00940dcef72809fec175832259b1edR36

Perhaps they were inherited from java protobuf or something...

devoncarew:
> I vaguely remember something about the short names being to make the non-minified dart2js output shorter. Not sure how meaningful that is @rakudrama might be the right one to ask.

I'm hoping that the short param names (which is what's adjusted in this PR) aren't (or are no longer) relevant to dart2js? I did rename and then changed back the `_i` BuilderInfo field, after I saw that it was used by the protobuf tree-shaker.

PR #1004, opened 2025-05-21
add a script to update the vendored protos

devoncarew:
- add a script to update the vendored protos
- add a new `google/api/routing.proto` proto; update existing protos (descriptor.proto and compiler/plugin.proto were older)
- regenerate dart files

### DIFF
--- a/protoc_plugin/tool/update_protos.dart
+++ b/protoc_plugin/tool/update_protos.dart
@@ -9,9 +9,9 @@ library;
 import 'dart:io';
 
 void main(List<String> args) async {
-  final dartToolDir = Directory('.dart_tool');
-  final protobufDir = Directory('.dart_tool/protobuf');
-  final googleapisDir = Directory('.dart_tool/googleapis');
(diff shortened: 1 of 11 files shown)


protoc_plugin/tool/update_protos.dart
```diff
@@ -0,0 +1,95 @@
...
+    final target = File('${to.path}/$file');
+
+    if (!target.parent.existsSync()) {
+      target.parent.createSync(recursive: true);
+    }
+
+    if (!target.existsSync() ||
+        target.readAsStringSync() != source.readAsStringSync()) {
+      print('  $file');
+      target.writeAsStringSync(source.readAsStringSync());
```
osa1, line 92: This reads the source file twice, once in the `if` condition and once here. Maybe have a variable to keep source contents when target doesn't exist?

osa1 (approved):

protoc_plugin/tool/update_protos.dart
devoncarew: 👍 , will do

PR #1009, opened 2025-05-23
move generated test code from out/ to test/gen/

devoncarew:
Some mechanical refactors to the test dirs layout:
- don't generate two copies of the test protos (into out/ and into out/constructor_args); just generate code (using constructor_args) into out/
- refactor the code that is testing use of constructor args from portions of two separate test files into one new test file (`constructor_args_test.dart`)
- move generated test code from `out/` to `test/gen/`

We have half as much generated code for use by the tests now, and having it in `test/gen/` is maybe a little more clear that it's used by the tests than having it in `out/`.

Some future work might be to de-dup the protos in `test/protos/google` and reuse the ones in `protos/google`.


protoc_plugin/Makefile
```diff
@@ -81,16 +81,7 @@ $(TEST_PROTO_LIBS): $(PLUGIN_SRC) $(TEST_PROTO_SRCS)
 	mkdir -p $(TEST_PROTO_DIR)
 
 	protoc\
-		--dart_out="disable_constructor_args:$(TEST_PROTO_DIR)"\
```
osa1, line 84: So we no longer `disable_constror_args`? Is it not necessary?

Maybe we could rely on g3 tests for this. In g3 `disable_constructor_args` is the default as it causes binary size bloat.

devoncarew: We're generating two versions of the test protos output for this one option - I'm just trying to dial things back a bit (we don't do this for other options). The disable_constructor_args option is tested in protoc_plugin/test/file_generator_test.dart and a bit in message_generator_test.dart.

Given the history in the changelog It sounds like having ctor args is the default externally, and not having them is the default internally.

PR #1016, opened 2025-05-29
revert the change to not generate empty enum files

devoncarew:
Revert the change to not generate empty enum files (partial revert of https://github.com/google/protobuf.dart/pull/1006). Some enum files export other enums files, and we don't have any easy way to determine whether the file we're exporting was not generated.


osa1 (approved):
It would be good to have a regression test so that we won't break this again.

devoncarew:
> It would be good to have a regression test so that we won't break this again.

For sure; I updated google/protobuf/unittest_import.proto to the latest from upstream; it now has a 'public import' of google/protobuf/unittest_import_public.proto (also added). This would have generated a broken export reference before this fix; the enum file for `unittest_import` would have had an export for the enum file for `unittest_import_public`; that would not have been generated as it's empty.

PR #1019, opened 2025-06-05
Add a helper for unchecked downcasts and use it

osa1:
Lints don't like `final T x = expr as dynamic` pattern, but that pattern is faster with dart2js compared to `expr as T`.

Add a helper to encapsulate the strange pattern and use it everywhere.

Not ready: this seems to reduce dart2js performance:

```
// Before
protobuf_query_encode_binary(RunTimeRaw): 9578.260869565218 us.

// After
protobuf_query_encode_binary(RunTimeRaw): 10208.333333333332 us.
```

I will revert changes to existing `as` casts and only use the new helper on `final T x = expr as dynamic`s.

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -72,7 +72,7 @@ class _FieldSet {
   /// If [_frozenState] contains a boolean, the hashCode hasn't been memoized,
   /// so it will return null.
   int? get _memoizedHashCode =>
-      _frozenState is int ? _frozenState as int : null;
+      _frozenState is int ? downcastUnchecked<int?>(_frozenState) : null;
 
   /// Maps a `oneof` field index to the tag number which is currently set. If
   /// the index is not present, the oneof field is unset.
@@ -214,7 +214,7 @@ class _FieldSet {
   /// Works for both repeated and non-repeated fields.
   dynamic _getFieldOrNull(FieldInfo fi) {
     if (fi.index != null) return _values[fi.index!];
-    return _extensions?._getFieldOrNull(fi as Extension<dynamic>);
+    return _extensions?._getFieldOrNull(fi);
   }
 
   bool _hasField(int tagNumber) {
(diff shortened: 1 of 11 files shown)


protobuf/lib/src/protobuf/json.dart
```diff
@@ -24,22 +24,22 @@ Map<String, dynamic> _writeToJsonMap(_FieldSet fs) {
...
+        final value = downcastUnchecked<double>(fieldValue);
         if (value.isNaN) {
           return _nan;
         }
         if (value.isInfinite) {
           return value.isNegative ? _negativeInfinity : _infinity;
         }
-        if (fieldValue.toInt() == fieldValue) {
-          return fieldValue.toInt();
+        if (value.toInt() == fieldValue) {
```
sigmundch, line 34: drive by comment: will this behave differently in wasm than in JS? Do you need `value.toInt().toDouble() == value` or something else to ensure we are in the same vaue domain?

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -287,8 +287,8 @@ class _FieldSet {
     assert(!fi.isRepeated);
     if (fi.index == null) {
       _ensureExtensions()
-        .._addInfoUnchecked(fi as Extension<dynamic>)
-        .._setFieldUnchecked(fi, value);
+        .._addInfoUnchecked(downcastUnchecked<Extension>(fi))
+        .._setFieldUnchecked(downcastUnchecked<Extension>(fi), value);
```
sigmundch, line 291: I doubt this matters, but would it be helpful to store the cast value here too? 

```
final fiExtension = downcastUnchecked<Extension>(fi);
_ensureExtensions()
   .._addInfoUnchecked(fi)
   .._setFieldUnchecked(fi, value);
```

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -72,7 +72,7 @@ class _FieldSet {
   /// If [_frozenState] contains a boolean, the hashCode hasn't been memoized,
   /// so it will return null.
   int? get _memoizedHashCode =>
-      _frozenState is int ? _frozenState as int : null;
+      _frozenState is int ? downcastUnchecked<int?>(_frozenState) : null;
```
osa1, line 75: This line makes the kernel worse, probably because of TFA results getting worse because of the function call introduced. (the call is inlined, but TFA runs before inlinings)

rakudrama (commented):
I feel that this CL is trying to make things better by adding more 'worse'.
The original code has several associations between values and type. (fieldType bits are correlated with FieldInto subtypes and the types of values).

Can this association be better expressed by looking at the algorithm at a higher level?
Can it be done in a way that is flexible enough that the codebase could converge on supporting protobuf 'editions'?

osa1:
> I feel that this CL is trying to make things better by adding more 'worse'.

It wouldn't be worse if inlining was done before TFA, or we had macros.

The current type casts confuse people all the time. See #905, #967, #1017.

If this worked we wouldn't have lint errors everywhere that people try to fix every quarter.

PR #1020, opened 2025-06-10
update the generator to emit formatted files

devoncarew:
- update the generator to emit formatted files

This uses package:dart_style to format the generated proto files.

There is additional complexity here as we support an option to generate metadata about the source location of symbols for `.pb.dart` and `.pbenum.dart` files. I see a reference to this being in support of the Kythe indexer. I'm not sure why it needs us to generate metadata files for it to use - it can just parse dart source directly. Perhaps this option is no longer needed? It would be nice to be able to remove it for general simplification.

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 22.4.0
+## 22.5.0-wip
 
 * Generated files are now formatted using the Dart formatter.
+
+## 22.4.0
+
 * Update how we calculate import prefixes ([#1010]); import prefixes are now
   unique per-library instead of being unique across all generated libraries.
 * Ignore `unused_import` diagnostics for `*.pbjson.dart` files. ([#1013])
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 22.4.0-wip
 
+* Generated files are now formatted using the Dart formatter.
 * Update how we calculate import prefixes ([#1010]); import prefixes are now
   unique per-library instead of being unique across all generated libraries.
 * Ignore `unused_import` diagnostics for `*.pbjson.dart` files.
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 22.5.0-wip
 
-* Generated files are now formatted using the Dart formatter.
+* Generated files are now formatted using the Dart formatter. The code is
+  formatted using the min. SDK for `package:protoc_plugin`; currently `3.6.0`.
 
 ## 22.4.0
 
--- a/protoc_plugin/lib/indenting_writer.dart
+++ b/protoc_plugin/lib/indenting_writer.dart
@@ -4,6 +4,7 @@
 
 import 'dart:collection';
 
+import 'src/formatter.dart' as formatter;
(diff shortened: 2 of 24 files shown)


devoncarew:
@osa1, once this lands (after review and such), I may have one more small style PR, then would likely want to do another publish.

osa1:
Why do you want the plugin to generate formatted files?

The problem with formatting generated files in protoc plugin is that the plugin and the project that uses generated files can have different formatting settings and foramtter versions. This may even be the common case for non-Google packages, or if you rarely update your protoc plugin as you update your Dart versions.

So you have to format generated files anyway with your project's formatter.

I don't remember if adding `dart_style` as a dependency was a problem for g3, but we should check that as well.

(I don't have access to g3 right now, I can check tomorrow.)

FWIW, grpc generator used to format generated files in the open source version but not in the g3 version. That may suggest that there's a problem with adding dart_style dependency to the g3 version. #702 then removed formatting in the open source version as well.

Maybe an alternative could be formatting the files with `dart format` (as a shell command) rather than with a formatter built into the plugin. That way the formatting would be done using the project's settings rather than the plugin's.

devoncarew:
After thinking about this a bit more, I updated this PR (https://github.com/google/protobuf.dart/pull/1020/commits/09eaaa851b2e56e4e0f27d81f4ce779708d5e3c9) to always format using the Dart version from the min. sdk constraint in the protoc_plugin pubspec. This:

- is predictable for users; a particular version of `protoc_plugin` will always format the same
- doesn't require us to plumb a new parameter through the dart plugin options, or, require the user to know about a new option
- means we don't have to do something potentially unpredictable, like looking around the filesystem in the output directory for a parent pubspec file, and reading the sdk constraint version from that

If its necessary for a particular use of protoc_plugin to have the files formatted at a specific version, they can always generate+format as part of their regular regeneration process.

osa1 (approved):
Btw, I asked around a little bit and found this prior art on formatting generated codes:

- https://github.com/dart-lang/build/issues/3887
- https://github.com/dart-lang/build/issues/3812

There's also a build_runner package for Dart protos: https://github.com/pikaju/dart-protoc-builder

I don't know if it does formatting itself.

protoc_plugin/lib/indenting_writer.dart
```diff
@@ -118,18 +125,27 @@ class IndentingWriter {
...
       for (final key in _suffixes.keys) {
         println(_suffixes[key]!);
       }
       _suffixes.clear();
     }
 
-    return _buffer.toString();
+    var source = _buffer.toString();
+
+    // We don't always want to format the source (for example, we don't want to
```
sigurdm, line 142: This is an interesting problem. It would be cool if we had some way of "anchoring" the annotated locations in the source.

Something like https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html#Line-Control-1 ...

@lrhn wdyt?

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.5.0-wip
+
+* Generated files are now formatted using the Dart formatter.
```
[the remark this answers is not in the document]
devoncarew: 👍 ; done

protoc_plugin/lib/indenting_writer.dart
sigurdm: I believe these are for references from .proto files to the generated code. Not sure a dart parser works directly here. (but perhaps via symbol names...)

PR #1021, opened 2025-06-13
improve the readability of generated gRPC client files

devoncarew:
Improve the readability of generated gRPC client files:

- emit the method descriptors at the end of the client class definition (after the service methods)
- emit the `foo_Pre()` and `foo()` methods together, instead of in separate parts of the service base class

### DIFF
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,10 +1,11 @@
-## 22.4.0-wip
+## 22.4.0
 
 * Update how we calculate import prefixes ([#1010]); import prefixes are now
   unique per-library instead of being unique across all generated libraries.
 * Ignore `unused_import` diagnostics for `*.pbjson.dart` files.
 * Revert the change to not generate empty `*.pbenum.dart` files; these can be
   exported from other enum files.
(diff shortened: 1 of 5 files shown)


devoncarew:
It looks like `pub publish --dry-run` is having trouble w/ the goldens in `test/` (which we don't expect to be analyzed):

```
Validating package...
  Package validation found the following potential issue:
  * line 16, column 1 of test/goldens/grpc_service.pbgrpc.dart: This package does not have grpc in the `dependencies` or `dev_dependencies` section of `pubspec.yaml`.
       ╷
    16 │ import 'package:grpc/service_api.dart' as $grpc;
       │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       ╵
```

We may need to move them into a directory which pub publish doesn't try to analyze, ignore the goldens somehow from the POV of pub, or make the files analyze correctly.

mosuem:
> We may need to move them into a directory which pub publish doesn't try to analyze, ignore the goldens somehow from the POV of pub, or make the files analyze correctly.

I think I normally do a `test_data` folder outside of `test` to avoid these kinds of issues. Not sure if it helps here.

PR #1024, opened 2025-07-10
Update SDK version to 3.7, reformat with 3.7 formatter

osa1:
Internally we're at SDK 3.7. Becuase it's difficult to make changes *without* formatting in g3, contributors create CLs with accidental formatting fixes.

In cl/781448639 we format the internal code, and this PR updates to the same SDK version with g3 and formats the code, making the internal and external versions of the code meet at the same SDK version and formatting settings.

### DIFF
--- a/benchmarks/bin/deep_copy.dart
+++ b/benchmarks/bin/deep_copy.dart
@@ -15,11 +15,14 @@ class Benchmark extends BenchmarkBase {
   final p3.GoogleMessage1 _message1Proto3;
   final GoogleMessage2 _message2;
 
-  Benchmark(super.name, List<int> message1Proto2Input,
-      List<int> message1Proto3Input, List<int> message2Input)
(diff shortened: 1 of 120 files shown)


devoncarew (approved):
👍

protobuf/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 4.2.0-wip
```
devoncarew, line 1: This could be a patch bump (we're not introducing new API), but a minor is fine as well.

protoc_plugin/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 22.5.0-wip
```
devoncarew, line 1: It's also possible to go to `22.4.1-wip` instead (a patch bump).

PR #1028, opened 2025-07-18
Sync JSON web decoders

osa1:
This syncs the web JSON decoder used internally.

The new decoder uses `js_interop` instead of the Dart standard library to
improve performance when compiled to JS.

## Benchmarks

master, JS:

    protobuf_from_json_string(RunTimeRaw): 1423.134328358209 us.

PR, JS:

    protobuf_from_json_string(RunTimeRaw): 971 us. 

Wasm and VM benchmarks don't change as they keep using the old implementation.

The new implementation is slower with Wasm as it does more JS interop, so we keep using the old one.


osa1:
@devoncarew  PTAL.

protobuf/lib/src/protobuf/json/json_web.dart
```diff
@@ -0,0 +1,463 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
devoncarew, line 1: ditto re: the date

protobuf/lib/src/protobuf/json/json_web.dart
devoncarew, line 1: For this file - given that it's a refactor - I'm guessing we're relying on existing tests to ensure correctness. I don't see us running tests on the CI for anything more than the vm however. I think we'll want to do the equivalent of:

- `dart test`
- `dart test -p chrome`
- `dart test -p chrome -c dart2wasm`

I'm not sure how to express that in the mono_repo config; perhaps you can crib it from https://github.com/dart-lang/web/blob/46d416f53d8ae148835cf82b26c32c189755b9a3/web/mono_pkg.yaml ?

protobuf/lib/src/protobuf/json/json_web.dart
```diff
@@ -0,0 +1,463 @@
...
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+
+import 'package:fixnum/fixnum.dart' show Int64;
+
+import '../consts.dart';
+import '../internal.dart';
+import '../utils.dart';
+
+@JS('JSON')
```
devoncarew, line 15: I don't think it's a big deal, but these extension types and external method means that `json_web.dart` has public API which is a strict superset of `json_vm.dart`. When using config specific imports do try to keep the APIs for each implementation of the library the same.

protobuf/lib/src/protobuf/json/json.dart
```diff
@@ -2,9 +2,17 @@
...
+import 'dart:convert' show base64Decode, base64Encode;
 
-Map<String, dynamic> _writeToJsonMap(FieldSet fs) {
+import 'package:fixnum/fixnum.dart' show Int64;
+
+import '../consts.dart';
+import '../internal.dart';
+import '../utils.dart';
+
+export 'json_vm.dart' if (dart.library.html) 'json_web.dart';
```
devoncarew, line 13: Iff this uses the vm impl on the VM and when compiled with dart2wasm, and the web impl for dart2js, you might add a comment to that effect.

devoncarew (approved):
I have a few comments; but given:

> This syncs the web JSON decoder used internally.

I think we can consider the code reviewed. I think my only blocking comment is that we should update the CI here to test against the vm, dart2js, and dart2wasm.

devoncarew (approved):

PR #1029, opened 2025-07-21
CI: Test PRs and commits to all branches, instead of just master

osa1:


osa1:
@devoncarew this allows testing PRs based on other PRs, like #1028 and #1027.

mono_repo.yaml
```diff
@@ -5,9 +5,7 @@ github:
   on:
     # Run CI on pushes to the master branch, and on PRs against master.
     push:
```
devoncarew, line 7: I think we want to keep this - to run the workflow on all pushes to `master` (mostly this would happen when a PR gets merged and squashed into a master commit).

```suggestion
    push:
      branches: master    
```

.github/workflows/dart.yml
```diff
@@ -2,9 +2,7 @@
 name: Dart CI
 on:
   push:
```
devoncarew, line 4:
```suggestion
  push:
    branches: master
```

devoncarew (approved):
lgtm w/ the two suggested edits (we don't want to include the branch filter for pull requests; we do want to retain it for the 'push' events).

PR #1030, opened 2025-07-24
Remove monorepo

mosuem:


.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    branches:
+    paths:
+      - '.github/workflows/benchmark.yaml'
+      - 'benchmark/**'
+  push:
+    branches: [ master ]
+    paths:
+      - '.github/workflows/benchmark.yaml'
+      - 'benchmark/**'
```
osa1, line 14: This should be `benchmarks` (plural)

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - run: dart analyze --fatal-infos
+
+      - run: dart format --output=none --set-exit-if-changed lib
+
+      - run: ./../tool/setup.sh
+
+      - run: ./tool/compile_protos.sh
+
+      - run: dart test
```
osa1, line 44: `dart test` (after all the preparation steps) fails on my system because there are no tests in this directory.

Instead of `dart test` we should analyze each of the files in `bin/` or maybe even run them.

To compile the programs we have a Dart script in `tool/compile_benchmarks.dart`. I feel like it we just analyze the files in `bin/` or compile them that should be fine. Running will take long and I'm not sure if running them can catch any issues. (after `dart analyze` passes)

.github/workflows/protobuf.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - run: dart pub get
+
+      - run: dart analyze --fatal-infos
+
+      - run: dart format --output=none --set-exit-if-changed .
+        if: ${{matrix.run-tests}}
+
+      - run: dart test
+        if: ${{matrix.run-tests}}
```
osa1, line 47: If one of these steps fail the rest won't run, right?

That's not ideal because then a formatting issue blocks test results. With multiple packages in the same repo that need preparation steps before testing, it's not uncommon to rely on CI for the test results.

So we should make these steps independent if possible and try to collect as many results as possible from each CI run.

.github/workflows/protoc_plugin.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      - run: make protos
+
+      - run: dart analyze --fatal-infos
+
+      - run: dart format --output=none --set-exit-if-changed lib
+        if: ${{matrix.run-tests}}
+
+      - run: dart test
+        if: ${{matrix.run-tests}}
```
osa1, line 48: Same as above, it would be good to make the steps independent.

.github/workflows/protobuf.yaml
```diff
@@ -0,0 +1,47 @@
...
+  push:
+    branches: [ master ]
+    paths:
+      - '.github/workflows/protobuf.yaml'
+      - 'protobuf/**'
+  schedule:
+    - cron: '0 0 * * 0' # weekly
+
+jobs:
+  build:
```
osa1, line 19: The job name is "build" but it builds, checks formatting, analyzes, tests. If we want all of these in a single job maybe a more generic name like "Check protobuf" would be good. (but I think we should split this into jobs, see below)

osa1:
@mosuem since your branch is on this repo (instead of a fork) I can take this from here if you prefer that or don't have the time.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - 'benchmarks/**'
+      - 'tool/setup.sh'
+  push:
+    branches: [ master ]
+    paths:
+      - '.github/workflows/benchmarks.yaml'
+      - 'benchmarks/**'
+      - 'tool/setup.sh'
+  schedule:
+    - cron: '0 0 * * 0' # weekly
```
osa1, line 18: Why do we want to run this periodically?

.github/workflows/protobuf.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
+        with:
+          sdk: ${{matrix.sdk}}
+
+      - run: dart pub get
+
+      - run: dart analyze --fatal-infos
+
+      - run: dart format --output=none --set-exit-if-changed .
+        if: ${{matrix.run-tests && (success() || failure()) }}
```
osa1, line 44: Is `matrix.run-tests` ever `false`? It looks hard-coded above.

.github/workflows/protoc_plugin.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - 'protoc_plugin/**'
+      - 'tool/setup.sh'
+  push:
+    branches: [ master ]
+    paths:
+      - '.github/workflows/protoc_plugin.yaml'
+      - 'protoc_plugin/**'
+      - 'tool/setup.sh'
+  schedule:
+    - cron: '0 0 * * 0' # weekly
```
osa1, line 18: Same as above, not sure why we want periodic jobs.

osa1 (commented):
In a previous revision of this CL you had `continue-on-error` or similar. What happened to that?

Will the current setup will collect as many results as possible before failing? Otherwise it gets really annoying when a formatting error blocks your test results.

.github/workflows/protoc_plugin.yaml
mosuem: See above.

mosuem:
> In a previous revision of this CL you had `continue-on-error` or similar. What happened to that?

That has the problem that it sets the check to success, so gives a green checkmark in the UI, regardless of whether any step failed or not.

> Will the current setup will collect as many results as possible before failing? Otherwise it gets really annoying when a formatting error blocks your test results.

Yes it does, it runs through all the checks, not stopping when of them fails.

.github/workflows/protobuf.yaml
mosuem: `run-tests` is true only for the `stable` sdk, so we don't test on `dev`. Feel free to change that if needed.

PR #1031, opened 2025-07-24
CI: Also run plugin tests on Chrome, with dart2js and dart2wasm

osa1:


osa1:
We need to skip golden tests when running tests in browsers as we can't use the file system API in the browser. I'll update.

protoc_plugin/mono_pkg.yaml
```diff
@@ -12,4 +12,6 @@ stages:
       - command: ./../tool/setup.sh
       - command: make protos
       - test
+      - test: -p chrome
```
devoncarew, line 15: I don't know that we need to run the protoc_plugin tests in the browser (but haven't examined the tests).

osa1: Tests that need proto files are in `protoc_plugin/test`, so lots of useful tests there. We have all kinds of tests, encoding, decoding, API, ... 7,434 lines of tests in protoc_plugin vs. 1,474 lines (excluding comments, whitespace, and generated files).

osa1:
This reveals an existing issue in the library or the tests. I'm working on fixing it.

devoncarew (approved):
🎉

PR #1037, opened 2025-08-06
Make proto3_json.dart aware of well known types

osa1:
This syncs cl/629836458.

From the CL:

This removes the requirement for proto3_json.dart to invoke outside code
(either via BuilderInfo or inheritance). This also improves tree shaking
of the proto3json helpers, since the proto3json callbacks attached to
BuilderInfo no longer retain those helpers.

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.2.0-wip
+## 5.0.0-wip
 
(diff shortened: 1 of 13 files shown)


devoncarew (approved):
lgtm as already reviewed (`This syncs cl/629836458`).

However, we may consider rev'ing package:protobuf from 4.2.0-wip to 4.2.0 as a separate PR and publishing. That would capture the ~2 changes that are in head but not yet published.

Issue #372, opened 2020-05-24
Dart enums generated by protoc don't follow dart naming conventions


Pikaju:
When giving a protobuf enum a scream case name, it will remain scream case in the generated Dart code, even though this does not match the Dart standard.

blindmatrix:
Any updates on this?

PR #1045, opened 2025-08-28
Follow Protobuf Enum style through a new cli option "protobuf-enum-style"

Spiritus2424:
This PR changes the enum generation to convert protobuf enum definition following the protobuf style guideline to respect dart naming convention at the end.

It is essential to take into account that people will follow the Protobuf style guideline and naming convention and that developer will expect the dart protoc plugin to match the dart naming conventions.

https://github.com/google/protobuf.dart/issues/372 Also motivates this change.

I tried to keep everything as compatible as possible. To not expose breaking change, I hide the new enum generation under an option (false by default) `protobuf-enum-style`.

Close https://github.com/google/protobuf.dart/issues/862


Spiritus2424 (commented):
I added comments to help the reviewer understand what the changes are and why this does not bring any breaking change.

@osa1

osa1:
Could you please remind me what Dart naming conventions the current generated code violates? Do you mean [the `constant_identifier_names` lint][1]? Or something else?

Why is this a blocker for your organization? We already generate `ignore_for_file: constant_identifier_names` in `.pbenum.dart` files. Is the generated code still causing issues?

[1]: https://dart.dev/tools/linter-rules/constant_identifier_names

Spiritus2424:
Not 100%, but yes, this code was assisted by an LLM. I am currently using Cursor which use GPT4/Claude behind the scenes. 

Does this cause an issue?

Spiritus2424:
Hi @osa1,

Would it be possible to get a quicker review on this? Our team is motivated to invest time and contribute to this library to support our needs. However, if we're unable to get smaller changes reviewed in a timely manner, we may need to consider maintaining our own fork.

Thanks in advance!

PR #1047, opened 2025-09-04
Sync cl/658131451

osa1:
(Not my CL, I'm just syncing it to open source)

CL description: (slightly edited)

Use a map to cache the `valueOf` function for enums.

This is a generated-code size optimization.

The main change is to generate the valueOf functions for protobuf enums programatically, rather than (1) use the per-enum `FooPbEnum.valueOf` static function tear-foo which (2) uses a per-enum static variable that is lazily initialized to map from int value to the protobuf enum value.

With ~1000 protobuf enums, this allows most of the `valueOf` static function tear-off closures to tree-shaken, together with their functions and the code for the initializer of the map.

The secondary change is to use more specialized 'add' methods to `BuilderInfo`, allowing many fields to be 'defined' with a call with fewer arguments.

Together, these changes reduce the size of the main unit of an app by about 1.3%, and would have a similar effect on native targets.

This change requires https://github.com/dart-lang/sdk/commit/175dc05db3cd626874c17ceb03336065a03375bb, which was released with Dart 3.6.0.

### DIFF
--- a/protoc_plugin/lib/src/protobuf_field.dart
+++ b/protoc_plugin/lib/src/protobuf_field.dart
@@ -262,6 +262,16 @@ class ProtobufField {
     } else if (isRepeated) {
       if (typeConstant == '$protobufImportPrefix.PbFieldType.PS') {
         invocation = 'pPS';
+      } else if (typeConstant == '$protobufImportPrefix.PbFieldType.PE') {
+        invocation = 'pPE<$type>';
+        named['enumValues'] = '$type.values';
+        final makeDefault = generateDefaultFunction(omitIfFirstEnumValue: true);
+        if (makeDefault != null) {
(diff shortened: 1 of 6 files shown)


sigurdm (approved):

protoc_plugin/lib/src/protobuf_field.dart
```diff
@@ -273,21 +283,35 @@ class ProtobufField {
         if (baseType.isMessage || baseType.isGroup) {
           named['subBuilder'] = '$type.create';
         } else if (baseType.isEnum) {
+          // TODO: remove this branch.
```
sigurdm, line 286: I wonder if we should open an issue or at least assign a name to this TODO.
Or perhaps expand it - why can this be removed? (not this PR)

PR #1059, opened 2025-09-29
Update CHANGELOGs for release

osa1:
@sigurdm @devoncarew do you have anything else to add to this release?

Note: pubspec versions are already have the right versions (no `-wip` suffix).

### DIFF
--- a/protobuf/CHANGELOG.md
+++ b/protobuf/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 5.0.0-wip
+## 5.0.0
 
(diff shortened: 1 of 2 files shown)


sigurdm (approved):
I'm good

osa1:
We could release fixnum and bump fixnum dependency here, to get the perf benefits with the new fixnum.

I restarted the g3 job to roll fixnum to g3 in https://dart-review.googlesource.com/c/sdk/+/449020 to see if the issues with conditional imports was fixed.

sigurdm:
Would the new fixnum not be compatible? Such that we really don't need to change the constraint?

PR #1062, opened 2025-10-14
Fix a few extension field bugs

osa1:
When an extension field set isn't initialized before freezing the message:

- `addExtension` allows adding and modifying extensions.
- `getExtension` returns non-frozen defaults that can be modified. 

Fix this by initializing the extension field set as frozen when the parent message is frozen.

Secondly, extension field set returns differently typed lists when it's frozen and not frozen. Fix this by always creating the list via the field's `FieldInfo`.

Also refactor the code a little bit: `GeneratedMessage` by-passes a layer of abstraction and accesses unknown field sets directly. Instead make it call `FieldSet` always.

New `FieldSet` methods are added as public (without an underscore in the name) to make it clear that they're a public interface (can be called from `GeneratedMessage`).

(Ideally `FieldSet` should be a separate library, but that requires a lot of refactoring to split the parts into libraries.)

Also remove duplicate "is repeated" and "ensure writable" checks when modifying extension fields.

---

cl/819146715

### DIFF
--- a/protobuf/lib/src/protobuf/field_set.dart
+++ b/protobuf/lib/src/protobuf/field_set.dart
@@ -106,18 +106,8 @@ class FieldSet {
   /// The [FieldInfo] for each non-extension field in tag order.
   Iterable<FieldInfo> get _infosSortedByTag => _meta.sortedByTag;
 
-  ExtensionFieldSet _ensureExtensions() {
-    var extensions = _extensions;
-    if (extensions != null) {
-      return extensions;
-    }
-    extensions = ExtensionFieldSet(this);
-    _extensions = extensions;
-    if (_isReadOnly) {
-      extensions._markReadOnly();
-    }
-    return extensions;
-  }
+  ExtensionFieldSet _ensureExtensions() =>
+      _extensions ??= ExtensionFieldSet(this, readOnly: _isReadOnly);
 
   UnknownFieldSet _ensureUnknownFields() {
     if (_unknownFields == null) {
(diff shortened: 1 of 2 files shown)


protobuf/CHANGELOG.md
```diff
@@ -5,7 +5,16 @@
 
   The new limits are consistent with the Java and C++ implementations. ([#1060])
 
+* Fix `GeneratedMessage.addExtension` returning non-frozen and
+  `GeneratedMessage.getExtension` allowing modifying an extension when the
+  message is frozen before initializing the extension field set. ([#1062])
+
+* Fix `GeneratedMessage.getExtension` returning differently typed lists when the
+  message extension field set is initialized and frozen and initialized but not
+  frozen. ([#1062])
```
osa1: Note: this is not fixing a bug caused by the previous entry, it exists in the master branch independently of the other bugs. However the repro is different in the master branch vs. this branch.

protobuf/lib/src/protobuf/field_set.dart
```diff
@@ -106,8 +106,18 @@ class FieldSet {
...
   Iterable<FieldInfo> get _infosSortedByTag => _meta.sortedByTag;
 
-  ExtensionFieldSet _ensureExtensions() =>
-      _extensions ??= ExtensionFieldSet(this);
+  ExtensionFieldSet _ensureExtensions() {
+    var extensions = _extensions;
+    if (extensions != null) {
+      return extensions;
+    }
+    extensions = ExtensionFieldSet(this);
```
sigurdm: Doesn't really matter, but if you passed the readOnly property in the constructor, you could go back to:
```
_extensions ??= ExtensionFieldSet(this, readOnly: _isReadOnly);
```

PR #1070, opened 2025-10-24
Don't iterate iterable arguments twice in PbList

### LINKED ISSUE #730: `PbList.addAll` traverses the iterable twice, causes side effects to run twice
Reported by an internal user.

https://github.com/google/protobuf.dart/blob/87135ba79d333a058358a363b6343c2bce813ed3/protobuf/lib/src/protobuf/pb_list.dart#L41-L46

The line `iterable.forEach(_check)` traverses the iterable once. If the iterable has side effects, those will be run for the first time here.

Then the next line `_wrappedList.addAll(iterable)` will again traverse the iterable, causing the side effects to run again.

I don't know if we should call this a bug, but I think it's fair to say that it's a bit unexpected.

Two possible fixes:

1. Copy iterable elements to a list, then traverse the list many times:
   ```dart
   void addAll(Iterable<E> iterable) {
     _checkModifiable('addAll');
     List<E> list = iterable is List ? iterable : iterable.toList();
     list.forEach(_check);
     _wrappedList.addAll(list);
   }
   ```
   The downside is the function will now require allocating an intermediate list, and the list may be large in some uses.

2. Check as we add elements to the wrapped list:
   ```dart
   void addAll(Iterable<E> iterable) {
     _checkModifiable('addAll');
     for (final e in iterable) {
       _check(e);
       _wrappedList.add(e);
     }
   }
   ```
   The downside is if a check fails the list will be updated. To avoid that, we could do:
   ```dart
   void addAll(Iterable<E> iterable) {
     _checkModifiable('addAll');
     var added = 0;
     for (final e in iterable) {
       try {
         _check(e);
       } catch (_) {
         _wrappedList.removeRange(_wrappedList.length - added, _wrappedList.length);
         rethrow;
       }
       _wrappedList.add(e);
     }
   }
   ```

osa1:
Update `PbList` methods that update the list with an `Iterable` argument to
avoid iterating the iterable twice: once for checking the element validity and
once again for actually adding the values.

Methods updated:

- addAll
- insertAll
- replaceRange
- setAll
- setRange

Exception handling behavior before this PR was undefined (same as the standard
library `List`), and it's slightly changed with this PR:

- addAll: previously if the iterator throws the list was left unchanged, now
  the elements until the exception will be added.

- Other methods: exception behaviors are now the same as the standard library
  `List` methods.

  It's hard to tell whether the previous behavior was the same or different
  with the standard library `List` methods, as the exception behavior of those
  are undefined.

To avoid allocating new iterators when a check function is not available, we
have conditionals in each of these methods and call the standard library `List`
methods directly when there isn't a check function.

To avoid increasing number of cases needed to be tested, we don't special case
iterable types like `PbList` and `List` in these methods. When the check
function is available we simply `map` them with the check function. Otherwise
call the same method on `wrappedList` directly.

Fixes #730.

---

cl/823443311

### DIFF
--- a/protobuf/lib/src/protobuf/annotations.dart
+++ b/protobuf/lib/src/protobuf/annotations.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-part of 'internal.dart';
-
 /// Annotation for marking accessors that belong together.
 class TagNumber {
   final int tagNumber;
--- a/protobuf/lib/src/protobuf/exceptions.dart
+++ b/protobuf/lib/src/protobuf/exceptions.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-part of 'internal.dart';
-
 const _truncatedMessageText = '''
 While parsing a protocol message, the input ended unexpectedly
 in the middle of a field. This could either mean that the input
--- a/protobuf/lib/src/protobuf/internal.dart
+++ b/protobuf/lib/src/protobuf/internal.dart
@@ -16,6 +16,7 @@ import 'package:fixnum/fixnum.dart' show Int64;
 import 'package:meta/meta.dart' show UseResult;
 
 import 'consts.dart';
+import 'exceptions.dart';
 import 'json/json.dart' as json_lib;
 import 'json_parsing_context.dart';
 import 'mixins/well_known.dart';
@@ -25,16 +26,16 @@ import 'permissive_compare.dart';
 import 'type_registry.dart';
 import 'utils.dart';
 
+export 'annotations.dart' show TagNumber, GrpcServiceName;
 export 'pb_list.dart' show PbList;
 export 'pb_map.dart' show PbMap;
 export 'type_registry.dart' show TypeRegistry;
+export 'exceptions.dart' show InvalidProtocolBufferException;
 
-part 'annotations.dart';
 part 'builder_info.dart';
 part 'coded_buffer.dart';
 part 'coded_buffer_reader.dart';
 part 'coded_buffer_writer.dart';
-part 'exceptions.dart';
 part 'extension.dart';
 part 'extension_field_set.dart';
 part 'extension_registry.dart';
--- a/protoc_plugin/CHANGELOG.md
+++ b/protoc_plugin/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 24.0.0
+
+* **Breaking:** Don't generate `createRepeated` methods.
+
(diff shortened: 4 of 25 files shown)


osa1:
@rakudrama I can't add you as a reviewer, could you also have a look?

protobuf/lib/src/protobuf/pb_list.dart
```diff
@@ -108,18 +112,32 @@ class PbList<E> extends ListBase<E> {
   void insertAll(int index, Iterable<E> iterable) {
     _checkModifiable('insertAll');
     if (_check != null) {
-      iterable.forEach(_check);
+      _wrappedList.insertAll(
+        index,
+        iterable.map((E e) {
+          _check(e);
+          return e;
+        }),
```
rakudrama: It is worth avoiding calling List.insertAll with a non-list.

Example, this implementation copies the iterable to a list anyway
https://github.com/dart-lang/sdk/blob/56ccc3f887009a04fccaca4817a20a6a755871e0/sdk/lib/_internal/wasm/lib/list.dart#L568

I would do the conversion to list here, and then exploit the fact that we have a list:
```
final List<E> list;
if (iterable is List<E>) {
  if (iterable is PbList<E>) {
    list = iterable._wrappedList;
  } else {
    list = iterable;
  }
} else {
  list = List.of(iterable);
}
if (check != null) list.forEach(_check);
_wrappedList.insertAll(index, list);
```

I am not a fan of the map pattern.

It creates a closure and does a dispatched call to `.map` which allocates some unknown implementation of `Iterable`, which often ends up being iterated at a place that is polymorphic, so the Iterator is allocated in another dispatched call. That is four objects (closure, closure-context, Iterable, Iterator), and one or two dispatched calls to `moveNext` and `current` for each element.

The copy of the list might often be smaller than the `.map` overhead.

I eyeballed the big app there are only a few calls.

Perhaps the 'list selection' code could be abstracted out and reused

```
List<E> _checkedValuesAsList(Iterable<E> iterable) {
  final List<E> list;
  if (iterable is List<E>) {
    if (iterable is PbList<E>) {
      list = iterable._wrappedList;
    } else {
      list = iterable;
    }
  } else {
    list = List.of(iterable);
  }
  if (check != null) list.forEach(_check);
  return list;
}

  void insertAll(int index, Iterable<E> iterable) {
    _checkModifiable('insertAll');
    _wrappedList.insertAll(index, _checkedValuesAsList(iterable));
  }
```

osa1:
With all the branching testing this becomes a quarter long engineering effort.. We need to test all combinations of:

(1) when the iterable is a: PbList, List, and other
(2) when _check is available and not
(3) for all possible error conditions (depends on the method)

Note that error conditions are already not tested today, I wouldn't be surprised if error checking is different based on the check function and argument type today.

I'm inclined to do the simplest possible thing here at the cost of performance. Use a `List` if you want fast `setRange` etc. and add your stuff to the message afterwards.

PR #1081, opened 2025-11-11
Add well-known type support

### LINKED ISSUE #483: Target of URI doesn't exist: '../../../google/protobuf/timestamp.pb.dart'.
# Steps to reproduce
`article.proto`
```protobuf
syntax = "proto3";
package example;

import "google/protobuf/timestamp.proto";

message Article {
  google.protobuf.Timestamp time = 1;
}
```

```
protoc --dart_out=. article.proto
```

The generated Dart code contains an import: `import 'google/protobuf/timestamp.pb.dart' as $0;` that points to a non-existent file path. I can't find any documentation on how to fix this.

### LINKED ISSUE #780: should well-known types be included into the published package?
It feels strange that one needs to manually copy `any.proto` and generate `any.pb.dart`. 

I understand why this is the case for monorepo setup, but for users not operating in monorepo we could probably include a number of prebuilt proto classes into the package itself?

### LINKED ISSUE #817: What is the correct way to include well known types?
generated code has e.g.
```dart

import '../../../google/protobuf/timestamp.pb.dart' as $2;
import '../../../google/protobuf/wrappers.pb.dart' as $3;
```

but those libraries aren't present.

What is the correct way to include those libraries?

### LINKED ISSUE #1073: Add to docs how to get Well Known Types
As can be seen [here](https://github.com/google/protobuf.dart/issues/170) and in many many other issues and stack overflow questions, the protobuf Well Known Types are not automatically generated by the Dart protobuf plugin. I saw some claims that Dart is the only plugin that behaves this way. I understand this is a technical decision from the team, but at least the docs should include how to get the WKT generated and working.

I have hit many caveats, until I finally understood what was going on. If it helps, my description is [here](https://github.com/google/protobuf.dart/issues/170#issuecomment-3464673914), and I believe the docs should include something similar.

osa1:
- Add generated Dart for the well-known protos to the package.

- Special case `google/protobuf/...` imports in the plugin to generate imports
  to the libraries in the protobuf package.

  Note: `protoc` already special cases these imports, you don't have to specify
  paths to them with `-I...`.

  With this change you also don't have to manually compile them to Dart
  classes.

Fixes #1073.
Fixes #817.
Fixes #780.
Fixes #483.

---

Changes to test files to use internal wkt protos: cl/831301744 (merged)
Output configuration change tested in: cl/831310910

### DIFF
--- a/protoc_plugin/pubspec.yaml
+++ b/protoc_plugin/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   dart_style: ^3.0.0
   fixnum: ^1.0.0
   path: ^1.8.0
-  protobuf: ^5.0.0
+  protobuf: ^5.2.0
   pub_semver: ^2.2.0
 
 dev_dependencies:


protoc_plugin/pubspec.yaml
sigurdm:
```suggestion
  protobuf: ^5.2.0
```

sigurdm: This would make the solver help out and avoid mismatches.

analysis_options.yaml
```diff
@@ -4,4 +4,3 @@ linter:
   rules:
   - avoid_bool_literals_in_conditional_expressions
   - comment_references
-  - prefer_relative_imports
```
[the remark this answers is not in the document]
osa1: We don't have control over the .proto files, and even if we did we don't have a way to add a `// ignore`  line in a `.proto` file (like with a Dart plugin specific option).

I also think the lint is a bit useless: no one will import `package:protobuf/lib/foo.dart` when they can just do `import 'foo.dart'`. If you do it then there's probably a good reason as you go out of your way. (like in this case)

PR #1083, opened 2025-11-14
Add dart=3.7 pragmas to protobuf lib code

osa1:
With this we can finally have the same formatting in the internal version and the open source version.

This also syncs some simple changes from internal to the open source.

With this the library code is almost identical internally and in the open source. There are still a few changes that need to be synced (enum names, reflection callbacks), which I'll do separately.

cl/832251175 adds the pragmas internally and reformats internal code. Also includes some trivial syncs from this version to the internal.

To avoid bumping SDK dependency from 3.7 to 3.10, we format with 3.7 instead of 3.10. However note that SDK 3.9 cannot be used to format the repo as it has a bug and the formatter in 3.9 does not respect the `@dart=...` lines.


osa1:
Why 3.7?

Why can't users who can't upgrade to 3.10 use the older version of the library?

The formatter in 3.9 doesn't respect the `// @dart=...` lines, that's why I waited until 3.10 to add these lines. We can add 3.7 as long as we always format with 3.10 or newer.

Why do you want to support 3.7 @natebosch?

You're right that if we have a min. supported version policy there should be checks to enforce it.

I've been maintaining this library for the past 3.5 years and I'm not aware of any such policy.

sigurdm:
I don't think we have a policy. The lower bound cannot be lower than the target of the generated code. So I guess it would make sense to make it ^3.10.0.

That means older darts won't benefit from the new protobuf improvements, but I think that is acceptable.

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -537,7 +539,7 @@ mixin ListValueMixin implements GeneratedMessage {
     if (json is List) {
       final subBuilder = message.info_.subBuilder(_valueFieldTagNumber)!;
       for (var i = 0; i < json.length; i++) {
-        final Object element = json[i];
+        final Object? element = json[i];
```
natebosch: Did you run `dart fix` or similar? Can we avoid code changes in formatting focused PR?

PR #1097, opened 2026-07-14
Tweak uses of RegExp and parsing using them.

lrhn:
Looked at the RegExps and how they are used and made some tweaks and fixes.

Fixed some actual bugs too:
* Negative durations with zero seconds.
* Large timestamps.

### DIFF
--- a/protobuf/lib/src/protobuf/mixins/well_known.dart
+++ b/protobuf/lib/src/protobuf/mixins/well_known.dart
@@ -168,7 +168,7 @@ String _typeNameFromUrl(String typeUrl) {
 }
 
 mixin TimestampMixin {
-  static final RegExp finalGroupsOfThreeZeroes = RegExp(r'(?:000)*$');
+  static final RegExp finalGroupsOfThreeZeroes = RegExp(r'(?:000)+$');
 
   Int64 get seconds;
   set seconds(Int64 value);
@@ -192,9 +192,15 @@ mixin TimestampMixin {
   ///
   /// Time zone information will not be preserved.
   static void setFromDateTime(TimestampMixin target, DateTime dateTime) {
-    final micros = dateTime.microsecondsSinceEpoch;
-    target.seconds = Int64((micros / Duration.microsecondsPerSecond).floor());
-    target.nanos = (micros % Duration.microsecondsPerSecond).toInt() * 1000;
+    var micros = dateTime.microsecondsSinceEpoch;
+    var seconds = micros ~/ Duration.microsecondsPerSecond;
+    micros = micros.remainder(Duration.microsecondsPerSecond);
+    if (micros < 0) {
+      seconds -= 1;
+      micros += Duration.microsecondsPerSecond;
+    }
+    target.seconds = Int64(seconds);
+    target.nanos = micros * 1000;
   }
 
   static String _twoDigits(int n) {
@@ -268,10 +274,10 @@ mixin TimestampMixin {
     JsonParsingContext context,
   ) {
     if (json is String) {
-      var jsonWithoutFracSec = json;
       var nanos = 0;
-      final Match? fracSecsMatch = RegExp(r'\.(\d+)').firstMatch(json);
-      if (fracSecsMatch != null) {
+      var jsonWithoutFracSec = json.replaceFirstMapped(RegExp(r'\.(\d+)'), (
+        fracSecsMatch,
+      ) {
         final fracSecs = fracSecsMatch[1]!;
         if (fracSecs.length > 9) {
           throw context.parseException(
@@ -280,12 +286,8 @@ mixin TimestampMixin {
           );
         }
         nanos = int.parse(fracSecs.padRight(9, '0'));
-        jsonWithoutFracSec = json.replaceRange(
-          fracSecsMatch.start,
-          fracSecsMatch.end,
-          '',
-        );
-      }
+        return '';
+      });
       final dateTimeWithoutFractionalSeconds =
           DateTime.tryParse(jsonWithoutFracSec) ??
           (throw context.parseException(
@@ -319,16 +321,30 @@ mixin DurationMixin {
     TypeRegistry typeRegistry,
   ) {
     final duration = message as DurationMixin;
-    final secFrac = duration.nanos
-        // nanos and seconds should always have the same sign.
+    final seconds = duration.seconds;
+    final nanos = duration.nanos;
+    // Extra sign needed if `seconds` is zero and nanos is negative.
+    final sign = seconds.isZero && nanos < 0 ? '-' : '';
+
+    return '$sign$seconds${_toNanosString(nanos)}s';
+  }
+
+  /// Nanoseconds as fractional seconds.
+  ///
+  /// Is empty if [nanos] is zero, otherwise the digits of
+  /// (absolute value) of `nanos` as billionths of seconds
+  /// with no trailing zeros.
+  static String _toNanosString(int nanos) {
+    if (nanos == 0) return '';
+    final digits = nanos
         .abs()
         .toString()
         .padLeft(9, '0')
         .replaceFirst(finalZeroes, '');
-    final secPart = secFrac == '' ? '' : '.$secFrac';
-    return '${duration.seconds}${secPart}s';
+    return '.$digits';
   }
 
+  // Matches `-?\d*\.?\d*s$`, grouping is only to control captures.
   static final RegExp durationPattern = RegExp(r'(-?\d*)(?:\.(\d*))?s$');
 
   static void fromProto3JsonHelper(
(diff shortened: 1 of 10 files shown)


protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -168,7 +168,7 @@ String _typeNameFromUrl(String typeUrl) {
 }
 
 mixin TimestampMixin {
-  static final RegExp finalGroupsOfThreeZeroes = RegExp(r'(?:000)*$');
+  static final RegExp finalGroupsOfThreeZeroes = RegExp(r'(?:000)+$');
```
lrhn, line 171: (Only used for replacing with `''`, no need to match an empty string.)

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -192,9 +192,15 @@ mixin TimestampMixin {
   ///
   /// Time zone information will not be preserved.
   static void setFromDateTime(TimestampMixin target, DateTime dateTime) {
-    final micros = dateTime.microsecondsSinceEpoch;
-    target.seconds = Int64((micros / Duration.microsecondsPerSecond).floor());
-    target.nanos = (micros % Duration.microsecondsPerSecond).toInt() * 1000;
+    var micros = dateTime.microsecondsSinceEpoch;
+    var seconds = micros ~/ Duration.microsecondsPerSecond;
```
lrhn, line 196: Going through doubles loses precision for large/small dates. Tests added.

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -192,9 +192,15 @@ mixin TimestampMixin {
...
+    micros = micros.remainder(Duration.microsecondsPerSecond);
+    if (micros < 0) {
```
lrhn, line 198: (Usual way to handle `~/`/`remainder` not working like `%`.)

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -268,10 +274,10 @@ mixin TimestampMixin {
     JsonParsingContext context,
   ) {
     if (json is String) {
-      var jsonWithoutFracSec = json;
       var nanos = 0;
-      final Match? fracSecsMatch = RegExp(r'\.(\d+)').firstMatch(json);
-      if (fracSecsMatch != null) {
+      var jsonWithoutFracSec = json.replaceFirstMapped(RegExp(r'\.(\d+)'), (
```
lrhn, line 278: (Could combine match and replace.)

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -319,16 +321,30 @@ mixin DurationMixin {
     TypeRegistry typeRegistry,
   ) {
     final duration = message as DurationMixin;
-    final secFrac = duration.nanos
-        // nanos and seconds should always have the same sign.
+    final seconds = duration.seconds;
+    final nanos = duration.nanos;
+    // Extra sign needed if `seconds` is zero and nanos is negative.
+    final sign = seconds.isZero && nanos < 0 ? '-' : '';
```
lrhn, line 327: This was a bug. It wrote 0 seconds and -500000000 nanos as `0.5`.

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -340,25 +356,33 @@ mixin DurationMixin {
...
-        );
-      } else {
+      if (match != null) {
         final secondsString = match[1]!;
-        final seconds =
-            secondsString == '' ? Int64.ZERO : Int64.parseInt(secondsString);
-        duration.seconds = seconds;
-        final nanos = int.parse((match[2] ?? '').padRight(9, '0'));
-        duration.nanos = seconds < 0 ? -nanos : nanos;
+        final isNegative = secondsString.startsWith('-');
```
lrhn, line 361: This was a bug in the opposite direction.
It parsed `-0.5` as 0 seconds and 500000000 nanos, and didn't negate the nanos since `0` was not negative.

protoc_plugin/lib/src/file_generator.dart
```diff
@@ -4,7 +4,7 @@
 
 part of '../protoc.dart';
 
-final RegExp _dartIdentifier = RegExp(r'^\w+$');
+final RegExp _dartIdentifier = RegExp(r'^[a-zA-Z_]\w*$');
```
lrhn, line 7: That didn't match Dart identifiers, it allowed starting with `0`.
(Also still doesn't allow `$`s, but I assume that's deliberate.)

protoc_plugin/lib/src/protobuf_field.dart
```diff
@@ -6,20 +6,10 @@ part of '../protoc.dart';
...
-    r'^[+-]?([0-9]*)\.[0-9]+(e[+-]?[0-9]+)?$',
-    multiLine: false,
-    caseSensitive: false,
-  );
-  static final RegExp _decimalLiteralRegexB = RegExp(
-    r'^[+-]?[0-9]+e[+-]?[0-9]+$',
-    multiLine: false,
-    caseSensitive: false,
+  static final RegExp _decimalLiteralRegex = RegExp(
+    r'^[+\-]?(\d*\.)?\d+([eE][+\-]?\d+)?$',
```
lrhn, line 12: Instead of three regexpsm just use one and switch on whether it has double-literal parts.

protoc_plugin/lib/src/protobuf_field.dart
```diff
@@ -6,20 +6,10 @@ part of '../protoc.dart';
 
 class ProtobufField {
   static final RegExp _hexLiteralRegex = RegExp(
-    r'^0x[0-9a-f]+$',
-    multiLine: false,
```
lrhn, line 10: `multiLine: false` is the default. 
(And I think being explicit is more readable than the `caseSensitive: false` flag.)

protoc_plugin/lib/src/shared.dart
```diff
@@ -71,25 +71,25 @@ String? toDartComment(String value) {
 
   if (value.isEmpty) return null;
 
-  var lines = LineSplitter.split(value).toList();
+  var lines = LineSplitter.split(value)
+      .map((line) => line.trimRight())
+      .toList();
```
lrhn, line 76: Doing `trimRight` early makes later tests easier, it turns all lines with only whitespace into empty lines.

protoc_plugin/lib/src/shared.dart
```diff
@@ -71,25 +71,25 @@ String? toDartComment(String value) {
...
-      lines =
-          lines
-              .map(
-                (line) => line.isEmpty ? line : line.substring(prefix.length),
-              )
-              .toList();
+      for (var i = 0; i < lines.length; i++) {
+        var line = lines[i];
+        if (line.isNotEmpty) lines[i] = line.substring(prefix.length);
+      }
```
lrhn, line 87: Just update the list in-place instead of building a new one.

protoc_plugin/lib/names.dart
```diff
@@ -592,7 +592,7 @@ String? _nameOption(FieldDescriptorProto field) =>
 
 bool _isDartFieldName(String name) => name.startsWith(_dartFieldNameExpr);
 
-final _dartFieldNameExpr = RegExp(r'^[a-z]\w+$');
+final _dartFieldNameExpr = RegExp(r'^[a-z]\w*$');
```
lrhn, line 595: (Seems unlikely that single-letter fields are not allowed, but two-letter ones are.)

protoc_plugin/test/proto3_json_test.dart
```diff
@@ -450,7 +450,7 @@ void main() {
       predicate((e) {
         if (e is FormatException) {
           final pathExpression =
-              RegExp(r'root(\["[^"]*"]*\])*').firstMatch(e.message)![0]!;
+              RegExp(r'root(?:\["[^"]*"\])*').firstMatch(e.message)![0]!;
```
lrhn, line 453: (I think the extra `]*` here was a typo. If it had been deliberate, it would have been escaped.
Since it matches the empty string, nobody noticed that you _could_ match `root["a"]]]]]`.

protobuf/lib/src/protobuf/mixins/well_known.dart
```diff
@@ -340,25 +356,33 @@ mixin DurationMixin {
...
+        duration.seconds =
+            (secondsString.length == (isNegative ? 1 : 0))
+                ? Int64.ZERO
+                : Int64.parseInt(secondsString);
+        var fractionalSeconds = match[2];
+        var nanos = 0;
+        if (fractionalSeconds != null) {
+          if (fractionalSeconds.length > 9) {
```
lrhn, line 369: The original code would parse `0.9999999999` as 9+ seconds.

I assume the input would be correctly generated and not have more than 9 digits, but better to be safe.
(Could also throw.)

